PR #2, opened 2025-06-30
[1/4] Setup monorepo and typescript

iivvaannxx:
## Description

This PR converts the repo into a [PNPM monorepo](https://pnpm.io/workspaces) with [Turbo](https://turborepo.com/docs). It opts for PNPM Workspaces instead of NPM Workspaces, primarily for the two reasons outlined below. If you're curious about the differences between them, check out [this blog post](https://refine.dev/blog/pnpm-vs-npm-and-yarn/#impact-on-project-structure).

- PNPM is faster and more space-efficient than NPM.
- NPM’s flattened `node_modules` structure can cause issues in a monorepo. With NPM, any package installed anywhere in the monorepo can be accessed by any other package, even if it's not declared in that package’s `package.json`. This often leads to subtle and hard-to-trace bugs. PNPM solves this by strictly isolating dependencies.

### Why use Turbo?
Turborepo is a powerful tool for managing production-ready monorepos. It introduces intelligent caching to speed up repeated tasks and ensures that tasks are executed in the correct dependency order. For example, if you have two packages (`pkg-a` and `pkg-b`), and `pkg-b` depends on `pkg-a`, Turbo will detect any changes in `pkg-a` and build it first before building `pkg-b` when you run the build task for `pkg-b`. If you want to read more, check [the docs](https://turborepo.com/docs).

## Changes Introduced

- Cleanup a lot of files (including linter and formatter), a subsequent PR will add formatting and linting through another tool.
- Create PNPM workspace file and Turbo config file.
- Create a base TypeScript config as a package that can be used by multiple libraries.


jnatherley:
LGTM

jnatherley (approved):

PR #4, opened 2025-06-30
[3/4]: Prepare `@adobe/aio-commerce-lib-auth`

iivvaannxx:
## Description

This PR refactors the pre-existing `aio-commerce-sdk-auth` to integrate it with the mono repo tools. It also replaces Jest by [Vitest](https://vitest.dev/), a faster and more modern testing solution that supports TypeScript by default, without requiring a transpilation step.

It configures [TSDown](https://tsdown.dev/), for building packages. This tool it is specifically designed for bundling TS libraries and it simplifies the process by quite a lot. Is it able to output JS for both CommonJS and ESM formats and it also generates declaration files of the final bundle.
 
It renames the package to `aio-commerce-lib-auth` as it's not an SDK on itself, while also fixing all linting issues that were present in the previous code and running Biome to format it. I refactored the lib to better use TypeScript features and improve type-robustness, which allowed me drop usages of `!` operator (which is not allowed by the linter), among other things.

Finally, it configures some `tasks` in the `turbo.json` and moves TypeScript dependency to the root `package.json` to use a shared TypeScript version across all packages.

## Changes Introduced

- Move Typescript to root `package.json`
- Rename `aio-commerce-sdk-auth` to `aio-commerce-lib-auth` and refactor it to fix lint issues and apply formatting rules.
- Introduce Vitest as the test tool, replacing Jest, and migrate `aio-commerce-lib-auth` tests to use it.
- Introduce TSDown for building packages.
- Adds `tasks` to `turbo.json`.


jnatherley (approved):

obarcelonap (approved):
LGTM
Something to add in the developer docs: I want to create a new lib within the repo, how should I do it? Can we document how to onboard a new lib in this repo setup: what's the minimum files should contain, where to register this new package, scripts should have, etc

iivvaannxx:
Next PR is metapackage + automation (including PR checks and a script to scaffold a new package). I will make sure everything is documented in the last pull request I will do for development guidelines 👍🏻

PR #10, opened 2025-07-07
Write docs and last configs fixes

iivvaannxx:
## Description

This PR adds docs for `CONTRIBUTING` plus `README` files for the current versions of each package at the time of writing. Also runs the formatter on some config files for consistency.


obarcelonap (commented):
Looks great, some comments:
* We need to keep initial `CONTRIBUTING` as it was as it is defined by Adobe OS standards.
* I'd propose to use `DEVELOPMENT` to put all the information you added about how to onboard on developing in this repo.
* Additionally, please ignore autoformat for license files which also part o Adobe OS standards. If you need to restore formatting, please use this repo as ref https://github.com/adobe/starter-repo

PR #69, opened 2025-08-28
[1/3] CEXT-4993: create api utilities

iivvaannxx:
## Description

> [!NOTE]
> Ignore the changes in the `turbo` folder, they are related to private packages generator which I tweaked to generate this private package.

This PR contains utilities for working with the Commerce and I/O Events APIs. This will be used in follow-up PRs for the implementation of the events library.

### DIFF
--- a/packages-private/aio-commerce-lib-api/source/lib/commerce/types.ts
+++ b/packages-private/aio-commerce-lib-api/source/lib/commerce/types.ts
@@ -1,0 +1,63 @@
+import type {
+  ImsAuthProvider,
+  IntegrationAuthParams,
+  IntegrationAuthProvider,
+} from "@adobe/aio-commerce-lib-auth";
+import type { KyInstance, Options } from "ky";
+import type { ImsAuthParamsWithOptionalScopes } from "~/utils/auth/ims-scopes";
+
+/** Defines the flavor of the Commerce instance. */
+export type CommerceFlavor = "paas" | "saas";
+
+/** Defines the configuration required to build an Adobe Commerce HTTP client. */
+export type CommerceHttpClientConfig = {
+  /** The base URL of the Commerce API. */
+  baseUrl: string;
+
+  /** Whether the target Commerce instance is a PaaS or SaaS. */
+  flavor: CommerceFlavor;
+
+  /**
+   * The store view code use to make requests to the Commerce API.
+   * @default "all"
+   */
+  storeViewCode?: string;
+
+  /**
+   * The version of the Commerce API to use. Currently only `v1` is supported.
+   * @default "v1"
+   */
+  version?: "v1";
+};
+
+/** Defines the configuration required to build an Adobe Commerce HTTP client for PaaS. */
+export type CommerceHttpClientConfigPaaS = CommerceHttpClientConfig & {
+  flavor: "paas";
+};
+
+/** Defines the configuration required to build an Adobe Commerce HTTP client for SaaS. */
+export type CommerceHttpClientConfigSaaS = CommerceHttpClientConfig & {
+  flavor: "saas";
+};
+
+// Type for SaaS configuration
+export type SaaSClientParams = {
+  auth: ImsAuthProvider | ImsAuthParamsWithOptionalScopes; // We provide default scopes.
+  config: CommerceHttpClientConfigSaaS;
+  fetchOptions?: Options;
+};
+
+// Type for PaaS configuration
+export type PaaSClientParams = {
+  auth: IntegrationAuthProvider | IntegrationAuthParams;
+  config: CommerceHttpClientConfigPaaS;
+  fetchOptions?: Options;
+};
+
+// Discriminated union of both types
+export type CommerceHttpClientParams = SaaSClientParams | PaaSClientParams;
+
+/** Defines a {@link KyInstance} which targets an Adobe Commerce instance. */
+export type CommerceHttpClient = KyInstance & {
+  config: CommerceHttpClientConfig;
+};
--- a/packages-private/aio-commerce-lib-api/source/lib/io-events/types.ts
+++ b/packages-private/aio-commerce-lib-api/source/lib/io-events/types.ts
@@ -1,0 +1,29 @@
+import type { ImsAuthProvider } from "@adobe/aio-commerce-lib-auth";
+import type { KyInstance, Options } from "ky";
+import type { ImsAuthParamsWithOptionalScopes } from "~/utils/auth/ims-scopes";
+
+/** Defines the configuration required to build an Adobe I/O HTTP client. */
+export type IoEventsHttpClientConfig = {
+  /** The consumer organization ID to use for the Adobe I/O API. */
+  consumerOrgId: string;
+
+  /** The environment to use for the Adobe I/O API. */
+  environment?: "prod" | "stage";
+};
+
+/** Defines the parameters required to build an HTTP client for the Adobe I/O Events API. */
+export type IoEventsHttpClientParams = {
+  /** The IMS authentication parameters. */
+  auth: ImsAuthProvider | ImsAuthParamsWithOptionalScopes;
+
+  /** The configuration for the I/O Events HTTP client. */
+  config: IoEventsHttpClientConfig;
+
+  /** Additional fetch options to use for the I/O Events HTTP requests. */
+  fetchOptions?: Options;
+};
+
+/** Defines the HTTP client for the Adobe I/O Events API. */
+export type IoEventsHttpClient = KyInstance & {
+  config: IoEventsHttpClientConfig;
+};
--- a/packages-private/aio-commerce-lib-api/source/utils/auth/hooks.ts
+++ b/packages-private/aio-commerce-lib-api/source/utils/auth/hooks.ts
@@ -1,0 +1,70 @@
+import {
+  getImsAuthProvider,
+  getIntegrationAuthProvider,
+} from "@adobe/aio-commerce-lib-auth";
+
+import type {
+  ImsAuthParams,
+  ImsAuthProvider,
+  IntegrationAuthParams,
+  IntegrationAuthProvider,
+} from "@adobe/aio-commerce-lib-auth";
+import type { KyRequest } from "ky";
+import type { ImsAuthParamsWithOptionalScopes } from "~/utils/auth/ims-scopes";
+
+type HttpMethod = "GET" | "POST" | "PUT" | "PATCH" | "DELETE";
+
+/**
+ * Type guard to check if the given auth object is an auth provider.
+ * @param auth The auth object to check.
+ */
+export function isAuthProvider(
+  auth:
+    | ImsAuthParams
+    | ImsAuthProvider
+    | ImsAuthParamsWithOptionalScopes
+    | IntegrationAuthParams
+    | IntegrationAuthProvider,
+): auth is ImsAuthProvider | IntegrationAuthProvider {
+  return "getHeaders" in auth;
+}
+
+/**
+ * Builds a before request hook for integration authentication.
+ * @param integrationAuth The integration authentication parameters or integration auth provider.
+ */
+export function buildIntegrationAuthBeforeRequestHook(
+  integrationAuth: IntegrationAuthParams | IntegrationAuthProvider,
+) {
+  const integrationAuthProvider = isAuthProvider(integrationAuth)
+    ? integrationAuth
+    : getIntegrationAuthProvider(integrationAuth);
+
+  return (request: KyRequest) => {
+    const headers = integrationAuthProvider.getHeaders(
+      request.method.toUpperCase() as HttpMethod,
+      request.url,
+    );
+
+    request.headers.set("Authorization", headers.Authorization);
+  };
+}
+
+/**
+ * Builds a before request hook for IMS authentication.
+ * @param imsAuth The IMS authentication parameters or IMS auth provider.
+ */
+export function buildImsAuthBeforeRequestHook(
+  imsAuth: ImsAuthParams | ImsAuthProvider,
+) {
+  const imsAuthProvider = isAuthProvider(imsAuth)
+    ? imsAuth
+    : getImsAuthProvider(imsAuth);
+
+  return async (request: KyRequest) => {
+    const headers = await imsAuthProvider.getHeaders();
+
+    request.headers.set("Authorization", headers.Authorization);
+    request.headers.set("x-api-key", headers["x-api-key"]);
+  };
+}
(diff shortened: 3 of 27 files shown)


packages-private/aio-commerce-lib-api/source/lib/commerce/helpers.ts
```diff
@@ -0,0 +1,137 @@
...
+  PaaSClientParams,
+  SaaSClientParams,
+} from "./types";
+
+const COMMERCE_SAAS_IMS_REQUIRED_SCOPES = [
+  ...BASE_IMS_REQUIRED_SCOPES,
+  "commerce.accs",
+];
+
+const DEFAULT_STORE_VIEW_CODE = "default";
```
pablomoreno61, line 26: "default" store view could not exist. If a store view code is not specified, then I would use "all" instead

packages-private/aio-commerce-lib-api/source/lib/commerce/helpers.ts
```diff
@@ -0,0 +1,137 @@
...
+const DEFAULT_API_VERSION = "v1";
```
jnatherley, line 27:
```suggestion
const DEFAULT_API_VERSION = "V1";
```

packages-private/aio-commerce-lib-api/source/lib/commerce/helpers.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Gets the Commerce URL for the given configuration and flavor.
+ * @param config The Commerce HTTP client configuration.
```
pablomoreno61, line 31: Why is the return type not specified in the docblock?

packages-private/aio-commerce-lib-api/source/lib/commerce/types.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { Options } from "ky";
+import type { ImsAuthParamsWithOptionalScopes } from "~/utils/auth/ims-scopes";
+
+type CommerceHttpClientConfigBase = {
+  /** The base URL of the Commerce API. */
+  baseUrl: string;
+
+  /**
+   * The store view code use to make requests to the Commerce API.
+   * @default "default"
```
pablomoreno61, line 15: "default" store view could result in a 404 error if it does not exist. I would use "all" instead

packages-private/aio-commerce-lib-api/source/lib/commerce/helpers.ts
pablomoreno61, line 27: Shouldn't we use the default constants from types.ts?

packages-private/aio-commerce-lib-api/source/lib/commerce/types.ts
```diff
@@ -0,0 +1,61 @@
...
+   */
+  storeViewCode?: string;
+
+  /**
+   * The version of the Commerce API to use. Currently only `v1` is supported.
+   * @default "v1"
+   */
+  version?: "v1";
```
jnatherley, line 23:
```suggestion
  version?: "V1";
```

The version will always be `V1` or some combination of `V(number)`.

packages-private/aio-commerce-lib-api/source/lib/io-events/helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+  const { auth, config, fetchOptions } = params;
+  const beforeRequestAuthHook = isAuthProvider(auth)
+    ? buildImsAuthBeforeRequestHook(auth)
+    : buildImsAuthBeforeRequestHook(
+        ensureImsScopes(auth, IO_EVENTS_IMS_REQUIRED_SCOPES),
+      );
+
+  const adobeIoBaseUrl =
+    config.environment === "stage"
+      ? "https://api-stage.adobe.io/events"
```
pablomoreno61, line 34: I'm not sure about adding "stage" here. I would like to know how other AIO libraries are doing this, as here we are exposing the name of an internal environment that is not customer-facing

packages-private/aio-commerce-lib-api/source/lib/io-events/types.ts
```diff
@@ -0,0 +1,24 @@
...
+import type { Options } from "ky";
+import type { ImsAuthParamsWithOptionalScopes } from "~/utils/auth/ims-scopes";
+
+/** Defines the configuration required to build an Adobe I/O HTTP client. */
+export type IoEventsHttpClientConfig = {
+  /** The consumer organization ID to use for the Adobe I/O API. */
+  consumerOrgId: string;
+
+  /** The environment to use for the Adobe I/O API. */
+  environment?: "prod" | "stage";
```
pablomoreno61, line 11: Same doubt here, as we refer to the stage as the internal Adobe environment, if I'm not mistaken.

packages-private/aio-commerce-lib-api/source/utils/auth/hooks.ts
```diff
@@ -0,0 +1,70 @@
...
+export function buildIntegrationAuthBeforeRequestHook(
+  integrationAuth: IntegrationAuthParams | IntegrationAuthProvider,
+) {
+  const integrationAuthProvider = isAuthProvider(integrationAuth)
+    ? integrationAuth
+    : getIntegrationAuthProvider(integrationAuth);
+
+  return (request: KyRequest) => {
+    const headers = integrationAuthProvider.getHeaders(
+      request.method.toUpperCase() as HttpMethod,
```
pablomoreno61, line 45: If this toUpperCase is required, then it should be done by the auth library, isn't it? @jnatherley

packages-private/aio-commerce-lib-api/source/utils/auth/hooks.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/**
+ * Builds a before request hook for IMS authentication.
+ * @param imsAuth The IMS authentication parameters or IMS auth provider.
+ */
+export function buildImsAuthBeforeRequestHook(
+  imsAuth: ImsAuthParams | ImsAuthProvider,
+) {
+  const imsAuthProvider = isAuthProvider(imsAuth)
```
pablomoreno61, line 60: This is the logic I would expect inside the auth library; otherwise, each library implementing auth will have to clone these methods

packages-private/aio-commerce-lib-api/source/utils/auth/ims-scopes.ts
```diff
@@ -0,0 +1,32 @@
+import type { ImsAuthParams } from "@adobe/aio-commerce-lib-auth";
+import type { SetOptional } from "type-fest";
+
+/** Common IMS required scopes. */
+export const BASE_IMS_REQUIRED_SCOPES = [
```
pablomoreno61, line 5: same for this. If this is a common need, then I think we should move it to the auth library

packages-private/aio-commerce-lib-api/source/lib/commerce/types.ts
jnatherley:
```suggestion
  version?: `V${number};
```
alternative with string interpolation

jnatherley (approved):

packages-private/aio-commerce-lib-api/source/lib/io-events/helpers.ts
jnatherley: I would prefer passing in the endpoint entirely, instead of having the concept of environment.

aio-lib works with the concept of environment, so it's abit painful to go against that.

packages-private/aio-commerce-lib-api/source/lib/commerce/helpers.ts
iivvaannxx: See [this](https://github.com/adobe/aio-commerce-sdk/blob/main/.github/DEVELOPMENT.md#documentation) in the `DEVELOPMENT.md`. This is something that I've been trying to predicate for a while but it's been difficult to make people understand it. Two reasons:

- Most of the time the language server can infer the return type directly, without needing to specify it. In most cases this should be just fine. In fact it's 100x times better than typing things as `object` or `any` (which I've seen countless times be done), as those return types basically erase all the type information from the return type and don't provide any type-safety at all.

- This is TypeScript, types in JSDoc are not needed when using TypeScript, as you have the type information built into the code, not in the doc comments. Having both is unnecessary and confusing, and although I am not sure of this, could be a potentially source of confusion for IDEs (what type information should the editor choose?).

If you're talking about a missing `@returns` without a type at all, I usually don't add them because 95% of the times they look exactly the same as the function description. Like in this case it would be:

```js
/**
 * Gets the Commerce URL for the given configuration and flavor.
 * @param config The Commerce HTTP client configuration
 * @returns The Commerce URL for the given configuration and flavor
 */
```

packages-private/aio-commerce-lib-api/source/lib/commerce/types.ts
iivvaannxx: See my point above

packages-private/aio-commerce-lib-api/source/lib/io-events/helpers.ts
iivvaannxx: This `environment` is something related to `aio-commerce-lib-auth` for IMS authentication using `aio-lib-ims`. [Their docs](https://github.com/adobe/aio-lib-ims?tab=readme-ov-file#ims-environment) discourage using `stage`  and reserves it for Adobe usage, but it's a correct value after all. I've also based this in [this code](https://github.com/adobe/aio-lib-events/blob/1d1676f228ca88edc53a6ca940137ecd73788ddc/src/index.js#L33) I saw for `aio-lib-events`.

packages-private/aio-commerce-lib-api/source/utils/auth/hooks.ts
iivvaannxx: Good point, I did this because the validation run by the auth library expects the `method` to be in uppercase.

emartinpalomas (approved):

aminakhyat (dismissed):

packages-private/aio-commerce-lib-api/source/lib/commerce/helpers.ts
pablomoreno61: I was thinking just in consistency, I saw @param and I missed @returns there, otherwise it looks like a method that does not return anything from the docblock perspective.

Being said that, I'm against creating comments that do not add value or explain what could be explained with a better naming. In this case, @param just says the same that the method signature is already saying, an the same for the method description. At the end, they become noise or potentially forgotten

packages-private/aio-commerce-lib-api/source/lib/io-events/helpers.ts
pablomoreno61: I would discourage anyone to expose our internal URLs publicly. See also this thread as an example of how misleading it is => https://experienceleaguecommunities.adobe.com/t5/adobe-developer-questions/where-to-find-api-key-whiltelisted-at-adobe-io/td-p/585453

packages-private/aio-commerce-lib-api/source/utils/auth/hooks.ts
pablomoreno61: The developer needs to understand the implementation details of the getHeaders() method to understand that you can get the token as headers.Authorization and the api key as headers["x-api-key"]. It's not very easy to understand IMO

packages-private/aio-commerce-lib-api/source/utils/auth/ims-scopes.ts
pablomoreno61: Our library should be specific to Commerce, shouldn't it?

Never mind, having the scopes here is fine, as they are fulfilling the purpose of the API calls

packages-private/aio-commerce-lib-api/source/lib/io-events/helpers.ts
jnatherley: > > I would prefer passing in the endpoint entirely, instead of having the concept of environment.
> 
> You mean a `baseUrl`?

yep, but again not possible since we're using an underlying library that has the concept of environment

jnatherley (dismissed):

packages-private/aio-commerce-lib-api/source/lib/commerce/helpers.ts
iivvaannxx: Discussed offline, return types are inferred by the language. Unless you need a `@returns` that doesn't "collide" with the method description most of the time is redundant. The same could be argued for parameters (they are also inferred by the language), but in this case they can use an actual description that doesn't collide with anything.

packages-private/aio-commerce-lib-api/source/lib/commerce/helpers.ts
iivvaannxx: Discussed offline, types are only available at compile type. The check that this comment was referring to (not referenced correctly) was at runtime, so this is not an option.

packages-private/aio-commerce-lib-api/source/lib/io-events/types.ts
iivvaannxx: > ims-lib implements this environment flag

True, nothing we can do for now

packages-private/aio-commerce-lib-api/source/utils/auth/hooks.ts
iivvaannxx: Discussed offline, although this is more a comment targeted to `lib-auth`, not this PR.

packages-private/aio-commerce-lib-api/source/utils/auth/ims-scopes.ts
iivvaannxx: Discussed offline, actually there's not much "specific" in Commerce regarding authentication/authorization. `lib-auth` is only a wrapper over the auth mechanisms, there's nothing more specific that is Commece or platform-dependant.

packages-private/aio-commerce-lib-api/source/lib/commerce/helpers.ts
iivvaannxx: Addressed in [`e404f8d`](https://github.com/adobe/aio-commerce-sdk/commit/e404f8d338751553a910d22097a19fe8ab40a0e6).

packages-private/aio-commerce-lib-api/source/lib/io-events/helpers.ts
iivvaannxx: Addressed in [`e404f8d`](https://github.com/adobe/aio-commerce-sdk/commit/e404f8d338751553a910d22097a19fe8ab40a0e6).

pablomoreno61 (approved):

aminakhyat (approved):

PR #83, opened 2025-09-09
CEXT-4993/CEXT-5217: `@adobe/aio-commerce-lib-events` Add Tests and README

iivvaannxx:
## Description

This PR adds tests for the `@adobe/aio-commerce-lib-events` library and a README and usage guides. It also updates the root README with the current state of the SDK libraries as it was out of sync. 

## Test Strategy

To test the API calls I followed [this guide](https://vitest.dev/guide/mocking.html#requests) in the Vitest documentation. The whole API clients are tested using a mock server in the form of an integration test, to ensure that validation, authentication headers, hooks and transformations work all together to produce the expected result.

For now this could be enough, but the library could use better testing around input validation for each API call and also some extra integration test to ensure that the API calls have the correct payloads expected by the Commerce and I/O APIs.

There are also some unit tests for the functionality that is not related to API invocations.


packages/aio-commerce-lib-events/test/integration/commerce-paas-events-api.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      if (!capture.headers) {
+        expect.fail("Captured headers are null");
+        return;
+      }
+
+      expect(capture.headers.get("X-Custom-Header")).toBe("test-value");
+    });
+
+    if (payload.hasInputValidation) {
```
jnatherley, line 116: Again, but this time we are conditionally adding a test on whether a conditional passes. Could you explain your reasoning here?

👀

jnatherley (approved):
LGTM. However, i would like an explanation around the usage of `hasInputValidation` and `hasCamelCaseTransformer`.

packages/aio-commerce-lib-events/test/fixtures/io-events-api-payloads.ts
```diff
@@ -0,0 +1,189 @@
...
+          label: "Event 1",
+          description: "Event 1 description",
+          eventCode: "event-1",
+        },
+        fetchOptions,
+      );
+    },
+
+    hasInputValidation: true,
+    hasCamelCaseTransformer: true,
```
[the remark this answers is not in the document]
iivvaannxx: Not always! But only a couple of situations they have false on them. Example cases are:

1. `hasInputValidation: false`: If there are no input params at all (for example in `getAllEventProviders` for Commerce API)
2. `hasCamelCaseTransformer: false`: If an API call returns a non-object response, like a `boolean` or `void`.

packages/aio-commerce-lib-events/test/integration/commerce-paas-events-api.test.ts
iivvaannxx: Explained in the above comments, if it's not because of this flags the tests fail because the logic that is tested/expected doesn't run.

> but this time we are conditionally adding a test 
like the above no?

emartinpalomas (approved):

PR #108, opened 2025-11-03
INTERNAL: Include runtime actions' templates in npm package

aminakhyat:
## Description

Template files were not being included in the built npm package, causing the action generation functionality to fail.

## Solution

- **Added `copy` configuration to `tsdown.config.ts`** to include template files in both CJS and ESM build outputs
- **Verified templates are copied correctly** to both `dist/cjs/hooks/templates/` and `dist/es/hooks/templates/`

### DIFF
--- a/.changeset/cyan-mice-knock.md
+++ b/.changeset/cyan-mice-knock.md
@@ -1,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-config": minor
+---
+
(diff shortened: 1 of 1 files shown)


iivvaannxx (approved):
You need to add a Changeset to trigger a re-release! Probably a patch bump will do

emartinpalomas (approved):

PR #113, opened 2025-11-04
CEXT-5447: Add description as optional field to schema

aminakhyat:
## Description

Support optional `description` field in configuration schema validation


packages/aio-commerce-lib-config/source/modules/schema/types.ts
```diff
@@ -22,6 +22,7 @@ export interface ConfigSchemaField {
   label: string;
   options?: ConfigSchemaOption[];
   default?: string;
+  description?: string;
 }
```
iivvaannxx: Just noticed that you are creating this type manually. You can actually let `valibot` infer that type for you. It's one of the core features of modern schema libraries.

With this you'll only have to deal with the schema, and the types will be inferred automatically. In fact it will do it in a more robust way, as thanks to the `v.variant` schema it will create a [discriminated union](https://www.totaltypescript.com/discriminated-unions-are-a-devs-best-friend) that will tell TypeScript, hey when `type` is `list`, the object will have a `options` property, but if it's `text` that property won't exist.


```suggestion
import * as v from "valibot"
import { VariantTypeSchema } from "./validation/schema"

export type ConfigSchemaField = v.InferInput<typeof VariantTypeSchema>;
export type ConfigSchemaOption = Extract<ConfigSchemaField, { type: "list" }>["options"][number];
```

In

PR #369, opened 2026-04-01
 CEXT-6029: Add uninstallation flow on app unassociation

oshmyheliuk:
## Description

https://jira.corp.adobe.com/browse/CEXT-6029

Added uninstallation flow on app unassociation.

- New `POST /installation/uninstallation` endpoint starts an async uninstallation workflow and returns 202 Accepted with the initial state.
- New `GET /installation/uninstallation` endpoint returns the current uninstallation status (200 with state or 204 if not started).
- New `POST /installation/uninstallation/execution` internal endpoint executes the uninstallation workflow steps asynchronously.
- New `DELETE /installation/uninstallation` endpoint clears the uninstallation state only, without triggering any offboarding.
- Removed `DELETE /installation` endpoint. Uninstallation is now handled exclusively through the `/installation/uninstallation` sub-path.
- Added uninstall method to the each step that requires uninstallation logic, including the Admin UI SDK, eventing, webhooks, and custom installation steps.
- `defineCustomInstallationStep` now accepts an object with `install` and optional `uninstall` handlers in addition to a plain function, enabling per-step cleanup logic.

Added delete operations for events cleanup during uninstallation.

- Commerce API: `deleteEventProvider` deletes an event provider by ID from a Commerce instance.
- Commerce API: `deleteEventSubscription` unsubscribes from an event in a Commerce instance.
- IO Events API: `deleteEventProvider` deletes an event provider from Adobe I/O Events.
- IO Events API: `deleteEventMetadataForProvider` deletes event metadata for a specific event code and provider.

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/source/management/installation/workflow/runner.ts
```diff
@@ -64,6 +67,9 @@ export type ExecuteWorkflowOptions = {
   hooks?: InstallationHooks;
 };
 
+/** Execution mode: "run" for installation, "uninstall" for uninstallation. */
+type ExecutionMode = "run" | "uninstall";
```
obarcelonap: `install | uninstall`? I know the method is called `run` but it is a bit strange.

packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts
```diff
@@ -67,15 +67,26 @@ function createCustomScriptStep(scriptConfig: CustomInstallationStep): AnyStep {
       // Only support default export
       if (typeof scriptModule !== "object" || !("default" in scriptModule)) {
         throw new Error(
-          `Script ${script} must export a default function. Use defineCustomInstallationStep helper.`,
+          `Script ${script} must export a default function or object. Use defineCustomInstallationStep helper.`,
```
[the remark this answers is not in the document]
oshmyheliuk: It would be breaking changes, as now we are using "script" key, unless we add two more `"install"` and `"uninstall"`, and keep the `script` as well.
```
  installation: {
    customInstallationSteps: [
      {
        script: "./scripts/setup-database.js",
        name: "Setup Database",
        description:
          "Create database collections and indexes for approval requests and execution logs",
      },
    ],
  },
```
The approach in this PR keeps backward compatibility by keeping the current approach and supporting a new approach of objects with `install/uninstall`.
What downsides do you see in using objects with `install/uninstall` functions?

packages/aio-commerce-lib-app/source/management/installation/workflow/runner.ts
oshmyheliuk: Yeah, I don't know why we haven't named `install` instead of `run` from the very beginning. Probably, we can rename it if it's not publicly exposed. But if so, I would prefer to do it as refactoring in a separate ticket, as this PR is too large.

packages/aio-commerce-lib-app/source/management/installation/webhooks/branch.ts
```diff
@@ -31,12 +32,23 @@ const subscriptionsStep = defineLeafStep({
     label: "Create Subscriptions",
     description: "Creates webhook subscriptions in Adobe Commerce",
   },
+  uninstallMeta: {
```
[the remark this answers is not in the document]
oshmyheliuk: Yes, I agree, we need to think about localization. We can extract all hardcoded text in a separate PR.

packages/aio-commerce-lib-app/source/actions/installation.ts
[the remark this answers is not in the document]
oshmyheliuk: Yes, It makes sense. I will update the PR it would also require changes in the App Management itself.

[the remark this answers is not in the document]
oshmyheliuk: It was actually my initial approach to introduce a new endpoint, but I remembered why I had not done it.
So, in app management, we are checking endpoints like this:
```
export function getExtensibilityEndpoints(extensionPointValues?: {
  workerProcess?: {href?: string}[];
}): ExtensibilityEndpoints {
  const processes = extensionPointValues?.workerProcess ?? [];
  const hrefs = processes.map((process: {href?: string}) => process?.href).filter(Boolean) as string[];

  return {
    extensionPoint: EXTENSION_POINT_EXTENSIBILITY,
    endpoints: {
      installation: hrefs.find((href: string) => href.endsWith('/app-management/installation')) ?? undefined,
      getAppConfig: hrefs.find((href: string) => href.endsWith('/app-management/app-config')) ?? undefined
    }
  };
}
```
So, to check if `/uninstallation` exists, it had to be added as a new runtime action. It's going against our strategy to have as few runtime actions as possible for quicker deployment.
So, now I'm not sure about this change, as it will introduce a new runtime action and slow the deployment.

packages/aio-commerce-lib-app/source/management/installation/workflow/step.ts
```diff
@@ -124,6 +127,16 @@ export type LeafStep<
     config: TConfig,
     context: ValidationExecutionContext<TStepCtx>,
   ) => ValidationIssue[] | Promise<ValidationIssue[]>;
+
+  /**
```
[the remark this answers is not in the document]
oshmyheliuk: Do you mean `uninstall` handler? This will be used only by us; it's not exposed to external developers. 
I would prefer to make it optional, as maybe some steps will not have anything to uninstall.

packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts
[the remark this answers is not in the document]
obarcelonap: In order to keep backwards compatibility, what if we add `install` and `uninstall` props and we keep `script` but adding a deprecation notice referring `install` should be used instead?
code can be adapted to check if `install` present use it otherwise fallback to `script`.
What I like about current approach (default function) is that forces single responsibility principle per file by design, which I think is good in order to have cleaner code bases.

packages/aio-commerce-lib-app/source/management/installation/workflow/runner.ts
obarcelonap: About PR being large, maybe would help to split the changes in two prs:
* one for aio-commerce-lib-events, these are just changes in the client that we can get quickly merged.
* another one for aio-commerce-lib-app that it will take more time since we have quite some discussions open.

packages/aio-commerce-lib-app/source/management/installation/webhooks/branch.ts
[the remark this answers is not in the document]
obarcelonap: I created this ticket to address the localization thingy [CEXT-6091](https://jira.corp.adobe.com/browse/CEXT-6091)

packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts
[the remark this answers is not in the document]
oshmyheliuk: I don't like that for single step user need to create two separate files. One script - one custom step, sounds better to me. 
Also, most probably, install/uninstall would share similar logic. With different files, it would require an additional file for shared logic.

packages/aio-commerce-lib-app/source/actions/installation.ts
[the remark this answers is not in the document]
oshmyheliuk: I updated endpoints according to our discussion in Slack https://magento.slack.com/archives/C0A3SFDC6HY/p1775758200028499

packages/aio-commerce-lib-app/source/management/installation/workflow/runner.ts
oshmyheliuk: My point is that what you mentioned is not related to changes in this PR. 
If the team agrees on renaming, we can do it in a separate PR as a separate task. Let's add this item to the technical refinement.

packages/aio-commerce-lib-app/source/management/installation/webhooks/branch.ts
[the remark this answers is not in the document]
obarcelonap: If you merge this pr and we release `uninstallMeta` will be part of the API so removing the hardcoded texts will be breaking change. How do you plan to manage it?

packages/aio-commerce-lib-app/source/management/installation/workflow/runner.ts
obarcelonap: `ExecutionMode` type is part of this PR. Actually I believe you never check for `mode === 'run'` so you can just probably change `install` (instead of `run`) in `ExecutionMode` and code will work (not verified).

obarcelonap: Have you considered splitting the PR? I think it would help reduce the cognitive load during review.

packages/aio-commerce-lib-app/source/management/installation/workflow/step.ts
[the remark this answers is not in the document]
obarcelonap: I can live with that.

packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts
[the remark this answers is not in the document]
obarcelonap: I believe the simple the better and supporting `object` adds complexity which I don't think is needed. Having different files might be verbose but also more clear.
In any case, not a blocker. Your call.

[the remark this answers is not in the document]
oshmyheliuk: 1. Having different files means that we need to update the parser and code generator to add them to the include before the app is deployed; they are not included automatically
2. In the future, we can add new logic to the custom scripts (like validate, post-install), creating and supporting a new file for each action will be more complex for the developers, and for us, it's much easier to add a new optional method.
So, I'm strongly for having one script file per configuration action.

packages/aio-commerce-lib-app/source/management/installation/workflow/runner.ts
oshmyheliuk: I haven't considered it, as I was testing and developing both `eventing` and `app` packages together. Changes in the `app` might lead to changes in `events` and vice versa

oshmyheliuk: > ExecutionMode type is part of this PR. Actually I believe you never check for mode === 'run' so you can just probably change install (instead of run) in ExecutionMode and code will work (not verified).

I thought that you were asking me to rename all `run`  methods in the defined step to `install`. Anyway, at the moment we have `run` for installation and `uninstall` for uninstallation. So it makes sense to keep `run` in the ExecutionMode for now, if we decide to rename `run` to `install` it would be changed in the ExecutionMode as well.

packages/aio-commerce-lib-app/source/management/installation/webhooks/branch.ts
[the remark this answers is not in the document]
oshmyheliuk: It will be managed in the same way as for the already released hardcoded meta. 
Removing it won't be a breaking change, as for uninstallation, the step data is returning in the same format, as a meta field
```
    meta:
      mode === "uninstall" && step.uninstallMeta
        ? step.uninstallMeta
        : step.meta,
```

obarcelonap (approved):

.changeset/green-beds-dance.md
```diff
@@ -0,0 +1,13 @@
...
+
+Add uninstallation flow on app unassociation.
+
+- New `POST /installation/uninstallation` endpoint starts an async uninstallation workflow and returns 202 Accepted with the initial state.
+- New `GET /installation/uninstallation` endpoint returns the current uninstallation status (200 with state or 204 if not started).
+- New `POST /installation/uninstallation/execution` internal endpoint executes the uninstallation workflow steps asynchronously.
+- New `DELETE /installation/uninstallation` endpoint clears the uninstallation state only, without triggering any offboarding.
+- Removed `DELETE /installation` endpoint. Uninstallation is now handled exclusively through the `/installation/uninstallation` sub-path.
+- Added uninstall method to each step that requires uninstallation logic, including the Admin UI SDK, eventing, webhooks, and custom installation steps.
+- `defineCustomInstallationStep` now accepts an object with `install` and optional `uninstall` handlers in addition to a plain function, enabling per-step cleanup logic.
```
iivvaannxx: All of these are implementation details that a user of the library doesn't care much about (in my opinion). I'd remove all this information and keep it simple, while also avoiding mentioning concepts related to App Management, as this library is not necessarily used with it.


```suggestion
Add uninstallation flow for both native steps and custom steps. `defineCustomInstallationStep` now accepts an object with `install` and an optional `uninstall` handler in addition to a plain function, enabling per-step cleanup logic.
```

.changeset/swift-owls-fly.md
```diff
@@ -0,0 +1,10 @@
+---
+"@adobe/aio-commerce-lib-events": minor
+---
+
+Add delete operations for events cleanup during uninstallation.
```
iivvaannxx: Same point as above, this Changeset description is tied to concepts of App Management, which this library is independent of. I'd suggest (keeping the individual new operation descriptions):
```suggestion
Add delete operations for both the I/O Events and Commerce Eventing APIs.
```

packages/aio-commerce-lib-app/source/management/installation/admin-ui-sdk/branch.ts
```diff
@@ -29,9 +29,16 @@ const registerExtensionStep = defineLeafStep({
     label: "Register Extension",
     description: "Registers the Admin UI SDK extension in Adobe Commerce",
   },
+  uninstallMeta: {
+    label: "Unregister Extension",
+    description: "Removes the Admin UI SDK extension from Adobe Commerce",
+  },
```
iivvaannxx: I think `meta` is generic enough for this to be in there. Instead of having multiple "metas" I suggest creating a `uninstall` key nested in the `meta` itself.

packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts
```diff
@@ -67,15 +67,26 @@ function createCustomScriptStep(scriptConfig: CustomInstallationStep): AnyStep {
...
+
+      if (typeof defaultExport === "function") {
+        runFunction = defaultExport;
+      } else if (
+        typeof defaultExport === "object" &&
+        defaultExport !== null &&
+        typeof (defaultExport as Record<string, unknown>).install === "function"
+      ) {
+        runFunction = (defaultExport as Record<string, unknown>).install;
+      }
```
iivvaannxx, line 104: This can be cleaned up a bit with Typescript type-assertions and a custom helper.

```ts
// This for the object types, another might be needed for the standalone install handler
function isCustomInstallationStepDefinition(obj: unknown): obj is CustomInstallationStepDefinition {}
```

packages/aio-commerce-lib-app/source/management/installation/custom-installation/define.ts
```diff
@@ -25,36 +25,64 @@ export type CustomInstallationStepHandler<TResult = unknown> = (
...
+  definition: CustomInstallationStepDefinition<TResult>,
+): CustomInstallationStepDefinition<TResult>;
+export function defineCustomInstallationStep<TResult = unknown>(
+  handlerOrDefinition:
+    | CustomInstallationStepHandler<TResult>
+    | CustomInstallationStepDefinition<TResult>,
+):
+  | CustomInstallationStepHandler<TResult>
+  | CustomInstallationStepDefinition<TResult> {
+  return handlerOrDefinition;
```
iivvaannxx: Does returning the union directly produce any type-error that the overload does not? Overloads in TypeScript are kind of verbose.

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
```diff
@@ -658,3 +659,227 @@ export async function onboardCommerceEventing(
...
+  const instanceId = generateInstanceId(
+    metadata,
+    provider,
+    appData.workspaceId,
+  );
+  const { registrations } = existingData;
+
+  const providerData = existingData.providersWithMetadata.find(
+    (p) => p.instance_id === instanceId,
+  );
```
iivvaannxx: This will most probably not work for IDs before the workspace ID change right? Not an issue, just want to confirm if we are fine with this or we want to have a fallback.

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
```diff
@@ -658,3 +659,227 @@ export async function onboardCommerceEventing(
...
+    logger.info(
+      `No I/O Events provider found with instance ID "${instanceId}", skipping offboarding.`,
+    );
+    return;
+  }
+
+  // The getAllRegistrations list endpoint does not populate events_of_interest,
+  // so we reconstruct the expected registration names the same way as during installation.
+  const actionEventsMap = groupEventsByRuntimeActions(events);
+  const clientId = runtimeParams.AIO_COMMERCE_AUTH_IMS_CLIENT_ID;
```
iivvaannxx: Picky: This is not re-used, if it's for verbosity I'm okay with, but maybe we can just use the `params` ref directly.

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
```diff
@@ -658,3 +659,227 @@ export async function onboardCommerceEventing(
...
+ * 2. All event metadata entries on the provider.
+ * 3. The provider itself.
+ *
+ * This is the reverse of {@link onboardIoEvents} and is called during uninstall.
+ * All deletion errors are caught and logged so that uninstall remains best-effort.
+ *
+ * @param params - Configuration identifying the provider to offboard.
+ * @param existingData - Current I/O Events data (providers and registrations).
+ */
+export async function offboardIoEvents(
```
iivvaannxx: Not a blocker, but I prefer if we split this a bit, it's quite large, similar to how we onboard.

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
```diff
@@ -658,3 +659,227 @@ export async function onboardCommerceEventing(
...
+ *
+ * Subscriptions are matched by their namespaced name, which is deterministic and built the
+ * same way as during {@link onboardCommerceEventing}. The provider is matched by its
+ * `instance_id`. Missing subscriptions or providers are silently skipped. All errors are
+ * caught and logged so that uninstall remains best-effort.
+ *
+ * @param params - Configuration identifying the provider and its events to offboard.
+ * @param existingData - Current Commerce eventing data (providers and subscriptions).
+ */
+export async function offboardCommerceEventing(
```
iivvaannxx: Same here about splitting. Again, not a blocker.

packages/aio-commerce-lib-app/test/unit/management/installation/admin-ui-sdk/branch.test.ts
iivvaannxx, line 1: #375 renamed `index.test.ts` to be `branch.test.ts` (small change as barrel files are not meant to be tested)

packages/aio-commerce-lib-app/test/unit/management/installation/events/helpers.test.ts
```diff
@@ -784,3 +785,125 @@ describe("onboardCommerceEventing", () => {
     expect(logger.error).toHaveBeenCalledWith(expect.any(String));
   });
 });
+
+describe("offboardIoEvents", () => {
```
iivvaannxx, line 680: Picky: Create a helper to create an offboarding scenario that you can re-use, like we test the onboarding, instead of re-creating in each test.

packages/aio-commerce-lib-app/test/unit/management/installation/events/branch.test.ts
iivvaannxx, line 1: Same thing on this being renamed to `branch.test.ts`. You might get a conflict here.

packages/aio-commerce-lib-app/test/unit/management/installation/root.test.ts
```diff
@@ -60,3 +63,52 @@ describe("createRootInstallationStep", () => {
...
+    // The custom installation step should have children (script steps directly)
+    if (
+      customInstallationStep.type === "branch" &&
+      customInstallationStep.children
+    ) {
+      expect(customInstallationStep.children.length).toBe(2);
+      expect(customInstallationStep.children[0].name).toBe("demoSuccess");
+      expect(customInstallationStep.children[0].meta.label).toBe(
+        "Demo Success",
+      );
```
iivvaannxx: I know this is for type safety, but the correct way would be to `expect.assert` and call the `isBranchStep` helper. It should produce the same result (TypeScript won't complain). There are some examples in other test files.

packages/aio-commerce-lib-events/source/commerce/api/event-providers/endpoints.ts
```diff
@@ -98,3 +100,32 @@ export async function createEventProvider(
...
+): Promise<void> {
+  const validatedParams = parseOrThrow(EventProviderDeleteParamsSchema, params);
+
+  return httpClient
+    .delete(
+      `eventing/eventProvider/${validatedParams.provider_id}`,
+      fetchOptions,
+    )
+    .then((_res) => {
+      // We set this `then` to make the response type `void`
```
iivvaannxx, line 129: Would it make sense to return `true/false` depending on success or not?

iivvaannxx (approved):

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
oshmyheliuk: Yes, looks like it won't work, as we already made that incompatible change. The new instance ID will be different from previously created instance IDs in the older SDK.

packages/aio-commerce-lib-app/source/management/installation/admin-ui-sdk/branch.ts
oshmyheliuk: Agree, I changed the meta object to the 
```
export type StepMeta = {
  install: StepMetaInfo;
  uninstall?: StepMetaInfo;
};
```
As having `run` (which is actually install) and `uninstall` didn't sound good, I also renamed `run` -> `install` in the step definition.
@obarcelonap already raised this question in comments in this PR, so after splitting the meta, it made sense to make renaming.
Such a change eliminates the naming asymmetry. If there are strong objections regarding this, we can discuss, and I will rename it back to `run`

oshmyheliuk: Changed in this [commit](https://github.com/adobe/aio-commerce-sdk/pull/369/changes/c3767d88d6f9f1fb69112643eb8c9156b11714cd). Ignore the wrong commit message.

packages/aio-commerce-lib-app/source/management/installation/custom-installation/define.ts
oshmyheliuk: I removed overloads.

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
oshmyheliuk: For me, it's more readable when the clientId is used in the filter method on the next lines. But yes, it can be used directly, just a style preference.

packages/aio-commerce-lib-app/test/unit/management/installation/admin-ui-sdk/branch.test.ts
oshmyheliuk: Renamed for admin-ui-sdk as [well](https://github.com/adobe/aio-commerce-sdk/pull/369/changes/6a00d5d23c75d70acc2b9cb23ff6bbc191da39ad)

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
oshmyheliuk: Done in https://github.com/adobe/aio-commerce-sdk/pull/369/changes/6dbca237d9cae66be013ee56e8682e675990b4b9

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
oshmyheliuk: Done in https://github.com/adobe/aio-commerce-sdk/pull/369/changes/6dbca237d9cae66be013ee56e8682e675990b4b9

packages/aio-commerce-lib-app/test/unit/management/installation/events/helpers.test.ts
oshmyheliuk: Done in https://github.com/adobe/aio-commerce-sdk/pull/369/changes/6dbca237d9cae66be013ee56e8682e675990b4b9

packages/aio-commerce-lib-app/test/unit/management/installation/root.test.ts
oshmyheliuk: Updated to use `isBranchStep` https://github.com/adobe/aio-commerce-sdk/pull/369/changes/e0a14ccb61833e85d106b3e369814e74bea5d9eb

packages/aio-commerce-lib-events/source/commerce/api/event-providers/endpoints.ts
oshmyheliuk: Callers of this method don't need a boolean to decide what to do next. It's already wrapped in try/catch, so I would leave it as it is.

packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts
oshmyheliuk: Done https://github.com/adobe/aio-commerce-sdk/pull/369/changes/dd772232d515713bd302e7f8db6b015b7293d808

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
iivvaannxx: Would it make sense to adding that fallback? Should be only an OR condition in the `find` right?

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
iivvaannxx: Then it's good, you can keep it.

packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts
```diff
@@ -20,11 +20,31 @@ import type {
...
   ExecutionContext,
 } from "#management/installation/workflow/step";
 
+function isCustomInstallationStepDefinition(
+  obj: unknown,
+): obj is CustomInstallationStepDefinition {
+  return (
+    typeof obj === "object" &&
+    obj !== null &&
+    typeof (obj as Record<string, unknown>).install === "function"
```
iivvaannxx: This is good, just not a fan of using `as` unless other options are a no-go. I would do:

 
```suggestion
    "install" in obj && typeof obj.install === "function"
```

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
```diff
@@ -623,3 +627,328 @@ export async function onboardCommerceEventing(
...
+      logger.info(
+        `Deleted registration "${registration.name}" (ID: ${registration.id}).`,
+      );
+    } catch (error) {
+      logger.warn(
+        `Failed to delete registration "${registration.name}" (ID: ${registration.id}): ${stringifyError(error)}. Continuing uninstall.`,
+      );
+    }
+  }
+}
```
iivvaannxx, line 695: This can be done in the future if we want to optimize API calls, not a blocker. 

I think it's best to use "individual" helpers (e.g. `deleteIoEventRegistration`) and do the `for` outside, so that if we want to use `Promise.all` for parallel requests, we have our custom logic around the call (in this case logging, but can be anything else) already accounted in. With this setup we would probably need to keep track of an array of manually constructed promises (or at least that's how I picture it in my head right now).

Might make testing a bit easier also.

iivvaannxx (approved):

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
oshmyheliuk: I don't think it's critical, as not many apps have been installed. 
We can add a fallback, but the problem is that we don't have an old version of `generateInstanceId` function unless we create something like `generateInstanceIdOld` with the previous logic.

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
oshmyheliuk: Agree, it can be improved, let's do it in a separate PR.

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
oshmyheliuk: @iivvaannxx I added the old function, marked it as deprecated, and now I am checking against bot instance IDs, old and new versions.

PR #387, opened 2026-04-13
CEXT-6071: Add PATCH endpoint for partial config updates

obarcelonap:
## Description
  - Add `PATCH ` handler to the config runtime action accepting partial payloads and null values for unsetting fields
  - Deprecate `PUT ` (still functional, returns `Deprecation: true` header)
  - Widen `sanitizeRequestEntries` to accept nullable config entries
  - Add unit tests covering PATCH semantics, null pass-through, masked-password behaviour, and deprecation header

## Types of changes

## Checklist:

### DIFF
--- a/.changeset/quiet-dragons-listen.md
+++ b/.changeset/quiet-dragons-listen.md
@@ -1,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": minor
+---
+
+Adding that PATCH endpoint and deprecating the PUT
--- a/.changeset/quiet-dragons-listen.md
+++ b/.changeset/quiet-dragons-listen.md
@@ -2,4 +2,5 @@
 "@adobe/aio-commerce-lib-app": minor
 ---
 
-Adding that PATCH endpoint and deprecating the PUT
+Adding new PATCH endpoint for configuration which allows to set configuration values.
+Deprecating PUT endpoint in favor of PATCH.
--- a/.changeset/quiet-dragons-listen.md
+++ b/.changeset/quiet-dragons-listen.md
@@ -2,5 +2,4 @@
 "@adobe/aio-commerce-lib-app": minor
 ---
 
-Adding new PATCH endpoint for configuration which allows to set configuration values.
-Deprecating PUT endpoint in favor of PATCH.
+Adding new PATCH endpoint for configuration which allows to partially set configuration values. Deprecating PUT endpoint in favor of PATCH.
--- a/.changeset/yummy-beans-jump.md
+++ b/.changeset/yummy-beans-jump.md
@@ -1,0 +1,6 @@
+---
+"@adobe/aio-commerce-lib-app": minor
+"@adobe/aio-commerce-lib-config": minor
+---
+
+Add PATCH handler to the config runtime action accepting partial payloads and null values for unsetting fields
--- a/.changeset/yummy-beans-jump.md
+++ b/.changeset/yummy-beans-jump.md
@@ -1,6 +1,5 @@
 ---
-"@adobe/aio-commerce-lib-app": minor
 "@adobe/aio-commerce-lib-config": minor
 ---
 
-Add PATCH handler to the config runtime action accepting partial payloads and null values for unsetting fields
+Adding support for config unsetting
--- a/.changeset/yummy-beans-jump.md
+++ b/.changeset/yummy-beans-jump.md
@@ -2,4 +2,4 @@
 "@adobe/aio-commerce-lib-config": minor
 ---
 
-Adding support for config unsetting
+Add support for unset configuration values.
--- a/packages/aio-commerce-lib-app/source/actions/config.ts
+++ b/packages/aio-commerce-lib-app/source/actions/config.ts
@@ -134,6 +134,12 @@ router.put("/", {
     logger.debug(`Setting configuration with scope id: ${req.body.scopeId}`);
     const { scopeId, config } = req.body;
 
+    const validatedSchema = validateCommerceAppConfigDomain(
+      configSchema,
+      "businessConfig.schema",
+    );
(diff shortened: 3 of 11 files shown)


packages/aio-commerce-lib-app/source/actions/config.ts
```diff
@@ -143,6 +147,38 @@ router.put("/", {
       },
     );
 
+    result.config = filterPasswordFields(configSchema, result.config);
+    return ok({
+      body: result,
+      headers: { "Cache-Control": "no-store", Deprecation: "true" },
```
iivvaannxx: Maybe we should use a custom header syntax? `x-deprecated`?
```suggestion
      headers: { "Cache-Control": "no-store", "x-deprecated": "true" },
```

obarcelonap: deprecation header is standard since RFC9745, adjusted the value to be compliant though.

packages/aio-commerce-lib-app/source/actions/config.ts
```diff
@@ -143,6 +147,41 @@ router.put("/", {
...
+  }),
+
+  handler: async (req, ctx) => {
+    const { logger, rawParams } = ctx;
+    const { configSchema } = rawParams;
+
+    logger.debug(`Patching configuration with scope id: ${req.body.scopeId}`);
+    const { scopeId, config } = req.body;
+
+    const result = await setConfiguration({ config }, byScopeId(scopeId), {
```
oshmyheliuk, line 193: Does it work without initializing the configuration? Our Docs says that `initialize` must be executed before setting the configuration.

```
You must initialize the library with your configuration schema before using any configuration functions. Calling configuration functions without initialization will throw an error.
```

obarcelonap: good catch, it seems the issue was already there because PUT did not called the `initialize` function. Fixed.

iivvaannxx: If it didn't call `initialize` and it worked, I think we might need to clarify the docs. Could it be that we only throw in some cases?

packages/aio-commerce-lib-config/test/unit/config-utils.test.ts
```diff
@@ -0,0 +1,267 @@
...
+        name: "currency",
+        value: "USD",
+        origin: { code: "global", level: "global" },
+      },
+      {
+        name: "locale",
+        value: "en_US",
+        origin: { code: "global", level: "global" },
+      },
+    ];
```
iivvaannxx: Can we use any of the ones in `fixtures`? If not, maybe add one there?

.changeset/yummy-beans-jump.md
```diff
@@ -0,0 +1,6 @@
+---
+"@adobe/aio-commerce-lib-app": minor
+"@adobe/aio-commerce-lib-config": minor
+---
+
+Add PATCH handler to the config runtime action accepting partial payloads and null values for unsetting fields
```
iivvaannxx: The changeset should describe separate changes. You need two of them. Otherwise you would be creating an update for `lib-config` that says "Add PATCH handler", when `lib-config` doesn't even expose any runtime action.

This PR is modifying two separate packages in two different ways (two Changesets):

- `lib-config`: Adding support for config unsetting
- `lib-app`: Adding that PATCH endpoint and deprecating the `PUT`

iivvaannxx (approved):

.changeset/yummy-beans-jump.md
obarcelonap: done

packages/aio-commerce-lib-config/test/unit/config-utils.test.ts
obarcelonap: done

.changeset/quiet-dragons-listen.md
```diff
@@ -0,0 +1,6 @@
+---
+"@adobe/aio-commerce-lib-app": minor
+---
+
+Adding new PATCH endpoint for configuration which allows to set configuration values.
+Deprecating PUT endpoint in favor of PATCH.
```
iivvaannxx:
```suggestion
Adding new PATCH endpoint for configuration which allows to partially set configuration values. Deprecating PUT endpoint in favor of PATCH.
```

packages/aio-commerce-lib-app/source/actions/config.ts
```diff
@@ -143,6 +153,47 @@ router.put("/", {
...
+    const { configSchema } = rawParams;
+
+    logger.debug(`Patching configuration with scope id: ${req.body.scopeId}`);
+    const { scopeId, config } = req.body;
+
+    const validatedSchema = validateCommerceAppConfigDomain(
+      configSchema,
+      "businessConfig.schema",
+    );
+    initialize({ schema: validatedSchema });
```
oshmyheliuk, line 191: The PUT endpoint has the logic to filter masked fields
```
    // The UI sent it to us as a masked value, which means the user didn't update it.
    const updatedFields = config.filter(
      (item) => item.value !== MASKED_PASSWORD_VALUE,
    );
```
Should it also be added to the PATCH handler?

obarcelonap: My assumption is that since the UI was sending all the config always this was a trick to detect if a masked value was changed.
In the PATCH endpoint this do not happen anymore since the contract has changed to only send what has changed.

oshmyheliuk: Anyway, I would keep it just in case, so as not to override passwords with a bunch of asterisks.

oshmyheliuk (approved):

PR #389, opened 2026-04-15
CEXT-6073: add required id field to ConfigOrigin

obarcelonap:
## Description

`ConfigOrigin` now carries a required `id` field alongside the existing `code` and `level` fields. This allows consumers to unambiguously identify which scope a configuration value is inherited from, using the scope's unique identifier rather than relying on code+level alone. Schema-default origins use the sentinel value `"default"` for this field.

Propagated through `mergeScopes`, `mergeWithSchemaDefaults`, and all internal merge helpers. The changeset is marked `minor` — `ConfigOrigin` only appears in read-only response types; no write API accepts it as input.

## Related Issue

https://jira.corp.adobe.com/browse/CEXT-6073

## Motivation and Context

The UI needs to confidently determine whether a config value is inherited or locally set. `code` + `level` alone are not sufficient since multiple scopes can share the same level. The scope `id` is the canonical unique identifier.

## How Has This Been Tested?

- 134 unit and integration tests pass (`npx vitest run`)
- TypeScript check passes (`npx tsc --noEmit`)
- New `test/unit/config-utils.test.ts` covers `mergeScopes` id propagation: new entries, overridden entries, preserved inherited origins, and fallback for origin-less entries

## Types of changes

## Checklist:


packages/aio-commerce-lib-config/source/config-utils.ts
```diff
@@ -328,7 +335,7 @@ export function getSchemaDefaults(schema: BusinessConfigSchema) {
     .map((field) => ({
       name: field.name,
       value: field.default as string,
-      origin: { code: "default", level: "system" },
+      origin: { id: "default", code: "default", level: "system" },
```
iivvaannxx: I think that [we need to change](https://magento.slack.com/archives/C0A3SFDC6HY/p1776110999168729) this to `global` for both `code` and `level`. What do you think?

Also this `id` being `default` it bugs me. IDs should always be a UUID, maybe we should search the scope with that `code` and `level` and grab the `id` from it. I am not sure if there's any with `code: default` and `level: system`, by default, but we should have the `code: global` and `level: global`, which is the reason I suggested changing it.

packages/aio-commerce-lib-config/source/config-utils.ts
```diff
@@ -451,7 +463,7 @@ function applySchemaDefaults(
       merged.set(name, {
         name,
         value: def,
-        origin: { code: "default", level: "system" },
+        origin: { id: "default", code: "default", level: "system" },
```
iivvaannxx, line 466: Idem here

obarcelonap:
After analysis, we're not adding id to ConfigOrigin.
The original motivation was to let the UI determine whether a config value is inherited or locally set. id was proposed as the unambiguous comparator, but it's not needed:
  - code + level is already guaranteed unique across the scope tree — lib-config validates this and throws on duplicate code + level combinations
  - app-management never exposes level when creating scopes, so in practice code alone is sufficient there
  - The UI check simply becomes origin.code !== currentCode || origin.level !== currentLevel instead of origin.id !== currentId — no added complexity
 
 The id: "default" sentinel for schema defaults (@iivvaannxx concern in https://github.com/adobe/aio-commerce-sdk/pull/389/changes#r3109982635) will be addressed separately by changing the origin to `{ code: "global", level: "global" }`, which is the actual scope of CEXT-6105

PR #390, opened 2026-04-15
CEXT-6073: add AGENTS.md and CLAUDE.md with repo conventions

obarcelonap:
## Description

Adds a `AGENTS.md` and `CLAUDE.md` file to the repo root so Claude Code automatically picks up project conventions at the start of every session. Covers commit/branch naming, pre-push checks, PR creation, and changeset rules.

## Related Issue

https://jira.corp.adobe.com/browse/CEXT-6073

## Motivation and Context

Avoids having to repeat project conventions in every Claude Code session.

## How Has This Been Tested?

Manual — no code changes.

## Types of changes

## Checklist:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,31 @@
+# Claude Code — Repo Conventions
+
+## Commits
+
+- Keep commit messages terse (e.g. `fix auth token refresh`)
+
+## Branches
+
+- Prefix with the Jira ticket (e.g. `CEXT-1234/short-description`)
+
+## Git push
+
+- Always run `pnpm test` and `pnpm typecheck` before pushing (`pnpm check:ci` already runs on commit via lint-staged)
+- Always ask before pushing code
+
+## Pull requests
+
+- Always ask before creating a PR
+- Prefix the PR title with the Jira ticket (e.g. `CEXT-1234: short description`)
+- Always follow `.github/PULL_REQUEST_TEMPLATE.md` when writing PR descriptions
+
+## Changesets
+
+- Every code change that requires a release must include a changeset (test-only changes do not need one)
+- Create one with `pnpm changeset add --empty`, then edit the generated `.changeset/<name>.md` file
+- Changeset messages must be user-facing and concise — avoid implementation details
+- Derive the bump type from semver rules:
+  - `patch` — bug fixes, no API surface change
+  - `minor` — additive non-breaking changes (new exports, new optional fields, enriched responses)
+  - `major` — breaking changes (removed exports, required fields added to input/write types, renamed types)
+- If the bump type is ambiguous, ask before proceeding
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -29,3 +29,4 @@
   - `minor` — additive non-breaking changes (new exports, new optional fields, enriched responses)
   - `major` — breaking changes (removed exports, required fields added to input/write types, renamed types)
 - If the bump type is ambiguous, ask before proceeding
+- Before each commit, check if the changeset message still accurately describes the change
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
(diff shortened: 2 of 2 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,30 @@
+# Claude Code — Repo Conventions
+
+## Commits
+
+- Keep commit messages terse (e.g. `CEXT-1234: add x to y`)
+- Prefix with the Jira ticket (e.g. `CEXT-1234: ...`)
```
iivvaannxx: Is this necessary? I usually use conventional commits for this. When we squash commits from PRs the final commit will already (most of the time) have that prefix due to the PR title, but for normal commits I don't usually do this.

CLAUDE.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Pull requests
+
+- Always ask before creating a PR
+- Always follow `.github/PULL_REQUEST_TEMPLATE.md` when writing PR descriptions
+
+## Changesets
+
+- Every code change must include a changeset
+- Create one by writing a `.changeset/<short-name>.md` file directly (the `pnpm changeset add` command is interactive and cannot be used)
```
iivvaannxx: You can suggest doing `pnpm changeset add --empty` and then ask the LLM to edit it. I think it also accepts a `--message` flag.

CLAUDE.md
```diff
@@ -0,0 +1,30 @@
...
+- Always ask before pushing code
+
+## Pull requests
+
+- Always ask before creating a PR
+- Always follow `.github/PULL_REQUEST_TEMPLATE.md` when writing PR descriptions
+
+## Changesets
+
+- Every code change must include a changeset
```
iivvaannxx: Every code change that requires a release. For example, testing changes do not need one. Also I would add that messages need to be user-facing and not too verbose (avoiding implementation details)

iivvaannxx (commented):
I would create an `AGENTS.md` and symlink a `CLAUDE.md` out of it.

CLAUDE.md
obarcelonap: removed

CLAUDE.md
obarcelonap: adapted

CLAUDE.md
obarcelonap: done

iivvaannxx (approved):

AGENTS.md
```diff
@@ -0,0 +1,31 @@
+# Claude Code — Repo Conventions
+
+## Commits
+
+- Keep commit messages terse (e.g. `fix auth token refresh`)
```
oshmyheliuk, line 5: I would add 
```
Run pnpx lint-staged before commit and fix related errors
```

CLAUDE.md
```diff
@@ -0,0 +1 @@
+AGENTS.md
```
oshmyheliuk: As far as I know, it should be @AGENTS.md to tell Claude to import files

AGENTS.md
oshmyheliuk, line 5: Also, `before each commit, check if the changeset needs to be updated`.
As with new commits, previously created changesets may be outdated.

obarcelonap: According to claude:

>   Doesn't make sense. The file already says pnpm check:ci runs automatically via lint-staged on commit. Adding an instruction to run it manually beforehand is redundant — if Claude commits and the hook fails,
>    it'll fix and retry. No value added.

So I'll skip the suggestion.

CLAUDE.md
obarcelonap: It was a symlink which is shown like this in github. In any case I applied your suggestion which I believe is more portable.

AGENTS.md
obarcelonap: Claude flagged that adding this might be redundant but I added in any case. Let's see how it goes.

oshmyheliuk (approved):

PR #393, opened 2026-04-16
CEXT-6085: Improve logging and error messages of lib-app

oshmyheliuk:
## Description

Improved installation and uninstallation error messages: HTTP API errors from Commerce and Adobe IO Events now include the response body in the error message, making it clearer which entity failed and why.

https://jira.corp.adobe.com/browse/CEXT-6085

## Types of changes

## Checklist:


danidelcar (approved):

packages/aio-commerce-lib-app/source/actions/installation.ts
```diff
@@ -249,7 +253,9 @@ router.post("/", {
   body: InstallationRequestBodySchema,
 
   handler: async (req, { logger, rawParams }) => {
```
iivvaannxx, line 255: Given the several instances of the same variables reused for logging, I see value on destructuring things like `appData` or `params` (as needed) for better readability

packages/aio-commerce-lib-app/source/management/installation/admin-ui-sdk/helpers.ts
```diff
@@ -30,18 +30,27 @@ export async function registerExtension(context: AdminUiSdkExecutionContext) {
...
-    })
-    .json<RegisterExtensionResponse>();
+      })
+      .json<RegisterExtensionResponse>();
+  } catch (error: unknown) {
+    const msg = await unwrapHttpError(error);
+    const enriched = `Failed to register Admin UI SDK extension: ${msg}`;
+    logger.error(enriched);
+    throw new Error(enriched);
+  }
```
iivvaannxx: I would suggest keeping it consistent with other API calls and use the `.then` and `.catch` notation for this?

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
```diff
@@ -89,12 +90,11 @@ async function createIoEventProvider(params: CreateIoProviderParams) {
...
-      logger.error(
-        `Failed to create provider "${provider.label}": ${stringifyError(error)}`,
-      );
-
-      throw error;
+    .catch(async (error) => {
+      const msg = await unwrapHttpError(error);
+      const enriched = `Failed to create I/O Events provider '${provider.label}': ${msg}`;
+      logger.error(enriched);
+      throw new Error(enriched);
```
iivvaannxx: Given the repeated pattern, what about creating a `handleError` function? It could, log, throw and maybe accept a way to "decorate" the message, either via a callback or a parameter

packages/aio-commerce-lib-api/source/utils/http/error.ts
iivvaannxx, line 1: I think this would be better located into the `utils` of `lib-api`, then just imported as needed

packages/aio-commerce-lib-app/source/management/installation/webhooks/helpers.ts
```diff
@@ -245,9 +245,8 @@ export async function deleteWebhookSubscriptions(
       logger.info(`Unsubscribed webhook: ${getWebhookName(webhook)}`);
       unsubscribedWebhooks.push(params);
     } catch (error) {
-      logger.warn(
-        `Failed to unsubscribe webhook "${getWebhookName(webhook)}": ${stringifyError(error)}. Continuing uninstall.`,
-      );
+      const msg = error instanceof Error ? error.message : String(error);
```
iivvaannxx: For this kind of message unwrapping I would keep `stringifyError`.

packages/aio-commerce-lib-app/test/unit/management/installation/admin-ui-sdk/helpers.test.ts
iivvaannxx, line 1: I'm good with the new tests, except for the tests unrelated to logging (or any other file). I was going to implement Admin UI SDK tests in a separate PR, in case you want to remove ones like:

- `calls DELETE with correct endpoint`
- `resolves without throwing when DELETE fails (best-effort)`

And similar ones

packages/aio-commerce-lib-app/test/unit/management/installation/events/helpers.test.ts
```diff
@@ -68,6 +69,17 @@ const DEFAULT_WORKSPACE_CONFIGURATION = JSON.stringify(
...
+const RE_FAIL_IO_PROVIDER = /^Failed to create I\/O Events provider '/;
+const RE_FAIL_IO_METADATA = /^Failed to register I\/O Events metadata for '/;
+const RE_FAIL_IO_REGISTRATION = /^Failed to create I\/O Events registration '/;
+const RE_FAIL_CONFIGURE_EVENTING =
+  /^Failed to configure Adobe Commerce eventing:/;
+const RE_FAIL_COMMERCE_PROVIDER =
+  /^Failed to create Adobe Commerce event provider '/;
+const RE_FAIL_COMMERCE_SUBSCRIPTION =
+  /^Failed to create Adobe Commerce event subscription for '/;
+const RE_CONTINUING_UNINSTALL = /Continuing uninstall\./;
```
iivvaannxx, line 81: Do we want to test concrete error messages? I am not against it, but I don't find it necessary

packages/aio-commerce-lib-app/test/unit/management/installation/admin-ui-sdk/helpers.test.ts
```diff
@@ -0,0 +1,162 @@
...
+  return {
+    ...mockInstallation,
+    commerceClient: {
+      post: postFn,
+      delete: vi
+        .fn()
+        .mockImplementation(overrides?.deleteImpl ?? (() => Promise.resolve())),
+    } as unknown as AdminUiSdkExecutionContext["commerceClient"],
+  };
+}
```
iivvaannxx: Should be moved to `fixtures`?

packages/aio-commerce-lib-app/source/actions/installation.ts
oshmyheliuk: Done in https://github.com/adobe/aio-commerce-sdk/pull/393/changes/bdbc1fa13427b675e227eb8d1cc5fc8ec728cdda

packages/aio-commerce-lib-app/source/management/installation/events/helpers.ts
oshmyheliuk: Makes sense, create a helper function for the repeated logic https://github.com/adobe/aio-commerce-sdk/pull/393/changes/9f0662a3285d4a1df9009445886ba8b95fbf6d3d

packages/aio-commerce-lib-app/source/management/installation/admin-ui-sdk/helpers.ts
oshmyheliuk: Done here https://github.com/adobe/aio-commerce-sdk/pull/393/changes/9f0662a3285d4a1df9009445886ba8b95fbf6d3d

packages/aio-commerce-lib-api/source/utils/http/error.ts
oshmyheliuk: I'm against moving it to another package, as no other packages need it today. Moving it to the lib api adds a public API commitment.

packages/aio-commerce-lib-app/source/management/installation/webhooks/helpers.ts
oshmyheliuk: Done here https://github.com/adobe/aio-commerce-sdk/pull/393/changes/fccb9e0463d5c5d120ec85c054916ef7e5213248

packages/aio-commerce-lib-app/test/unit/management/installation/events/helpers.test.ts
oshmyheliuk: I would leave it in this case. We are not testing the whole error message, but only the part that we know must be there. It's a kind of message contract, we know that we want to show in UI and need to test it.

packages/aio-commerce-lib-app/test/unit/management/installation/admin-ui-sdk/helpers.test.ts
oshmyheliuk: Moved https://github.com/adobe/aio-commerce-sdk/pull/393/changes/cf332112bdaa4435d5556d2f0fa38469e7cbad5b

packages/aio-commerce-lib-app/test/unit/management/installation/admin-ui-sdk/helpers.test.ts
oshmyheliuk: Removed a few tests https://github.com/adobe/aio-commerce-sdk/pull/393/changes/a6d43ec92a4ed570f5664794d1195a432e279845

obarcelonap (approved):
LGTM but again in the future we have to think about localization since these errors are user facing.

packages/aio-commerce-lib-api/source/utils/http/error.ts
iivvaannxx: Well it's not necessary for another package to need it. I think this provides enough value for users of the libraries that might have the same issue we're trying to solve (unwrap the http errors of API calls made via events library, or webhooks library, or the HTTP clients of `lib-api`, to understand the underlying error). 

It adds a public API commitment, but it's a really small helper, so I don't see the issue there. It's part of the core value of these libs.

packages/aio-commerce-lib-app/test/unit/management/installation/events/helpers.test.ts
iivvaannxx: Fair enough 👍🏻

packages/aio-commerce-lib-app/test/fixtures/http-error.ts
```diff
@@ -0,0 +1,35 @@
...
+    new Error(`Request failed with status code ${status}`),
+    {
+      response,
+      request: new Request("https://example.com"),
+      options: {},
+    },
+  );
+  Object.setPrototypeOf(error, HTTPError.prototype);
+  return error as HTTPError;
+}
```
iivvaannxx, line 26: Doesn't a simple `new HTTPError` with mocked inputs work?

iivvaannxx (approved):

packages/aio-commerce-lib-api/source/utils/http/error.ts
oshmyheliuk: Ok, makes sense, moved it to lib-api https://github.com/adobe/aio-commerce-sdk/pull/393/changes/774f62126648e1ae13368a786abd322d5539db3b

packages/aio-commerce-lib-app/test/fixtures/http-error.ts
oshmyheliuk: I simplified the code of this fixture

packages/aio-commerce-lib-api/source/utils/http/error.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export async function throwHttpError(
+  logger: { error: (msg: string) => void },
+  error: unknown,
+  prefix: string,
+): Promise<never> {
+  const message = `${prefix}: ${await unwrapHttpError(error)}`;
+  logger.error(message);
+  throw new Error(message);
+}
```
iivvaannxx: The throw one I think it's fine kept internal for now. I mainly referred to the other. This logic is specific enough to our use case I'd say, that it doesn't belong here. Let me know if you think otherwise.

packages/aio-commerce-lib-api/source/utils/http/error.ts
iivvaannxx: Nice! Thanks 👍🏻

iivvaannxx (approved):

iivvaannxx:
Might be good adding the new helper documentation into the `usage.md` of `lib-api`!

packages/aio-commerce-lib-api/source/utils/http/error.ts
oshmyheliuk: I think, as we already shared `unwrapHttpError`, it will also be nice to have `throwHttpError`, probably most SDK callers will use the same pattern, and if they don't, they can still use `unwrapHttpError` directly.

oshmyheliuk:
Updated usage.md of lib-api

packages/aio-commerce-lib-api/source/utils/http/error.ts
iivvaannxx: The only thing that bugs me is that we call it `throwHttpError`, but we also ask for a logger, and a prefix. So the logic inside the helper is not fully inferrable from just the name, and might generate confusion in a Public API.

iivvaannxx:
> Updated usage.md of lib-api

@oshmyheliuk Did you push?

oshmyheliuk:
@iivvaannxx sorry, I forgot to push, here is the commit https://github.com/adobe/aio-commerce-sdk/pull/393/changes/1e6a1fad98193b4d15c09034830b3fd4feed56a9

obarcelonap (commented):
Looked again at the code. 
While current approach is sufficient, please consider we use a structured error instead of string-concatenated messages.
Something like:

```typescript
interface ApiErrorPayload {
  errorCode: string;
  errorResponse: {
    body: unknown;
    statusCode: number;
  };
}

class ApiError extends Error {
  payload: ApiErrorPayload;

  constructor(payload: ApiErrorPayload) {
    super(`API error: ${payload.errorCode} (${payload.errorResponse.statusCode})`);
    this.payload = payload;
  }
}
```

Why:
- Callers can handle errors programmatically via `errorCode` instead of parsing strings
- The raw response (`body`, `statusCode`) is preserved, so no information is lost and each caller decides how to handle it
- `errorCode` gives us a stable key for localization and conditional logic
- Extends `Error`, so stack traces and catch blocks work as expected

packages/aio-commerce-lib-api/source/utils/http/error.ts
obarcelonap: I kind of agree with Ivan, I don't like logging in utilities I feel is a concern of the main execution.
Also the prefix as arg, please consider structured error (see global msg related to this topic).

iivvaannxx:
> Extends Error, so stack traces and catch blocks work as expected

Might be good extending from our internal error, it also extends `Error`, but we designed in a way that it handles some quirks of the native `Error` out of the box, plus it comes with some niceties.

This way we can enhance it in a single place for every API call across the whole SDK. Something like `CommerceSdkApiError`. See how we do it for validation errors:

https://github.com/adobe/aio-commerce-sdk/blob/1659779782e696c07755003b73029c64f58a00b0/packages/aio-commerce-lib-core/source/error/validation-error.ts#L103-L127

packages/aio-commerce-lib-api/source/utils/http/error.ts
oshmyheliuk: Ok, moved `throwHttpError` back, now it's not public

oshmyheliuk:
@obarcelonap I agree that we can add the structured error, but it can be done in a separate PR. 

`unwrapHttpError` and `ApiError` solve different problems. `unwrapHttpError` is just a message-formatter (HTTPError -> readable string), the other is a domain error type. Introducing ApiError doesn't remove the need for the formatter.

oshmyheliuk:
@iivvaannxx @obarcelonap, please create a separate ticket if you think that we need to introduce a new Api Error.

PR #402, opened 2026-04-20
CEXT-6117: Support Admin UI components

oshmyheliuk:
## Description

Added full Admin UI SDK support: generate the `commerce/backend-ui/1` registration action from `adminUiSdk.registration` config, covering menu items, mass actions, grid columns, view buttons, custom fees, and banner notifications.

https://jira.corp.adobe.com/browse/CEXT-6117

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/source/commands/generate/actions/main.ts
```diff
@@ -267,3 +287,44 @@ export async function generateCustomScriptsTemplate(
   const result = template.replace(CUSTOM_IMPORTS_PLACEHOLDER, importStatements);
   return result.replace(CUSTOM_SCRIPTS_MAP_PLACEHOLDER, scriptMap);
 }
+
+function toJsObjectLiteral(value: unknown): string {
+  const json = JSON.stringify(value, null, 2);
+  return json.replace(/^(\s*)"([A-Za-z_$][\w$]*)":/gm, "$1$2:");
+}
```
iivvaannxx: Let me suggest you to use `prettier` for this. This is how we do it for the `app.commerce.config.ts` generation in the `init` script. It is able to take an object and produce a valid JS string from it, while also nicely formatting it.

https://github.com/adobe/aio-commerce-sdk/blob/23798c596bab43e2f9f51f5dff8e95ea0ae772d4/packages/aio-commerce-lib-app/source/commands/init/lib.ts#L100-L120

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
```diff
@@ -10,26 +10,168 @@
...
+      order: v.optional(v.array(MassActionBannerSchema)),
+      product: v.optional(v.array(MassActionBannerSchema)),
+      customer: v.optional(v.array(MassActionBannerSchema)),
+    }),
+  ),
+  orderViewButtons: v.optional(v.array(OrderViewButtonBannerSchema)),
+});
+
+const MenuItemSchema = v.object({
+  id: nonEmptyStringValueSchema("menu item ID"),
```
iivvaannxx, line 237: From what I see in documentation, examples, and samples, we usually prefix the `appId` to these. Do we want to automate that? Similar to webhooks and events?

iivvaannxx (commented):
Do we need to generate the `extension-registration.json`?

packages/aio-commerce-lib-app/source/commands/generate/actions/main.ts
oshmyheliuk: Fixed in https://github.com/adobe/aio-commerce-sdk/pull/402/changes/bb5526f0b91781caa96dda12f68135d8783d9332

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
oshmyheliuk: It makes sense, but in the documentation we currently have next example:

```
menuItems: [
    {
        id: `${extensionId}::first`,
        title: 'App on App Builder',
        parent: `${extensionId}::apps`,
        sortOrder: 1
    }
]
```
And in [code samples.](https://github.com/adobe/adobe-commerce-samples/blob/main/admin-ui-sdk/menu/custom-menu/src/commerce-backend-ui-1/actions/registration/index.js)
So it's already expected from the user to add an extension ID. If we prepend another one, it can be too much.
@asalloum5 What do you think would be better here?

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
```diff
@@ -10,26 +10,168 @@
...
+
+const massActionBaseEntries = {
+  actionId: nonEmptyStringValueSchema("mass action ID"),
+  label: nonEmptyStringValueSchema("mass action label"),
+  title: v.optional(nonEmptyStringValueSchema("mass action page title")),
+  confirm: v.optional(MassActionConfirmSchema),
+  path: nonEmptyStringValueSchema("mass action path"),
+  displayIframe: v.optional(booleanValueSchema("displayIframe")),
+  timeout: v.optional(positiveNumberValueSchema("timeout")),
+  sandbox: v.optional(nonEmptyStringValueSchema("sandbox")),
```
asalloum5: For sandbox we support limited values with single space separated. Do we want to add a check here on the provided value?
```
Applies additional restrictions to the content within an iFrame. Allowed values are allow-downloads, allow-modals, and allow-popups. Separate multiple values with a single space (" ").
```

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
asalloum5: I prefer not to automate it to avoid breaking changes on apps already using Admin UI SDK so it would be easier to migrate to App Management without any issue on Commerce side.

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
```diff
@@ -10,26 +10,168 @@
...
+  confirm: v.optional(ViewButtonConfirmSchema),
+  path: nonEmptyStringValueSchema("view button path"),
+  level: v.optional(ViewButtonLevelSchema),
+  sortOrder: v.optional(positiveNumberValueSchema("sortOrder")),
+  displayIframe: v.optional(booleanValueSchema("displayIframe")),
+  timeout: v.optional(positiveNumberValueSchema("timeout")),
+  sandbox: v.optional(nonEmptyStringValueSchema("sandbox")),
+});
+
+const CustomFeeSchema = v.object({
```
asalloum5, line 181: I think Dani had in mind to deprecate custom fees from Admin UI SDK, I'd suggest to double check this before adding it in App Management.

packages/aio-commerce-lib-app/source/commands/generate/actions/config.ts
```diff
@@ -211,3 +218,46 @@ export function buildBusinessConfigurationExtConfig() {
...
+
+    runtimeManifest: {
+      packages: {
+        [ADMIN_UI_SDK_PACKAGE_NAME]: {
+          license: "Apache-2.0",
+          actions: {
+            registration: {
+              function: `${ADMIN_UI_SDK_ACTIONS_PATH}/index.js`,
+              web: "yes",
+              runtime: "nodejs:22",
```
obarcelonap, line 246: for the sake of maintainability this should be configurable (no need to do it now)

obarcelonap (approved):
LGTM

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
iivvaannxx: I'm fine with this then, but might be worth validating it? Should the extension ID be the same as the app ID? Like a simple check of `startsWith`

asalloum5: No need, they can decide to use another id that they generate, a UUID for example would work. The uniqueness is with commerce menu ids and we can't validate anything here. The usage of extensionId is a recommendation only.

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
oshmyheliuk: I added everything that is currently supported, except the `page` that we decided to remove.
@danperezAdobe Should we skip adding a Custom Fee for now?

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
oshmyheliuk: Added for all elements with sandbox except menuItems, as according to the description of menuItems that rule does not apply
https://developer.adobe.com/commerce/extensibility/admin-ui-sdk/extension-points/menu/

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
oshmyheliuk: @asalloum5 Dani approved the CustomFees support for now, so I think we good

packages/aio-commerce-lib-app/source/commands/generate/actions/config.ts
```diff
@@ -211,3 +218,46 @@ export function buildBusinessConfigurationExtConfig() {
...
+        "EXTENSION=backend-ui/1 $packageExec aio-commerce-lib-app hooks pre-app-build",
+    },
+
+    operations: {
+      workerProcess: [
+        {
+          type: "action",
+          impl: `${ADMIN_UI_SDK_PACKAGE_NAME}/registration`,
+        },
+      ],
```
asalloum5: Admin UI SDK doesn't have a workerProcess for now, it might in the future after the Spike is finalized. Maybe we want to remove this for now and we can add it later if we decide to create the `commerce/backend-ui/2` version.

asalloum5 (approved):

packages/aio-commerce-lib-app/source/commands/generate/actions/config.ts
oshmyheliuk: Removed from schema

PR #373, opened 2026-04-07
CEXT-6074: Add CLI Tests for `lib-app`

iivvaannxx:
## Description

Adds integration and unit tests for CLI commands

## Comparison

This PR was not focused on coverage, just adding tests for untested logic. But still, showing some numbers for reference:

Before:

image

<br />

After:

image

<br />

The uncovered lines of the `init` script are the interactive bits (user prompting...), which aren't worth testing to me, as long as the underlying logic is tested.

## Ticket

[CEXT-6074](https://jira.corp.adobe.com/browse/CEXT-6074)

PR #403, opened 2026-04-20
CEXT-6074: Fix issues surfaced during test implementation + `lib-app` CLI tests

iivvaannxx:
## Description

Source code changes of issues/code organization made during #373.

## Summary

- Implement `touch` to create an empty `.env` when running `encryption setup` instead of throwing because of a non-existing `.env` file.
- Fix logic that detects when the `installation` action is needed (not generated for `webhooks` and `admin-ui-sdk`)
- Fix a bug with project root dir resolution in a couple of commands
- Improve testability by introducing pluggable parameters for some flows
- Cleanup CLI commands `main` files by code-splitting
- Fix currently existing test issues due to the changes


packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
```diff
@@ -0,0 +1,215 @@
+/*
+ * Copyright 2025 Adobe. All rights reserved.
```
obarcelonap: 2026

packages/aio-commerce-lib-app/source/commands/utils.ts
```diff
@@ -22,3 +34,42 @@ export async function loadAppManifest() {
 
   return appConfig;
 }
+
```
obarcelonap, line 55: Aren't we missing the arg docs in the functions of this file? I am surprised linter is not complaining

packages/aio-commerce-lib-app/source/commands/init/lib.ts
```diff
@@ -77,40 +82,49 @@ export async function ensureCommerceAppConfig(cwd = process.cwd()) {
...
-      trailingComma: "all",
-      tabWidth: 2,
-      useTabs: false,
-      printWidth: 80,
-      filepath: path,
-    });
 
-    await writeFile(path, formattedConfig, "utf-8");
+    // Skip formatting during tests, prettier cold-start is expensive and formatting is not under test.
+    if (process.env.VITEST) {
```
jcuerdo: process.env.VITEST in production code, a cleaner approach would be to inject the formatter as an optional parameter with a default.

jcuerdo (approved):

packages/aio-commerce-lib-config/source/commands/encryption/setup/main.ts
oshmyheliuk: The run function was changed to `async`, looks like you need to add `await` here.

packages/aio-commerce-lib-app/source/commands/utils.ts
```diff
@@ -22,3 +34,42 @@ export async function loadAppManifest() {
...
+  );
+}
+
+/** Path to the generated configuration schema file, relative to the project root. */
+export function getSchemaPath() {
+  return join(
+    getGeneratedDir(CONFIGURATION_EXTENSION_POINT_ID),
+    CONFIG_SCHEMA_FILE_NAME,
+  );
+}
```
oshmyheliuk, line 125: I don't see any usage of these functions

packages/aio-commerce-lib-app/source/commands/init/lib.ts
```diff
@@ -278,6 +292,10 @@ export async function runGeneration(
   try {
     await generateActionsCommand(appConfig);
     await generateManifestCommand(appConfig);
+
+    if (hasConfigDomain(appConfig, "businessConfig.schema")) {
```
oshmyheliuk: This condition looks redundant as `generateSchemaCommand` command has the same validation with early return and additional logging. 

```
  if (!hasBusinessConfigSchema(appConfig)) {
    consola.warn("Business configuration schema not found");
    return;
  }
```

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
oshmyheliuk, line 1: I don't understand how moving functions from main.ts to lib.ts made the code more testable.
Code from `lib` is used only in `main` and is not reusable elsewhere.
In `run.test.ts`, nothing changed, only import the source from main to lib.
I don't mind splitting, but I don't see any benefits in this case or how it's improving testability.

iivvaannxx: Testability didn't came from that. Maybe it was a bold statement. The PR doesn't improve overall testability, but only on some parts, for me:

- Separating the `exec` into a `run` function with pluggable and easily mockable inputs.

- Adding the `InitOptions` for the interactive prompting part of the `init`

- The extra `templatesDir` argument on the generate commands.

Not sure if I left some but those are the main ones I recall.

Separating code wasn't intended for that. I only did it because I want to keep the `main` files lean and small so that they can be easily read and you can understand what each command does at a glance. You can consider it more of a cleanup.

packages/aio-commerce-lib-app/source/commands/init/lib.ts
iivvaannxx: But then you would get that warn message when running the `init`. It's odd that we log a warning when:

1. User intentionally omits scaffolding with business configuration.

2. User has an already created `app.commerce.config.ts without it.

In both cases, logging that warning might induce the user into thinking it has/did something wrong, when is not the case.

If you have a better way to handle this please suggest. If not I can add a comment to clarify this.

oshmyheliuk: Maybe we can change the warning message to debug, something like `Business configuration schema not found, omitting this <name> step.`.  And move the `starting` log after it. 

```
export async function run(appConfig: CommerceAppConfigOutputModel) {
  if (!hasBusinessConfigSchema(appConfig)) {
    consola.debug("Business configuration schema not found. Skipping configuration schema generation.");
    return;
  }

  consola.info("Generating configuration schema...");
```

.changeset/wild-geckos-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-config": patch
```
asalloum5, line 2: `lib-app` has no changeset entry.

packages/aio-commerce-lib-app/source/commands/init/lib.ts
```diff
@@ -77,40 +82,49 @@ export async function ensureCommerceAppConfig(cwd = process.cwd()) {
...
-      printWidth: 80,
-      filepath: path,
-    });
 
-    await writeFile(path, formattedConfig, "utf-8");
+    // Skip formatting during tests, prettier cold-start is expensive and formatting is not under test.
+    if (process.env.VITEST) {
+      await writeFile(path, configContent, "utf-8");
+    } else {
+      const prettier = await import("prettier");
```
asalloum5: This pattern works, but conditional test-only code paths in production source are a smell. 

Two lighter alternatives:
- Make prettier formatting opt-in via a parameter ({ format?: boolean }) and have tests pass false. The production entry point passes true. No env sniffing in prod code.
- Mock prettier in the test setup.

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
asalloum5:
```suggestion
 * Copyright 2026 Adobe. All rights reserved.
```

asalloum5 (commented):
Please consider updating the description of the PR since it adds a lot of changes and can be a bit confusing to follow.

packages/aio-commerce-lib-app/source/commands/init/lib.ts
iivvaannxx: Agree, didn't like it neither, but did not think of this, will do 👍🏻

packages/aio-commerce-lib-app/source/commands/utils.ts
iivvaannxx: They are actually used in #373. Seems I missed them when I segregated the `source` changes from the `tests`. Will keep them as they will be used by the incremental PR anyways 👍🏻

packages/aio-commerce-lib-config/source/commands/encryption/setup/main.ts
iivvaannxx: Fixed in [`d494361`](https://github.com/adobe/aio-commerce-sdk/pull/403/commits/d4943617ff3ca9f3a78026129a1a84e058718300)

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
iivvaannxx: Fixed in [`9f7f8eb`](https://github.com/adobe/aio-commerce-sdk/pull/403/commits/9f7f8eb52b40046412cd887c80e9386c90e72a0d)

.changeset/wild-geckos-knock.md
iivvaannxx: Fixed in [`9f7f8eb`](https://github.com/adobe/aio-commerce-sdk/pull/403/commits/9f7f8eb52b40046412cd887c80e9386c90e72a0d)

packages/aio-commerce-lib-app/source/commands/init/lib.ts
iivvaannxx: Done in [`47119f6`](https://github.com/adobe/aio-commerce-sdk/pull/403/commits/47119f69865742f19a77c8cc63fd835a7c8b7a8d)

packages/aio-commerce-lib-app/source/commands/init/lib.ts
iivvaannxx: Done in [`a47b0d4`](https://github.com/adobe/aio-commerce-sdk/pull/403/commits/a47b0d48c91ab3c5ba7d956311ff7af8417b8eea)

iivvaannxx:
> Please consider updating the description of the PR since it adds a lot of changes and can be a bit confusing to follow.

@asalloum5 Updated description with a summary!

asalloum5 (approved):

oshmyheliuk (approved):

obarcelonap (approved):

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
iivvaannxx: Fixed!

packages/aio-commerce-lib-app/source/commands/utils.ts
iivvaannxx: Oh the linter doesn't complain about that, not sure if there's a rule for it. If there is, I can guarantee those will not be the only instances

iivvaannxx: Addressed in [`3a102d4`](https://github.com/adobe/aio-commerce-sdk/pull/403/commits/3a102d46ac13a4e298fffcb5944cb4a68156da74)

jcuerdo (approved):

PR #405, opened 2026-04-20
CEXT-6084: Fix support for other package managers

iivvaannxx:
## Description

Vaimo raised an issue when using `pnpm` as a package manager, where app hooks were not working as expected. This PR fixes the issues with binary resolution for `lib-app` when using package managers other than `npm` and `bun` (which worked fine).

## Ticket

[CEXT-6084](https://jira.corp.adobe.com/browse/CEXT-6084)


asalloum5 (approved):

packages-private/scripting-utils/source/project.ts
```diff
@@ -135,7 +136,55 @@ export async function makeOutputDirFor(fileOrFolder: string) {
...
+    if (major) {
+      return "yarn-berry";
+    }
+  }
+
+  if (existsSync(join(rootDirectory, ".yarnrc.yml"))) {
+    return "yarn-berry";
+  }
+
+  try {
```
oshmyheliuk: What about using  some popular lib for this, like https://github.com/antfu-collective/package-manager-detector
It's more reliable than our custom solution.
Can be refactored to something like this:

```
import { detect } from "package-manager-detector";

export async function detectPackageManager(
  cwd = process.cwd(),
): Promise<PackageManager> {
  const result = await detect({ cwd });
  switch (result?.agent) {
    case "yarn":
      return "yarn-classic";
    case "yarn@berry":
      return "yarn-berry";
    case "pnpm":
    case "pnpm@6":
      return "pnpm";
    case "bun":
      return "bun";
    default:
      return "npm";
  }
}
```

iivvaannxx: It might be worth it at this point yes. Will do 👍🏻

iivvaannxx: Done in [`e0f3961`](https://github.com/adobe/aio-commerce-sdk/pull/405/commits/e0f396183b50b2842012c68b055e03f1e09bc32b)

obarcelonap (approved):

oshmyheliuk (approved):

PR #406, opened 2026-04-20
CEXT-6105: Use global as default level in byCode()

jcuerdo:
## Types of changes

## Checklist:

### DIFF
--- a/.changeset/open-rules-clap.md
+++ b/.changeset/open-rules-clap.md
@@ -2,4 +2,4 @@
 "@adobe/aio-commerce-lib-config": minor
 ---
 
-Use global as default level in byCode()
+Make scope selector optional in getConfiguration, getConfigurationByKey and setConfiguration, defaulting to the global scope
--- a/.changeset/open-rules-clap.md
+++ b/.changeset/open-rules-clap.md
@@ -1,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-config": minor
+---
+
+Use global as default level in byCode()
--- a/packages/aio-commerce-lib-config/source/config-manager.ts
+++ b/packages/aio-commerce-lib-config/source/config-manager.ts
@@ -10,7 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
-import { byCodeAndLevel } from "#config-utils";
+import { byCode } from "#config-utils";
 import {
   requireGlobalSchema,
   setGlobalSchema,
@@ -308,10 +308,7 @@ export async function unsyncCommerceScopes() {
  * ```
  */
 export async function getConfiguration(
-  selector: SelectorBy = byCodeAndLevel(
-    DEFAULT_CUSTOM_SCOPE_LEVEL,
-    DEFAULT_CUSTOM_SCOPE_LEVEL,
-  ),
+  selector: SelectorBy = byCode(DEFAULT_CUSTOM_SCOPE_LEVEL),
   options?: ConfigOptions,
 ) {
   const context = {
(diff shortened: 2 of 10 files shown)


asalloum5 (approved):

iivvaannxx:
I think we should also change the default value of `level` to `global`:

In here:

https://github.com/adobe/aio-commerce-sdk/blob/23798c596bab43e2f9f51f5dff8e95ea0ae772d4/packages/aio-commerce-lib-config/source/config-utils.ts#L325-L335

And here:

https://github.com/adobe/aio-commerce-sdk/blob/23798c596bab43e2f9f51f5dff8e95ea0ae772d4/packages/aio-commerce-lib-config/source/config-utils.ts#L445-L458

iivvaannxx:
Also I would add a default to all the operations, `getConfiguration`, `getConfigurationByKey`, `setConfiguration`.... to (`byCodeAndLevel('global', 'global'`), as after this change it should always resolve to something.

jcuerdo:
@iivvaannxx all done ¡¡

.changeset/open-rules-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-config": minor
+---
+
+Use global as default level in byCode()
```
iivvaannxx: Try asking the AI to improve this a bit based on the changes of the PR :)

packages/aio-commerce-lib-config/source/utils/constants.ts
```diff
@@ -20,4 +20,4 @@ export const DEFAULT_NAMESPACE = "aio-commerce-config";
 export const DEFAULT_CACHE_TIMEOUT = 300;
 
 /** Default custom scope level. */
-export const DEFAULT_CUSTOM_SCOPE_LEVEL = "base";
+export const DEFAULT_CUSTOM_SCOPE_LEVEL = "global";
```
iivvaannxx, line 23: ~Given that this constant is both used now for `code` and `level` maybe we can rename for clarity?~

Dismiss this comment as per comment below.

packages/aio-commerce-lib-config/source/config-manager.ts
```diff
@@ -303,7 +308,10 @@ export async function unsyncCommerceScopes() {
  * ```
  */
 export async function getConfiguration(
-  selector: SelectorBy,
+  selector: SelectorBy = byCodeAndLevel(
+    DEFAULT_CUSTOM_SCOPE_LEVEL,
+    DEFAULT_CUSTOM_SCOPE_LEVEL,
```
iivvaannxx: actually with the changes in this PR, the shorter version of this, `byCode(DEFAULT_CUSTOM_SCOPE_LEVEL)` should work right?

packages/aio-commerce-lib-config/source/config-utils.ts
```diff
@@ -328,7 +328,7 @@ export function getSchemaDefaults(schema: BusinessConfigSchema) {
     .map((field) => ({
       name: field.name,
       value: field.default as string,
-      origin: { code: "default", level: "system" },
+      origin: { code: "default", level: DEFAULT_CUSTOM_SCOPE_LEVEL },
```
iivvaannxx: My bad @jcuerdo I had a chat with Claude about the scope of the ticket and how this `level` works and it does not need to be changed, it will be misleading, you can revert this back.

`default` actually indicates that the origin of the value comes from the schema default, I understood that differently. `system` doesn't mean much here, but it should not change, solves nothing about what the ticket asked.

iivvaannxx:
Also I think there are some mentions in the docs about the `base` level you changed. I think that may need updating

iivvaannxx (approved):

jcuerdo:
@danidelcar as far as there are changes in this doc, please check if a change in public docs is needed 👍

PR #411, opened 2026-04-23
CEXT-6138: spec process setup and capability discovery design

obarcelonap:
## Description

Introduces spec-driven development infrastructure and the capability discovery spec.

**Spec infrastructure (`specs/`):**
- `specs/conventions.md` — when to write a spec, two-PR workflow, status model, one-spec-per-ticket rule
- `specs/features/_template.md` — template for new specs
- `AGENTS.md` updated with spec authoring guidance

**Feature spec (`specs/features/CEXT-6138-capability-discovery.md`):**
- Filtered OpenAPI spec generated by `pre-app-build` based on active config domains, bundled into the `app-config` action and served at `/app-config/openapi.json`
- Auth-gated with IMS bearer token to prevent deployment fingerprinting
- Replaces the current pattern of inferring feature availability by inspecting `getAppConfig` response shapes

## Related Issue

[CEXT-6138](https://jira.corp.adobe.com/browse/CEXT-6138)

## How Has This Been Tested?

N/A — spec and process only, no runtime code changes.

## Types of changes

## Checklist:


specs/conventions.md
```diff
@@ -0,0 +1,42 @@
+# Spec Conventions
+
+## Structure
+
+Specs live in `specs/features/`. Each spec is a single Markdown file named after the feature it describes (e.g. `capability-discovery.md`).
```
oshmyheliuk: This can become very messy with a large list of specs. A flat features/ directory has no domain grouping, no chronological hint, no visual structure. I would suggest to add ad list prefix as a ticket ID:
`cext-6138-capability-discovery.md`. It will be much easier to find a spec for the specific ticket.

specs/conventions.md
```diff
@@ -0,0 +1,42 @@
...
+
+| Ticket                                                    | Description     |
+| --------------------------------------------------------- | --------------- |
+| [CEXT-XXXX](https://jira.corp.adobe.com/browse/CEXT-XXXX) | Initial spec    |
+| [CEXT-YYYY](https://jira.corp.adobe.com/browse/CEXT-YYYY) | Added X section |
+```
+
+## Keeping specs current
+
+A spec is a living document. When implementation reveals that the approach needs to change, update the spec before (or alongside) the code change. The metadata table entry for the corresponding ticket serves as the changelog.
```
oshmyheliuk: It would be very difficult to track that the spec was updated on each change. How would we control it?

For example, PR was created based on the spec. The reviewer of that PR left 5 - 10 comments.
The implementor made changes:
1.  Without spec updating - who should control it?
2. With spec updates: Should the specs be updated/approved and merged before implementing the fix?

```
The biggest failure mode in AI-assisted teams at scale: spec drift. The spec says X, the implementation does Y (because AI took a shortcut or the implementer adapted), and the spec is never updated. Within 3 months your specs are decoration and nobody trusts them.
```

specs/features/_template.md
```diff
@@ -0,0 +1,51 @@
...
+## Background
+
+<!-- What problem does this feature solve and why does it matter?
+     Describe the current state and its shortcomings. -->
+
+## Goals
+
+<!-- What must the feature achieve? State as outcomes, not implementation steps. -->
+
+-
```
oshmyheliuk: AI recommends to add Status field to the template
```
Status field in spec metadata — Draft | In Review | Approved | Implemented | Superseded; lets you see at a glance which features have approved specs vs. not
```

specs/conventions.md
```diff
@@ -0,0 +1,42 @@
+# Spec Conventions
+
```
oshmyheliuk, line 2: We can add a description of how detailed the spec should be, like not too vague and not too detailed, but I don't know what the ideal size for specs is )

specs/features/_template.md
```diff
@@ -0,0 +1,51 @@
...
+| Ticket                                                    | Description  |
+| --------------------------------------------------------- | ------------ |
+| [CEXT-XXXX](https://jira.corp.adobe.com/browse/CEXT-XXXX) | Initial spec |
+
+## Background
+
+<!-- What problem does this feature solve and why does it matter?
+     Describe the current state and its shortcomings. -->
+
+## Goals
```
oshmyheliuk: It is also recommended to add Non-Goals. section to the spec:


- Prevent Scope Creep: Non-goals prevent teams from over-engineering or building unplanned features that waste time.
- Set Boundaries: They draw a clear line between what is being built and what is not, ensuring everyone understands the project boundaries.
- Clarify Intent: They provide clarity on the purpose of the project, helping to avoid, for example, "No multi-language support in v1".
- Increase Efficiency: They reduce the time spent on unnecessary discussions and refactoring.

specs/features/capability-discovery.md
```diff
@@ -0,0 +1,79 @@
...
+The same hook will also generate a **filtered `openapi.json`** — a subset of the full spec containing only the endpoints that correspond to the active config domains — and emit it alongside the built app. The generated file includes the SDK version and the app version for traceability.
+
+Availability is determined by what the app developer declares in `app.commerce.config.ts`, not by which SDK modules are imported. The `pre-app-build` hook is the correct and only insertion point for generating the filtered spec, since it is the only place where both the full SDK spec and the app's config are available simultaneously.
+
+The filtering logic mirrors what already drives `ext.config.yaml` generation:
+
+| Config domain active                                                                 | Endpoints included     |
+| ------------------------------------------------------------------------------------ | ---------------------- |
+| Always                                                                               | `app-config`           |
+| `eventing.commerce` or `eventing.external` or `installation.customInstallationSteps` | `installation`         |
```
oshmyheliuk: Probably need to add `or webhooks, or admin ui sdk`

specs/features/capability-discovery.md
```diff
@@ -0,0 +1,79 @@
...
+| `businessConfig`                                                                     | `config`, `scope-tree` |
+
+### 3. Metadata action
+
+The filtered `openapi.json` is served by a dedicated `metadata` action (not a static file) protected by the same IMS authentication as the rest of the SDK endpoints. This avoids unauthenticated disclosure of deployment-specific configuration.
```
oshmyheliuk: The spec doesn't explain how the file gets into the action at runtime, or what the action's implementation actually does (reads the file or returns it inline? from where?).

specs/conventions.md
obarcelonap: Done

specs/features/_template.md
obarcelonap: My idea was to create separate PRs for spec and implementation. I added an implementation flag and added a section in conventions
https://github.com/adobe/aio-commerce-sdk/blob/179d70689958aae085c874808a390aa033a9dfb1/specs/conventions.md?plain=1#L44-L61

specs/conventions.md
obarcelonap: added in conventions a note

specs/features/_template.md
obarcelonap: added non-goals

specs/features/capability-discovery.md
obarcelonap: removed table as that will depend on the state of the sdk at implementation time. Added point in unresolved questions

specs/features/capability-discovery.md
obarcelonap: added

specs/conventions.md
obarcelonap: I rewrote the whole conventions file, please take a look now.

obarcelonap:
@oshmyheliuk thanks for the review, I did a major rework which included changing the spec template, please take a look.

specs/features/CEXT-6138-capability-discovery.md
```diff
@@ -0,0 +1,155 @@
...
+`openapi.json` with the `metadata` action.
+
+No manual spec maintenance is required. New SDK features are automatically included
+if their config domain is already active. If a new feature introduces a new config
+domain, the app developer opts in by declaring it in `app.commerce.config.ts` and
+rebuilding.
+
+## Drawbacks
+
+- Adds a new `metadata` runtime action to every SDK-based deployment.
```
oshmyheliuk: I remember we discussed how to reduce the number of runtime actions.
Maybe we can add a new endpoint to one of the already existing actions?

oshmyheliuk (approved):

specs/features/_template.md
```diff
@@ -0,0 +1,67 @@
...
+
+Explain the feature as if it were already shipped and you were showing it to an
+app developer using the SDK. This means:
+
+- What new concepts or APIs does it introduce?
+- Show concrete usage examples (config, code, CLI output).
+- How should developers think about this feature and how does it change how they
+  build apps?
+- If applicable, show what error messages or warnings look like.
+- Does this affect how developers read or maintain existing app code?
```
iivvaannxx: Does the AI understand what `app code` means without some context?

specs/features/_template.md
```diff
@@ -0,0 +1,67 @@
...
+- What is the impact of not doing this?
+- Could this be achieved in user-space (i.e. without SDK changes)?
+
+## Unresolved questions
+
+- What needs to be resolved through discussion before this spec is accepted?
+- What needs to be resolved during implementation before the feature ships?
+- What related problems are explicitly out of scope for this spec?
+
+## Future possibilities
```
iivvaannxx, line 63: How will this know about planned SDK features?

specs/features/_template.md
```diff
@@ -0,0 +1,67 @@
+# Feature Name
+
+- **Ticket:** [CEXT-XXXX](https://jira.corp.adobe.com/browse/CEXT-XXXX)
+- **Created:** YYYY-MM-DD
+- **Implemented:** [ ]
```
oshmyheliuk: btw: putting the brackets at the end will not create a functional checkbox in standard Markdown.

```
markdown
- [ ] This is an incomplete task
- [x] This is a completed task
```

specs/features/CEXT-6138-capability-discovery.md
iivvaannxx: The only place this would make sense to add is the `GET /app-config` which is already "taken". We can return it as additional properties maybe tho. For what is worth, if the `metadata` action is going to return purely static data, the build time should be close to 0. Maybe also the deploy time.

jcuerdo (approved):

specs/conventions.md
```diff
@@ -0,0 +1,76 @@
...
+whitespace.
+
+## Workflow
+
+A feature goes through two pull requests:
+
+1. **Spec PR** — contains only the spec file. Reviewers align on the design
+   before any implementation begins. Once merged, the spec is considered
+   approved and the feature is ready to implement.
+2. **Implementation PR** — contains the code. References the approved spec.
```
jcuerdo, line 61: At that point, is it a regular PR? Same rules?

forcha: AFAIK, yes

specs/features/CEXT-6138-capability-discovery.md
```diff
@@ -0,0 +1,155 @@
...
+obtain an IMS token, fetch the spec, adapt accordingly.
+
+## Design
+
+### 1. Full spec in the SDK package
+
+`aio-commerce-lib-app` ships an `openapi.json` (OpenAPI 3.x) documenting **all**
+endpoints the SDK can expose. This file is included in the published npm package
+under a well-known path (e.g. `dist/openapi.json`) and is the single source of
+truth for the SDK's API contract.
```
iivvaannxx, line 71: Not sure if we are planning here how would this would be implemented but thought that the below info would be useful. If we are planning on writing this OpenAPI manually, then forget about it.

I'm pretty sure this can be more or less automated though. The `HttpActionRouter` is built to accept any kind of schema via [Standard Schema](https://standardschema.dev/). There are ways to convert those schemas to Open API (e.g see [this package](https://www.npmjs.com/package/@standard-community/standard-openapi)). 

This means that we can make a script that loads that router info and uses the internal schemas that it has (for `body`, `query`...) and pass them through a pipeline that can produce that spec.

specs/features/_template.md
obarcelonap: fixed.

specs/features/_template.md
obarcelonap: rephrased.

specs/features/_template.md
obarcelonap: This section is to dump anything you know is coming that might interact with this feature, no need to be planned nor accurate. Hope it answers the question.

specs/features/CEXT-6138-capability-discovery.md
obarcelonap: what if we use `GET /app-config/openapi.json`? this is the same discussion we had about adding new actions for uninstallation process and we decided to keep the actions to the minimum.
@oshmyheliuk @iivvaannxx if you agree I change the spec with this approach.

iivvaannxx: Didn't think of that, sounds good to me

specs/features/CEXT-6138-capability-discovery.md
obarcelonap: added a point in "Unresolved questions" regarding this topic. Personally I'd start manually and automate later if is a quick win.

specs/features/CEXT-6138-capability-discovery.md
oshmyheliuk: It works, but it's a little confusing that `/app-config/openapi.json` will describe not only `/app-config` part but the whole project.
I'm ok with adding a new runtime action as well.

obarcelonap: Agree is not elegant but I'd like to avoid adding a new runtime action and `/app-config` is the runtime action we can guarantee will be always available. Will adapt the spec in this direction.

obarcelonap: done https://github.com/adobe/aio-commerce-sdk/pull/411/changes/e3e9f6bc4475365b1019f8679529c0593444b653

iivvaannxx (approved):

PR #412, opened 2026-04-23
 CEXT-6079: add unit tests for lib-app runtime actions

vinayrao2000:
CEXT-6079: add unit tests for lib-app runtime actions

## Description

Added a focused unit test suite for REST runtime actions in `@adobe/aio-commerce-lib-app` to cover handler behavior and regression scenarios.

### Added test files
- `packages/aio-commerce-lib-app/test/fixtures/actions.ts`
- `packages/aio-commerce-lib-app/test/unit/actions/app-config.test.ts`
- `packages/aio-commerce-lib-app/test/unit/actions/config.test.ts`
- `packages/aio-commerce-lib-app/test/unit/actions/scope-tree.test.ts`
- `packages/aio-commerce-lib-app/test/unit/actions/installation.test.ts`

### Coverage scope
- `appConfigRuntimeAction`
- `configRuntimeAction`
- `scopeTreeRuntimeAction`
- `installationRuntimeAction`

The tests are unit-level with mocked dependencies and no external service calls.

## Related Issue

CEXT-6079

## Motivation and Context

`lib-app` REST runtime action behavior needed direct automated coverage for request/response paths and error handling.  
This change improves confidence in action logic and catches regressions without requiring external integration setup.

## How Has This Been Tested?

Executed from repo root:

- `pnpm test`
- `pnpm typecheck`

Both completed successfully in the local workspace.

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        AIO_COMMERCE_CONFIG_ENCRYPTION_KEY: "encryption-key",
+      }),
+    );
+
+    expect(initializeMock).toHaveBeenCalledWith({ schema: configSchema });
+    expect(byScopeIdMock).toHaveBeenCalledWith("store-1");
+    expect(getConfigurationMock).toHaveBeenCalledWith(
+      { scopeId: "store-1" },
+      { encryptionKey: "encryption-key" },
+    );
```
iivvaannxx: These lines seem to assert implementation details that this test should not care about. I'd remove them

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
```diff
@@ -0,0 +1,224 @@
...
+          scopeId: "store-1",
+          config: [
+            { name: "apiKey", value: "*****" },
+            { name: "mode", value: "live" },
+          ],
+        },
+      }),
+    );
+
+    expect(initializeMock).toHaveBeenCalledWith({ schema: configSchema });
```
iivvaannxx: Idem on this initialize check. Don't think this test needs to care whether we call initialize or not.

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
```diff
@@ -0,0 +1,224 @@
...
+    expect(setConfigurationMock).toHaveBeenCalledWith(
+      {
+        config: [{ name: "mode", value: "live" }],
+      },
+      { scopeId: "store-1" },
+      { encryptionKey: undefined },
```
iivvaannxx: Let's assert only the parameters we want to actually check (in this case the first one). For others we can just `expect.any(Object)` or something like that. This way if those parameters are different in the future, these tests won't break.

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        Deprecation: "Wed, 15 Apr 2026 00:00:00 GMT",
+      },
+      body: {
+        scopeId: "store-1",
+        config: [
+          { name: "apiKey", value: "*****", origin: "scope" },
+          { name: "mode", value: "live", origin: "scope" },
+        ],
+      },
+    });
```
iivvaannxx: Similarly, I would not assert the result in this test. I would only do it in a test that is intended to do that.

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        schema: configSchema,
+        values: {
+          scopeId: "store-1",
+          config: [
+            { name: "apiKey", value: "*****", origin: "scope" },
+            { name: "mode", value: "live", origin: "scope" },
+          ],
+        },
+      },
+    });
```
iivvaannxx: You should probably only assert the `result` in a specific test for the 200 result.

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
```diff
@@ -0,0 +1,224 @@
...
+      },
+      { scopeId: "store-1" },
+      { encryptionKey: undefined },
+    );
+    expect(result).toEqual({
+      type: "success",
+      statusCode: 200,
+      headers: {
+        "Cache-Control": "no-store",
+        Deprecation: "Wed, 15 Apr 2026 00:00:00 GMT",
```
iivvaannxx: I would add a specific test only for this header, to ensure future updates don't delete it.

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        scopeId: "store-1",
+        config: [
+          { name: "apiKey", value: "*****", origin: "scope" },
+          { name: "mode", value: "live", origin: "scope" },
+        ],
+      },
+    });
+  });
+
+  test("forwards partial updates and null unsets with PATCH /", async () => {
```
iivvaannxx: This test it's a bit confusing to me. I would refactor it to two separate tests that assert the following:

- PUT doesn't accept `null` values
- PATCH does accept `null` values.

iivvaannxx (commented):
I only reviewed a couple of files for now. Probably the comments I added there also apply to the others. 

Let's keep the tests focused to testing very specific things instead of having macro tests that assert multiple different things. Also make sure that you only "assert" those things that the test title describes, otherwise it can generate confusion if that test starts failing because of a change that seems unrelated 👍🏻

jcuerdo (approved):

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
vinayrao2000, line 1: the changes has been done as per the feedback given please review and let me know the feedback. Based on that will do the changes for other files @iivvaannxx

packages/aio-commerce-lib-app/test/unit/actions/app-config.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    expect(result).toEqual({
+      type: "success",
+      statusCode: 200,
+      body: minimalValidConfig,
+    });
+  });
+
+  test("returns a 500 error when the app config is invalid", async () => {
+    const handler = appConfigRuntimeAction({
+      appConfig: {} as unknown as CommerceAppConfigOutputModel,
```
iivvaannxx: I suggest using `@ts-expect-error` on any test that needs working with invalid data, instead of this typical dual-cast.

packages/aio-commerce-lib-app/test/unit/actions/scope-tree.test.ts
```diff
@@ -0,0 +1,239 @@
...
+    expect(resolveCommerceHttpClientParamsMock).toHaveBeenCalledWith(
+      {
+        AIO_COMMERCE_AUTH_IMS_TOKEN: "ims-token",
+        AIO_COMMERCE_API_BASE_URL: "https://commerce.example.com",
+        AIO_COMMERCE_API_FLAVOR: "stage",
+        __ow_method: "post",
+        __ow_path: "/commerce",
+        __ow_body: JSON.stringify({
+          commerceBaseUrl: "https://commerce.example.com",
+          commerceEnv: "stage",
```
iivvaannxx: I would not assert the presence of these. The only thing we care about in this call is the first three (`AIO_COMMERCE_*`).

packages/aio-commerce-lib-app/test/unit/actions/installation.test.ts
```diff
@@ -0,0 +1,793 @@
...
+  ) => Promise<void>;
+  onStepSuccess: (
+    event: { stepName: string },
+    state: InstallationState,
+  ) => Promise<void>;
+  onStepFailure: (
+    event: StepFailedEvent,
+    state: InstallationState,
+  ) => Promise<void>;
+};
```
iivvaannxx: I think you don't need to recreate this type, you should be fine with [this one](https://github.com/adobe/aio-commerce-sdk/blob/310aa577befe1af5dd56b382d6671d9fc5dec14c/packages/aio-commerce-lib-app/source/management/installation/workflow/hooks.ts#L52)

packages/aio-commerce-lib-app/test/unit/actions/installation.test.ts
```diff
@@ -0,0 +1,793 @@
...
+          return installationStore;
+        }
+
+        if (prefix === "uninstallation") {
+          return uninstallationStore;
+        }
+
+        throw new Error(`Unexpected store prefix: ${String(prefix)}`);
+      },
+    );
```
iivvaannxx, line 113: I believe it would be good to have this mock abstracted into the `fixtures` folder. Together with the rest. But it's not a blocker, feel free to skip this one.

packages/aio-commerce-lib-app/test/unit/actions/installation.test.ts
```diff
@@ -0,0 +1,793 @@
...
+    isFailedStateMock.mockImplementation(
+      (state: InstallationState) => state.status === "failed",
+    );
+    isInProgressStateMock.mockImplementation(
+      (state: InstallationState) => state.status === "in-progress",
+    );
+    isSucceededStateMock.mockImplementation(
+      (state: InstallationState) => state.status === "succeeded",
+    );
```
iivvaannxx: I believe this mocks are applying the exact same logic as the functions they are mocking. Is there any value in that?

iivvaannxx (commented):
Looks a lot better! One extra thing I would suggest is grouping the different endpoints in different `describe` blocks so that they are more easily interpretable. Something like:

```ts
describe("installationRuntimeAction", () => {
  describe("GET /", () => { /* ... */ })
  describe("POST /", () => { /* ... */ })
  describe("GET /uninstallation", () => { /* ... */ })
  // ...
})
```

packages/aio-commerce-lib-app/test/unit/actions/app-config.test.ts
vinayrao2000: replaced the dual cast with @ts-expect-error as per suggestions

packages/aio-commerce-lib-app/test/unit/actions/installation.test.ts
vinayrao2000: Agreed, no value in that, dropped them

packages/aio-commerce-lib-app/test/unit/actions/installation.test.ts
vinayrao2000: As you suggested I am reusing the existing type InstallationHooks. Changes made to replace it accordingly

iivvaannxx (approved):

PR #416, opened 2026-04-27
CEXT-6117: Fix empty worker process causing `aio app build` to fail

iivvaannxx:
## Description

This PR does the following:

- Fix an issue where an empty `workerProcess: []` was generated for Admin UI and caused `aio app build` to fail. 
- Adds some test cases for iframe Admin UI SDK
- Adds a `TODO` note on some code that can be cleaned up after an upstream issue in `valibot` is solved


packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
```diff
@@ -24,30 +24,31 @@ const SANDBOX_VALUES = [
...
-        .every((value) =>
-          (SANDBOX_VALUES as readonly string[]).includes(value),
-        ),
+    (val) => val.split(" ").every(isSandboxValue),
     `sandbox must contain only single-space-separated values from: ${SANDBOX_VALUES.map((t) => `"${t}"`).join(", ")}`,
   ),
 );
 
+const ColumnAlignSchema = v.picklist(["left", "right", "center"]);
+const ViewButtonLevelSchema = v.picklist([-1, 0, 1]);
```
oshmyheliuk: What is the goal of moving these lines? )

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
```diff
@@ -58,12 +59,24 @@ const ViewButtonConfirmSchema = v.object({
...
+
+const IframeEnabledEntries = {
+  ...IframeBaseEntries,
+  displayIframe: v.literal(true),
   sandbox: v.optional(SandboxSchema),
 };
 
+const IframeDisabledEntries = {
+  ...IframeBaseEntries,
+  displayIframe: v.optional(v.literal(false)),
```
oshmyheliuk: After changes in this pr the error for the wrong type of `displayIframe` is 
```
Invalid type: Expected (true | (false | undefined)) but received 1
```
Was 
```
Expected a boolean value for 'displayIframe'
```

Not an issue, but the previous variant sounds better.

packages/aio-commerce-lib-app/source/commands/generate/actions/templates/admin-ui-sdk/registration.js.template
```diff
@@ -15,6 +15,8 @@
 
 import { registrationRuntimeAction } from "@adobe/aio-commerce-lib-app/actions/registration";
 
-// {{REGISTRATION_JSON}}
+// The registration config is always at this relative constant path from the action.
+import registration from "../../registration.json" with { type: "json" };
 
-export const main = registrationRuntimeAction({ registration });
+const args = { registration };
```
oshmyheliuk: Why add a redundant constant instead of passing directly to `registrationRuntimeAction`?

packages/aio-commerce-lib-app/source/commands/generate/actions/templates/admin-ui-sdk/registration.js.template
```diff
@@ -15,6 +15,8 @@
 
 import { registrationRuntimeAction } from "@adobe/aio-commerce-lib-app/actions/registration";
 
-// {{REGISTRATION_JSON}}
+// The registration config is always at this relative constant path from the action.
+import registration from "../../registration.json" with { type: "json" };
```
oshmyheliuk: I don't see why we need to create an additional file for this; it was working fine with everything in the action code.

iivvaannxx: Benefits are small, but I find them worth when considering maintainability for the app project. It also aligns with the other extension point generated files, keeping overall consistency.

- Separates code from data, which makes the generated action easier to read.
- Makes the registration payload easier to inspect as plain JSON (doesn't mix license, comments, imports...).
- Produces more stable output and cleaner diffs.
- Keeps the action file small even as the registration grows (registrations can grow up quite a bit).
- Makes tests less brittle by checking the loader and payload separately. Also you can check for strict equality more easily

> it was working fine

It works the same, just because something works doesn't mean it can't be improved (in my opinion). The old approach was a bit simpler operationally. But from a maintainability perspective, I think the separation is a net positive

packages/aio-commerce-lib-app/source/commands/generate/actions/templates/admin-ui-sdk/registration.js.template
iivvaannxx: This could be rewritten. In the `installation` action we do it like so that we can have a bit of a more "robust" string replacement of the arguments of the runtime action factory (when we have custom scripts). In here might not be needed.

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
iivvaannxx: Unintended, I don't recall doing that. Maybe it was the AI lol 😆 

If it bothers you I can revert

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
iivvaannxx: Agree though, will see if I can make Valibot use that message. If not, I think it's a good reason to keep your `partialCheck`.

iivvaannxx: I ended up reverting all the `iframe` changes. I wanted to keep your `partialCheck` trick as it was really nice but also avoid any type cast or explicit typing (for schemas I think it's better to let them flow naturally), but I just wasn't able to. I pulled every trick on the book (that I know) and I wasn't able to reconcile types. 

However, this is because of an upstream issue. I opened it in https://github.com/open-circle/valibot/issues/1459, in case you want to follow.

I left a `TODO` to cleanup that after the issue is solved

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
iivvaannxx: This has been undone with the revert I explained in [here](https://github.com/adobe/aio-commerce-sdk/pull/416/changes#r3153487502)

packages/aio-commerce-lib-app/source/commands/generate/actions/templates/admin-ui-sdk/registration.js.template
iivvaannxx: Done in [`71bcb82`](https://github.com/adobe/aio-commerce-sdk/pull/416/commits/71bcb8297adabb1a762fff8bd7a5f68ac29d3ff1)

obarcelonap (approved):
LGTM, no strong opinion on the import vs self contained discussion

packages/aio-commerce-lib-app/source/commands/generate/actions/templates/admin-ui-sdk/registration.js.template
iivvaannxx: @oshmyheliuk I also noticed that with this change we don't need a "custom" function to generate the template and we can rely purely on `generateActionFiles` (the function used by the other extension points). 

I also incorporated [that in the PR](https://github.com/adobe/aio-commerce-sdk/pull/416/commits/bf137b82b3dd5711980ed8ed492ea8e82db42eb6), let me know what you thoughts and whether you think it justifies the change a bit more or not.

oshmyheliuk: > It works the same, just because something works doesn't mean it can't be improved (in my opinion). The old approach was a bit simpler operationally. But from a maintainability perspective, I think the separation is a net positive

It's not an improvement in my opinion. The code is auto-generated and shouldn't be manually updated, so from the maintainability perspective it's easier to generate one file, than a bunch of files and then include them. 

> Separates code from data, which makes the generated action easier to read.

Yes, it's easier to read as it's empty and only returns configuration from another generated file. This requires finding and reading another file to understand what action returns.

> Makes the registration payload easier to inspect as plain JSON (doesn't mix license, comments, imports...).

How? The same content was in action as now in registration.json

> Produces more stable output and cleaner diffs.

What? How is it more stable? action code are auto-generated

> Keeps the action file small even as the registration grows (registrations can grow up quite a bit).

Keeps the action small for what? Just to have another big file?


For me, it looks like just a change for the sake of change, without real value.

oshmyheliuk (approved):

packages/aio-commerce-lib-app/source/commands/generate/actions/templates/admin-ui-sdk/registration.js.template
iivvaannxx: Reverted in [1c6bd36](https://github.com/adobe/aio-commerce-sdk/pull/416/commits/1c6bd3687f9292013874ff5bbc7a10de0e9cc5e3)

PR #418, opened 2026-05-01
CEXT-6080: add conditional schema properties based on commerce environment type

vinayrao2000:
## Description

Enables conditional rendering of business configuration fields based on the selected Commerce environment (`paas` or `saas`).

- Adds optional `env: ("paas" | "saas")[]` property to business config schema fields in `@adobe/aio-commerce-lib-config`. Fields without `env` apply to all environments; fields with `env` apply only to the listed environments.
- Updates the `configRuntimeAction` GET handler in `@adobe/aio-commerce-lib-app` to accept an optional `?commerceEnv=paas|saas` query parameter. When provided, the returned schema is filtered to fields scoped to that environment; otherwise the full schema is returned (backward compatible).

App Management forwards `commerceEnv` per request based on the selected instance, so apps can declare environment-specific fields and the form will only show the relevant ones.

## Related Issue

CEXT-6080

## Motivation and Context

App Builder apps often need configuration fields that only make sense for a specific Commerce environment. Previously the schema was environment-agnostic, so every user saw every field regardless of which Commerce instance they had selected. This change provides the SDK-side mechanism for environment-aware schemas; the companion change in `commerce-app-management` provides the per-request signal.

## How Has This Been Tested?

Unit tests in `aio-commerce-lib-config` cover:
- Schema validation accepts/rejects `env` correctly (single environment, multiple environments, missing, empty array, unknown value).

Unit tests in `aio-commerce-lib-app` cover the GET handler:
- Returns full schema when `commerceEnv` query parameter is omitted.
- Filters by `?commerceEnv=paas`.
- Filters by `?commerceEnv=saas`.
- Returns 400 on invalid `commerceEnv` value.

End-to-end manual verification on stage:
- PaaS instance → response excludes SaaS-only fields.
- SaaS instance → response excludes PaaS-only fields.
- Caller without `commerceEnv` → full schema returned, current behavior preserved.

`pnpm test`, `pnpm typecheck`, `pnpm lint` — all green.

## Screenshots

N/A — non-visual change.

## Types of changes

- [ ] Bug fix
- [ ] Breaking change

## Checklist

- [x] Adobe Open Source CLA
- [x] Code style
- [x] Documentation updated (`packages/aio-commerce-lib-config/docs/usage.md`)
- [x] Changesets added (`@adobe/aio-commerce-lib-config: minor`, `@adobe/aio-commerce-lib-app: minor`)
- [x] Tests added
- [x] All tests pass.


packages/aio-commerce-lib-app/source/actions/config.ts
```diff
@@ -91,7 +109,16 @@ router.get("/", {
       "businessConfig.schema",
     );
 
-    initialize({ schema: validatedSchema });
+    const flavor =
+      req.query.commerceEnv ??
+      normalizeCommerceFlavor(rawParams.commerceEnv) ??
+      normalizeCommerceFlavor(rawParams.AIO_COMMERCE_API_FLAVOR);
```
iivvaannxx: This is not necessary, the schema in the `query` will already validate `commerceEnv` and return a 400 if not correct. Given that we will be invoking this from the UI only we don't need the `AIO_COMMERCE_API_FLAVOR` neither, as it's an env that would need to be set in the `inputs` of the application project, which doesn't have that information.

packages/aio-commerce-lib-config/docs/usage.md
```diff
@@ -305,19 +305,6 @@ Use text fields for free-form input values like merchant identifiers or custom s
...
-
-```javascript
-{
-  name: "enabled",
-  label: "Enable Feature",
-  type: "boolean",
-  default: true
-}
-```
-
```
iivvaannxx, line 320: Why is this deleted? Maybe a mistake?

packages/aio-commerce-lib-config/docs/usage.md
```diff
@@ -419,6 +406,44 @@ Multiple selection example:
...
+
+```typescript
+import { filterBusinessConfigSchemaByFlavor } from "@adobe/aio-commerce-lib-config";
+
+const saasFields = filterBusinessConfigSchemaByFlavor(schema, "saas");
+const paasFields = filterBusinessConfigSchemaByFlavor(schema, "paas");
+```
+
+Fields without `env` are always included. Field order is preserved.
+
```
iivvaannxx: Let's be consistent on `env` vs `flavor` and not mix the two.

packages/aio-commerce-lib-config/source/modules/schema/fields.ts
```diff
@@ -26,6 +47,7 @@ const BaseOptionSchema = v.object({
   description: v.optional(
     v.string("Expected a string for the field description"),
   ),
+  env: v.optional(EnvSchema),
```
iivvaannxx, line 54: this should apply a default of both envs? or is that done somewhere else?

packages/aio-commerce-lib-config/source/index.ts
```diff
@@ -39,5 +42,6 @@ export type {
   BusinessConfigSchemaField,
   BusinessConfigSchemaListOption,
   BusinessConfigSchemaValue,
+  CommerceFlavor,
```
iivvaannxx: I believe we export that variable from somewhere else, I don't think it needs to be part of the public API of the package.

packages/aio-commerce-lib-config/test/unit/modules/schema/utils.test.ts
```diff
@@ -80,5 +85,92 @@ describe("schema/utils", () => {
...
+    });
+
+    test("should reject an env entry that is not a known flavor", () => {
+      expect(() =>
+        validateBusinessConfigSchema([
+          {
+            name: "scoped",
+            type: "text",
+            label: "Scoped",
+            env: ["onprem" as never],
```
iivvaannxx:
```suggestion
            // @ts-expect-error - Testing an invalid value on purpose
            env: ["onprem"],
```

packages/aio-commerce-lib-config/source/index.ts
```diff
@@ -25,7 +25,10 @@ export {
   type SelectorByCodeAndLevel,
   type SelectorByScopeId,
 } from "./config-utils";
-export { SchemaBusinessConfig } from "./modules/schema";
+export {
+  filterBusinessConfigSchemaByFlavor,
```
iivvaannxx: I would not add this to the public API, it's not something a consumer should need at all. If it actually needs it, we can add it in the future. For now, I'd move the filtering logic into `lib-app` and keep it just there.

packages/aio-commerce-lib-config/source/modules/schema/fields.ts
vinayrao2000: env is optional field, when omitted the field applies to all flavors. The filter handles that case (field.env === undefined || field.env.includes(flavor)), so no default is needed.

packages/aio-commerce-lib-config/source/index.ts
vinayrao2000: As suggested removed the export from public API

packages/aio-commerce-lib-config/source/index.ts
vinayrao2000: As suggested filter logic moved to lib-app

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      statusCode: 200,
+    });
+
+    if (result.type !== "success") {
+      throw new Error("Expected a successful response");
+    }
+
+    if (!result.body) {
+      throw new Error("Expected response body");
+    }
```
iivvaannxx:
```suggestion
    expect.assert(result.type === "success");
    expect.assert(result.body);
```

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
```diff
@@ -0,0 +1,195 @@
...
```
iivvaannxx:
```suggestion
    expect.assert(result.type === "success");
    expect.assert(result.body);
```

packages/aio-commerce-lib-config/source/modules/schema/fields.ts
```diff
@@ -16,6 +16,27 @@ const DEFAULT_BOOLEAN_VALUE = false as const;
...
+ * specific Commerce flavors. When omitted, the field applies to all flavors.
+ */
+const EnvSchema = v.pipe(
+  v.array(
+    CommerceFlavorSchema,
+    "Expected an array of Commerce flavors for the field env",
+  ),
+  v.nonEmpty("The env array must contain at least one Commerce flavor"),
+);
+
```
iivvaannxx: Let's call it just commerce env, forget about the "flavor" thing, might be confusing for some developers. Try to keep it consistent in all comments/messages.

packages/aio-commerce-lib-app/source/actions/config.ts
```diff
@@ -52,6 +52,25 @@ interface ConfigActionContext extends BaseContext {
 // Placeholder value for password fields.
 const MASKED_PASSWORD_VALUE = "*****";
 
+/** The set of valid Commerce flavors a configuration field can be scoped to. */
+type CommerceFlavor = "paas" | "saas";
```
iivvaannxx:
```suggestion
type CommerceEnv = "paas" | "saas";
```

packages/aio-commerce-lib-app/source/actions/config.ts
```diff
@@ -52,6 +52,25 @@ interface ConfigActionContext extends BaseContext {
...
+ * @param flavor - The Commerce flavor to filter by.
+ */
+function filterSchemaByFlavor(
+  schema: BusinessConfigSchema,
+  flavor: CommerceFlavor,
+): BusinessConfigSchema {
+  return schema.filter(
+    (field) => field.env === undefined || field.env.includes(flavor),
+  );
+}
```
iivvaannxx, line 73:
```suggestion
/**
 * Filters a business configuration schema to the fields applicable to the
 * given Commerce environment. Fields without an `env` property apply to all
 * environments and are always included.
 * @param schema - The business configuration schema to filter.
 * @param env - The Commerce environment to filter by.
 */
function filterSchemaByEnv(
  schema: BusinessConfigSchema,
  env: CommerceEnv,
): BusinessConfigSchema {
  return schema.filter(
    (field) => field.env === undefined || field.env.includes(env),
  );
}
```

iivvaannxx (approved):
@vinayrao2000 Approving with some final suggestions. Should be minor changes. Good job! 👍🏻

packages/aio-commerce-lib-config/docs/usage.md
```diff
@@ -419,6 +419,33 @@ Multiple selection example:
 > [!NOTE]
 > For `selectionMode: "multiple"`, the `default` value must be an array of strings, even if only one option is selected by default.
 
+### Conditional Fields by Commerce Flavor
+
+Each schema field accepts an optional `env` property to scope it to specific Commerce flavors. The value is an array of flavors (`"paas"`, `"saas"`) and accepts any combination of one or more.
+
+When `env` is omitted, the field applies to all Commerce flavors. When `env` is set, the field is only relevant for the listed flavors and can be filtered out for the others — for example, when rendering the configuration form in App Management for a SaaS Commerce instance.
```
iivvaannxx: Remember to update terminology to work only with the concept of "environment" and flavor

packages/aio-commerce-lib-app/source/actions/config.ts
vinayrao2000: changed as per suggestion

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
vinayrao2000: changed as per suggestion

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
vinayrao2000: changed as per suggestion

packages/aio-commerce-lib-config/docs/usage.md
vinayrao2000: changed as per suggestion

packages/aio-commerce-lib-config/source/modules/schema/fields.ts
```diff
@@ -16,6 +16,27 @@ const DEFAULT_BOOLEAN_VALUE = false as const;
...
+ * specific Commerce Env. When omitted, the field applies to all Env.
+ */
+const EnvSchema = v.pipe(
+  v.array(
+    CommerceEnvSchema,
+    "Expected an array of Commerce Env for the field env",
+  ),
+  v.nonEmpty("The env array must contain at least one Commerce Env"),
+);
+
```
iivvaannxx, line 39: When a message is user-facing, like these validation error messages, try to not use abbreviations, unless you're referring to a literal option.

```suggestion
const EnvSchema = v.pipe(
  v.array(
    CommerceEnvSchema,
    'Expected an array of commerce environments for the field "env"',
  ),
  v.nonEmpty('The "env" array must contain at least one commerce environment'),
);

```

packages/aio-commerce-lib-app/test/unit/actions/config.test.ts
```diff
@@ -0,0 +1,185 @@
...
+  test("returns full schema when commerceEnv is omitted", async () => {
+    const handler = configRuntimeAction({ configSchema: schema });
+
+    const result = await handler({
+      __ow_method: "get",
+      __ow_path: "/",
+      __ow_query: "scopeId=scope-1",
+      __ow_headers: {},
+      LOG_LEVEL: "debug",
+    } as unknown as RuntimeActionParams);
```
iivvaannxx: I think you probably don't need this cast. Does TypeScript complain if you remove `as unknown as RuntimeActionParams`?

iivvaannxx:
Oh, and important, please add a Changeset! See [this section](https://github.com/adobe/aio-commerce-sdk/blob/main/.github/DEVELOPMENT.md#release-process) from the `DEVELOPMENT.md` or ask an LLM to do it for you.

packages/aio-commerce-lib-config/source/modules/schema/fields.ts
vinayrao2000: changed as per suggestion

.changeset/hungry-heads-decide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@adobe/aio-commerce-lib-config": minor
+"@adobe/aio-commerce-lib-app": minor
+---
+
+Add optional `env` property on configuration schema fields to scope them to specific Commerce environments (`"paas"`, `"saas"`). The config runtime action accepts a `commerceEnv` query parameter to filter the returned schema to fields applicable to the requested environment; fields without `env` apply to all environments.
```
iivvaannxx: This should be two separate changesets. This is because the "config runtime action" is something that is specific only to `lib-app`, `lib-config` doesn't know anything about that, so we can't have a Changeset with that kind of message targeting `lib-config` as then the CHANGELOG will reference concepts that users of the library might not know about.

The correct way to create Changesets for this PR would be:

- 1 changeset for `lib-app` explaining the new `commerceEnv` query parameter
- 1 changeset for `lib-config` explaining the new `env` property to configuration fields

iivvaannxx (approved):

.changeset/silver-pandas-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": minor
+---
+
+The business configuration runtime action's `GET /` endpoint now accepts an optional `commerceEnv` query parameter (`"paas" | "saas"`). When provided, the returned schema is filtered to only include fields scoped to that environment via the `env` property. Fields without an `env` property are always included.
```
iivvaannxx: This leaks some unnecessary implementation details. Keep it short as if it was a description to a user of the library. A user doesn't care how it works or whether it's the `GET` endpoint that supports this.

Something like:

```suggestion
Support ability to filter configuration fields of applications by the Commerce environment where they are installed.
```

.changeset/hungry-heads-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-config": minor
+---
+
+Add optional `env` property on business configuration schema fields to scope them to specific Commerce environments (`"paas"`, `"saas"`). Fields without `env` apply to all environments.
```
iivvaannxx: Idem on leaking implementation details.

```suggestion
Add optional `env` property on business configuration schema fields to scope them to specific Commerce environments (`"paas"`, `"saas"`).
```

vinayrao2000: Updated as suggested

.changeset/hungry-heads-decide.md
vinayrao2000: Updated as suggested

.changeset/silver-pandas-march.md
vinayrao2000: Updated as suggested

packages/aio-commerce-lib-config/source/modules/schema/fields.ts
```diff
@@ -16,6 +16,27 @@ const DEFAULT_BOOLEAN_VALUE = false as const;
...
+ */
+const EnvSchema = v.pipe(
+  v.array(
+    CommerceEnvSchema,
+    'Expected an array of commerce environments for the field "env"',
+  ),
+  v.nonEmpty('The "env" array must contain at least one commerce environment'),
+);
+
 /** Base schema for configuration field options with name, optional label, and optional description */
```
oshmyheliuk: JSDoc can be updated as a new field was added

oshmyheliuk (approved):

packages/aio-commerce-lib-config/test/fixtures/configuration-schema.ts
```diff
@@ -97,6 +97,24 @@ export const VALID_CONFIGURATION = [
...
+    env: ["saas"],
+  },
+  {
+    name: "paasOnlyToken",
+    type: "text",
+    label: "PaaS Token",
+    env: ["paas"],
+  },
+  {
+    name: "bothFlavorsField",
```
oshmyheliuk: The "Flavor" terminology can also be removed from tests in this and other files in this PR

oshmyheliuk: As it was already raised in the other comments to this PR

packages/aio-commerce-lib-config/source/modules/schema/fields.ts
vinayrao2000: updated the JSDoc

packages/aio-commerce-lib-config/test/fixtures/configuration-schema.ts
vinayrao2000: renamed and updated the related description

PR #423, opened 2026-05-05
CEXT-6077: Function Schema Utils

iivvaannxx:
## Description

Implement `syncFunctionSchema`, `asyncFunctionSchema` and `syncOrAsyncFunctionSchema` Valibot utilities in private package as a pre-work for the dynamic schema. Also add a `withPrefixedMessage` utility to modify any schema to have an arbitrary string prefixed as extra information on issues (to improve debugging experience).

Function validation with Valibot requires some manual and repetitive scaffolding that this helpers intend to solve. They provide helpful shorthands to create different schemas for both sync and async functions with optional arguments and return values, and enhanced error messages when validation happens.


packages-private/common-utils/source/valibot/actions/returns-sync-or-async.ts
iivvaannxx, line 1: For anyone revising this, I didn't find a way to make Valibot infer the return type of a function to be `T | Promise<T>`, which is the type we need for dynamic schemas (a factory function can be either sync or async). I needed to throw this in following [Valibot's extensibility guide](https://valibot.dev/guides/extend-valibot).

packages-private/common-utils/source/valibot/schemas/functions/async.ts
```diff
@@ -0,0 +1,164 @@
...
+  args?: undefined;
+  output: OutputSchema;
+  message?: string;
+}): ReturnType<typeof createAsyncFunctionSchemaWithOutput<OutputSchema>>;
+
+export function asyncFunctionSchema(options?: {
+  args?: undefined;
+  output?: undefined;
+  message?: string;
+}): ReturnType<typeof createAsyncFunctionSchema>;
```
iivvaannxx, line 125: Overloads are needed because return types differ based on arguments sent in. This applies to all three types of function schemas (`async`, `sync` or `sync-or-async`).

PR #426, opened 2026-05-05
[ON HOLD] CEXT-6151: Add ACL admin UI schema and helper to check permissions

oshmyheliuk:
## Description

https://jira.corp.adobe.com/browse/CEXT-6151

- Update adminUISdk schema to support ACL fields for menu items
-  Add a public helper that extensions can call from the SPA and from runtime actions to check whether the current user is allowed to use a given ACL resource.

## Related Issue

https://github.com/magento-commerce/adobe-commerce-backend-uix/pull/352

## Types of changes

## Checklist:


obarcelonap:
Lib-api is designed as generic Commerce API utility library, this PR adds Admin UI SDK code.

The precedent is Commerce Events: eventing is wired up during installation (lib-app), the actual API client lives in its own dedicated lib-events package. Admin UI SDK is the same pattern, lib-app owns the config schema and the installation step, but the permission client is a standalone API concern.

I'd spawn a new package dedicated for Admin UI SDK.

The cost is another package to maintain, but given that Admin UI SDK is still growing and under refinement, doing the split now is lower effort than after more surface has accumulated across lib-app and lib-api.

packages/aio-commerce-lib-api/source/lib/commerce/admin-ui-sdk-permissions/types.ts
```diff
@@ -0,0 +1,46 @@
...
+export interface AdminUiSdkPermissionClientOptions {
+  /** Milliseconds to cache a permission result. Default: 300_000 (5 minutes). Set to 0 to disable caching. */
+  cacheTtlMs?: number;
+  /** Return false instead of throwing when a network or parse error occurs. Default: true. */
+  denyOnError?: boolean;
+  httpClient: AdobeCommerceHttpClient;
+}
+
+/** Client for checking the current user's Admin UI SDK resource permissions. */
+export interface AdminUiSdkPermissionClient {
```
obarcelonap: Wouldn't be better to do not limit the scope to just permissions? What if we need to add in the future an operation for AdminUiSdk but not related to permissions? We will create a new client?

I'd propose to open the scope of the client, so other concepts fit in:
```
interface AdminUiSdkClient {
  checkPermission()
  invalidatePermission()
  requirePermission()
}
```

packages/aio-commerce-lib-api/source/lib/commerce/admin-ui-sdk-permissions/with-permission.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import type { AdminUiSdkPermissionClient } from "./types";
+
+const DENIED_RESPONSE = {
+  statusCode: 403,
+  body: { error: "Forbidden" },
+} as const;
+
+/** Wraps an App Builder action handler with an Admin UI SDK permission check. */
+export function withAdminUiSdkPermission<
```
obarcelonap: I like this idea but not sure if we have this kind of wrappers anywhere else

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
```diff
@@ -235,13 +235,39 @@ const BannerNotificationSchema = v.object({
   orderViewButtons: v.optional(v.array(OrderViewButtonBannerSchema)),
 });
 
+const ACL_RESOURCE_ID_REGEX =
+  /^[A-Z][A-Za-z0-9]+_[A-Z][A-Za-z0-9]+::[A-Za-z_0-9]+$/;
+const ACL_RESOURCE_ID_REGEX_MESSAGE =
+  "must follow the Vendor_Module::resource_name format";
```
obarcelonap, line 240: Please align the msg to what is usually used
for module names i'd say PascalCase: Vendor_Module -> VendorModule
for resource I believe snake_case is usually used so I think is fine

obarcelonap (commented):
LGTM but I'd like to hear your opinion about spawning a new package before approving.

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
oshmyheliuk: What do you mean by aligning? This is a format required by Commerce. If you try to put something like `VendorModule::resource_name` in the ACL Resource name, there will be an exception in the Commerce application.

packages/aio-commerce-lib-api/source/lib/commerce/admin-ui-sdk-permissions/with-permission.ts
oshmyheliuk: Yeah, I wasn't too sure whether I should add this. This wrapper is optional, I can remove it.

packages/aio-commerce-lib-api/source/lib/commerce/admin-ui-sdk-permissions/types.ts
oshmyheliuk: We already have an admin ui sdk related operation, but they are using the Commerce Client directly.
https://github.com/adobe/aio-commerce-sdk/blob/main/packages/aio-commerce-lib-app/source/management/installation/admin-ui-sdk/helpers.ts#L36
Probably can be moved to the "new Client" as well.

oshmyheliuk:
Creating a new package sounds like a logical step to me, to align with what we already have for eventing/webhooks.
I can update this PR with related changes.

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
obarcelonap: My bad, I though pascal case was used for module name.

packages/aio-commerce-lib-api/source/lib/commerce/admin-ui-sdk-permissions/with-permission.ts
obarcelonap: Maybe we could just create a require helper function instead of wrapping.
```
requireAdminUiSdkPermission("Acme_Promotions::edit", client)
```
Does it make sense change resource as array? In case of multiple permissions are allowed? Or even multiple permissions could be required?

packages/aio-commerce-lib-api/docs/usage.md
```diff
@@ -178,6 +178,74 @@ export const main = async function (params) {
...
+  getAdminUiSdkPermissionClient,
+  resolveCommerceHttpClientParams,
+  withAdminUiSdkPermission,
+} from "@adobe/aio-commerce-lib-api";
+
+async function updatePromotion(params) {
+  return { statusCode: 200, body: { updated: true } };
+}
+
+export const main = async (params) => {
```
obarcelonap: I overlooked the wrapper I thought it was meant to be used wrapping the whole function so you don't need to create the clients.
```typescript
export const main = withAdminUiSdkPermission(
    "Acme_Promotions::edit", 
    (params) => {
        // protected handler here
    },
)
```

packages/aio-commerce-lib-admin-ui-sdk/docs/usage.md
```diff
@@ -0,0 +1,111 @@
+# `@adobe/aio-commerce-lib-admin-ui-sdk` Documentation
```
obarcelonap: should we add experimental warning

packages/aio-commerce-lib-admin-ui-sdk/docs/usage.md
```diff
@@ -0,0 +1,111 @@
...
+```typescript
+import { createAdminUiSdkApiClient } from "@adobe/aio-commerce-lib-admin-ui-sdk/api";
+
+const client = createAdminUiSdkApiClient({
+  baseUrl: "https://commerce.example.com",
+  // ...other CommerceHttpClientParams
+});
+```
+
+In install/uninstall actions where only a subset of operations is needed, prefer `createCustomAdminUiSdkApiClient` to keep the bundle lean:
```
obarcelonap: While I understand this is to be more performant I don't think it is an issue in the app-management endpoints since they're not in the critical path of the applications. Therefore I'd remove it for the sake of simplicity.

packages/aio-commerce-lib-admin-ui-sdk/package.json
```diff
@@ -0,0 +1,96 @@
...
+    "#*": "./source/*.ts",
+    "#test/*": "./test/*.ts"
+  },
+  "files": [
+    "dist",
+    "package.json",
+    "CHANGELOG.md",
+    "README.md"
+  ],
+  "scripts": {
```
obarcelonap, line 61: Missing docs script and @aio-commerce-sdk/config-typedoc devDependency

packages/aio-commerce-lib-app/docs/usage.md
obarcelonap, line 1: review this docs because claude is flagging some ids may not work with the new validations
```
docs/usage.md in lib-app still has "my-app::menu" — which now fails the new regex
  packages/aio-commerce-lib-app/docs/usage.md:476 still shows a hyphenated ID. The new id regex /^[A-Za-z0-9/:_]+$/ disallows hyphens, so any user copying from the docs
  would get an unexpected parse failure. The example needs updating (e.g. "my_app::menu"). If real apps use hyphenated IDs this is potentially a major bump.
```

packages/aio-commerce-lib-admin-ui-sdk/docs/usage.md
oshmyheliuk: Added warning to the docs

packages/aio-commerce-lib-admin-ui-sdk/docs/usage.md
oshmyheliuk: Removed

packages/aio-commerce-lib-admin-ui-sdk/package.json
oshmyheliuk: Docs script was has been added

packages/aio-commerce-lib-app/docs/usage.md
oshmyheliuk: Thanks for spotting this, the docs were updated according to the regex, where hyphens are not allowed.

packages/aio-commerce-lib-admin-ui-sdk/package.json
```diff
@@ -0,0 +1,98 @@
...
+  },
+  "imports": {
+    "#*": "./source/*.ts",
+    "#test/*": "./test/*.ts"
+  },
+  "files": [
+    "dist",
+    "package.json",
+    "CHANGELOG.md",
+    "README.md"
```
obarcelonap, line 59: This is missing, could you add a README to the new package? Something like https://github.com/adobe/aio-commerce-sdk/tree/main/packages/aio-commerce-lib-app

packages/aio-commerce-lib-app/source/management/installation/admin-ui-sdk/helpers.ts
```diff
@@ -16,34 +16,24 @@ import { throwHttpError } from "../utils/http-error";
...
-    .post("adminuisdk/extension", {
-      json: {
-        extension: {
-          extensionName: process.env.__OW_NAMESPACE,
-          extensionTitle: appData.projectTitle,
-          extensionUrl: `https://${process.env.__OW_NAMESPACE}.adobeio-static.net/index.html`,
-          extensionWorkspace: appData.workspaceName,
-        },
-      },
+  await adminUiSdkClient
```
obarcelonap: I'd probably throw if `process.env.__OW_NAMESPACE` is not available. I know won't ever happen but using empty string as default feels is not correct.

packages/aio-commerce-lib-app/source/management/installation/admin-ui-sdk/utils.ts
```diff
@@ -27,32 +26,40 @@ export type AdminUiSdkConfig = CommerceAppConfigOutputModel & {
   adminUiSdk: NonNullable<CommerceAppConfigOutputModel["adminUiSdk"]>;
 };
 
-/** Context shared across Admin UI SDK steps. */
+function createAdminUiSdkClient(params: RuntimeActionParams) {
```
obarcelonap: this is very useful function, I feel we have the same for events clients. Should we export it and add it to usage.md?

packages/aio-commerce-lib-admin-ui-sdk/package.json
```diff
@@ -0,0 +1,98 @@
+{
+  "name": "@adobe/aio-commerce-lib-admin-ui-sdk",
```
obarcelonap: In #480 @danipt is suggesting to drop sdk term (see https://github.com/adobe/aio-commerce-sdk/pull/480#discussion_r3324685654). I'd rename to this package to `@adobe/aio-commerce-lib-admin-ui`

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
```diff
@@ -235,13 +235,39 @@ const BannerNotificationSchema = v.object({
...
+    nonEmptyStringValueSchema(label),
+    v.regex(ACL_RESOURCE_ID_REGEX, `${label} ${ACL_RESOURCE_ID_REGEX_MESSAGE}`),
+  );
+}
+
+const AclResourceSchema = v.object({
+  id: aclResourceIdSchema("ACL resource ID"),
+  title: nonEmptyStringValueSchema("ACL resource title"),
+  parent: v.optional(aclResourceIdSchema("ACL resource parent")),
+  sortOrder: v.optional(positiveNumberValueSchema("ACL resource sortOrder")),
```
obarcelonap, line 253: `positiveNumberValueSchema` should probably be used in `sortOrder` of `MenuItemSchema`

obarcelonap (approved):
No blockers

.changeset/add-admin-ui-sdk-library.md
```diff
@@ -0,0 +1,7 @@
+---
+"@adobe/aio-commerce-lib-admin-ui-sdk": minor
+"@adobe/aio-commerce-sdk": minor
+---
+
+Add `@adobe/aio-commerce-lib-admin-ui-sdk` library for checking Admin UI SDK ACL resource permissions.
+Add Admin UI SDK API exports to `@adobe/aio-commerce-sdk`.
```
iivvaannxx: These should be two separate changesets. Otherwise the CHANGELOG of `admin-ui-sdk` will include this line about "adding exports to aio-commerce-sdk", which should not.

packages/aio-commerce-lib-admin-ui-sdk/docs/usage.md
```diff
@@ -0,0 +1,98 @@
+# `@adobe/aio-commerce-lib-admin-ui-sdk` Documentation
+
+> **Experimental:** This package is not yet production-ready. The API may change in future releases.
```
iivvaannxx: Just so that it stands out a bit better as a warning.
```suggestion
> [!WARNING]
> **Experimental**: This package is not yet production-ready. The API may change in future releases.
```

packages/aio-commerce-lib-admin-ui-sdk/source/api/extensions/endpoints.ts
```diff
@@ -0,0 +1,72 @@
...
+ * @param httpClient - The {@link AdobeCommerceHttpClient} to use to make the request.
+ * @param params - The extension registration parameters.
+ * @param fetchOptions - Optional Ky fetch options.
+ *
+ * @throws An `HTTPError` if the status code is not 2XX.
+ */
+export async function registerExtension(
+  httpClient: AdobeCommerceHttpClient,
+  params: ExtensionRegistrationParams,
+  fetchOptions?: Options,
```
iivvaannxx, line 41: Probably not urgent, but be aware of #462 which changes this type to decouple our methods from `ky` (which was blocking a major library update). If the other PR merges first you'll need to update it

packages/aio-commerce-lib-admin-ui-sdk/source/api/extensions/schema.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import * as v from "valibot";
+
+/** Parameters for POST /V1/adminuisdk/extension. */
+export const ExtensionRegistrationParamsSchema = v.object({
+  extensionName: v.string(),
+  extensionTitle: v.string(),
+  extensionUrl: v.string(),
+  extensionWorkspace: v.string(),
+});
```
iivvaannxx: Should we add a bit more strict validation? Or is this done by the Commerce module? Such as `v.url` for the `extensionUrl` or some regex patterns/length checks for those that we know. I would argue it's useful to catch errors as early as possible (avoiding network round-trips).

packages/aio-commerce-lib-admin-ui-sdk/source/api/extensions/endpoints.ts
```diff
@@ -0,0 +1,72 @@
...
+): Promise<void> {
+  const { workspaceName, extensionName } = parseOrThrow(
+    UnregisterExtensionParamsSchema,
+    params,
+  );
+
+  await httpClient.delete(
+    `adminuisdk/extension/${workspaceName}/${extensionName}`,
+    fetchOptions,
+  );
```
iivvaannxx: I did not verify this, but I think that when you want to return `Promise<void>` you need to use `.then(() => {})` with an empty handler, like in [here](https://github.com/adobe/aio-commerce-sdk/blob/cffcc6af58c734486de4447c1f1eaa7a4784a4e2/packages/aio-commerce-lib-events/source/commerce/api/event-subscriptions/endpoints.ts#L103-L107)

packages/aio-commerce-lib-admin-ui/source/lib/permission-client.ts
```diff
@@ -0,0 +1,185 @@
...
+      if (denyOnError) {
+        return {
+          error: toPermissionError(error),
+          cacheable: false,
+        };
+      }
+
+      throw toPermissionError(error);
+    }
+  }
```
iivvaannxx, line 109: Not a blocker, can be done in the future, but would like to see this "cache" feature built directly into `lib-api` or HTTP clients for better reusability.

packages/aio-commerce-lib-admin-ui-sdk/source/index.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/** biome-ignore-all lint/performance/noBarrelFile: This is the `@adobe/aio-commerce-lib-admin-ui-sdk/api` entrypoint. */
+
+export * from "./api/index";
```
iivvaannxx: I would not double-export the `/api` entrypoint here. I'd agree it's better for convenience, but it might affect bundle size and it provides two different access paths for the same helpers.

packages/aio-commerce-lib-admin-ui/source/lib/permission-client.ts
```diff
@@ -0,0 +1,185 @@
...
+    const trackedPromise = promise
+      .then((result) => {
+        if (
+          cacheTtlMs > 0 &&
+          result.cacheable &&
+          inFlight.get(resource) === trackedPromise
+        ) {
+          cache.set(resource, {
+            value: result.allowed,
+            expiresAt: Date.now() + cacheTtlMs,
```
iivvaannxx, line 139: Is this `expiration` enforced anywhere for revalidation?

packages/aio-commerce-lib-admin-ui-sdk/package.json
iivvaannxx: Don't have a strong opinion here, but isn't the Commerce module named `backend-uix` or something like that? Maybe it's not as explicit as "admin-ui", but would help people correlate them more easily.

packages/aio-commerce-lib-admin-ui-sdk/package.json
```diff
@@ -0,0 +1,98 @@
...
+  "bugs": {
+    "url": "https://github.com/adobe/aio-commerce-sdk/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/adobe/aio-commerce-sdk.git",
+    "directory": "packages/aio-commerce-lib-admin-ui-sdk"
+  },
+  "publishConfig": {
+    "exports": {
```
iivvaannxx, line 33: I believe you're missing the root entrypoint of `source/index.ts`?

packages/aio-commerce-lib-admin-ui/package.json
```diff
@@ -0,0 +1,98 @@
...
+  },
+  "devDependencies": {
+    "@aio-commerce-sdk/common-utils": "workspace:*",
+    "@aio-commerce-sdk/config-tsdown": "workspace:*",
+    "@aio-commerce-sdk/config-typedoc": "workspace:*",
+    "@aio-commerce-sdk/config-typescript": "workspace:*",
+    "@aio-commerce-sdk/config-vitest": "workspace:*",
+    "@aio-commerce-sdk/scripting-utils": "workspace:*",
+    "@aio-commerce-sdk/scripts": "workspace:*",
+    "msw": "catalog:",
```
iivvaannxx, line 96: Is this dependency used? I believe it's for integration tests, but I didn't see any of those, only unit.

packages/aio-commerce-lib-admin-ui-sdk/vitest.config.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { baseConfig } from "@aio-commerce-sdk/config-vitest/vitest.config.base";
+import { defineConfig, mergeConfig } from "vitest/config";
+
+const BARREL_FILES = ["./source/index.ts"];
```
iivvaannxx, line 16: This can also use a glob to any `index.ts`, as there are currently two of them.

packages/aio-commerce-lib-app/source/management/installation/admin-ui-sdk/utils.ts
iivvaannxx: If we were to do this I would do it from each respective package, as it has nothing to do with any `lib-app` concern.

packages/aio-commerce-sdk/source/admin-ui-sdk/api.ts
iivvaannxx, line 1: This is also missing the `admin-ui-sdk` root entrypoint.

iivvaannxx (approved):

packages/aio-commerce-lib-admin-ui-sdk/docs/usage.md
iivvaannxx: I'd argue against removing it. I don't think it adds that much complexity and enables more flexibility to optimize bundle size. Not a blocker for now as it can be re-added at any time, but wanted to leave my opinion. Also it would be consistent with other packages which also have it.

oshmyheliuk:
@obarcelonap @iivvaannxx I extracted the new package creation to a separate pull request: https://github.com/adobe/aio-commerce-sdk/pull/489.
I addressed all related comments there.
As the ACL may be changed and delayed, the new package part can already be merged, and it would simplify this PR after the merge.

.changeset/admin-ui-permission-client.md
iivvaannxx, line 1: This doesn't need to target `@adobe/aio-commerce-sdk`, we don't do updates in that package for anything that changes in their subpaths, because it's "implicit" that they are always 1:1. Because you're updating `admin-ui,` the metapackage will also release a `patch` and get updated automatically.

oshmyheliuk:
Recreated the PR -> https://github.com/adobe/aio-commerce-sdk/pull/522

PR #427, opened 2026-05-06
CEXT-6138: SDK Capability Discovery

iivvaannxx:


packages/aio-commerce-lib-app/generated/openapi.gen.json
obarcelonap, line 1: is this the final content? I am missing all the endpoints
additionally, I guess we will need to filter the content in `pre-app-build` hook

obarcelonap (commented):
Don't forget to flag the spec CEXT-6138 as implemented.
Also, we need to document from user perspective that filtered openapi spec is available as new endpoint, usage.md in aio-commerce-lib-app?

packages/aio-commerce-lib-app/generated/openapi.gen.json
iivvaannxx: Content filtering is pending yeah, thought it wasn't necessary but it is, will add. The final content I'll push just before merge, it only adds noise to diff count of PR and I didn't want to scare people with a 6K lines PR

packages-private/common-utils/source/actions/http/openapi.ts
jcuerdo, line 1: Missing test for this file.  Check others

jcuerdo (commented):
Changeset is missing too.

packages/aio-commerce-lib-app/generated/openapi.gen.json
iivvaannxx: I ended up doing so in the `app-config` at runtime to avoid tampering and because I think it's a bit simpler than having to script the generation of an extra file, let me know what you think.

iivvaannxx:
Closing as superseded by #481

PR #428, opened 2026-05-06
CEXT-6093: add foundational Commerce App Management agentic tooling

obarcelonap:
## Description

Adds foundational agentic tooling for Commerce App Management under `plugins/commerce-app-management/`, following the [agentskills.io](https://agentskills.io) open standard. The plugin is registered as a Claude Code marketplace plugin and compatible with any agent that supports the agentskills.io spec.

**Strategy:** skills follow single-responsibility — one skill per domain concern. `commerce-app-init` is the entry point (metadata only); domain skills chain off it. Each skill is independently installable. `adminUiSdk` domain excluded until its schema stabilises (CEXT-6092).

## Related Issue

CEXT-6093 — Establish foundational content for Commerce Developer AI Agent to support App Management workflows (epic CEXT-6092).

## Motivation and Context

The Commerce Developer AI Agent needs grounding content to guide developers through creating and extending Commerce apps with the aio-commerce-sdk. Skills are the portable, agent-agnostic format for this content — compatible with Claude Code, Cursor, VS Code Copilot, Gemini CLI, and other supported agents.

## How Has This Been Tested?

- Ran evals for `commerce-app-init` across 2 iterations; iteration-2 achieves 100% pass rate with skill vs 13% without (3 evals × with/without baseline)
- No SDK code changes — no automated tests required

N/A

## Types of changes

## Checklist:


README.md
```diff
@@ -62,6 +62,10 @@ The guide includes detailed information about:
 
 See also the [Contributing Guide](./.github/CONTRIBUTING.md) for more information on contributing guidelines.
 
+## Agentic Tooling
+
+Agentic tooling lives under [`plugins/`](./plugins/). Each plugin is independently installable and may include agent skills, MCP configurations, or other agentic content. See [`plugins/commerce-app-management/README.md`](./plugins/commerce-app-management/README.md) for the Commerce App Management plugin.
```
iivvaannxx, line 67: I don't have anything against `plugins` but given that eventually maybe we will host different things separately wouldn't something like `ai/` work better? For general AI related things? Or for this specific Claude plugin, `ai/plugins/`?

Example: We may separate things by category: `skills`, `agents`, `evals`... And then symlink those to their respective plugins (because different AI providers may have different plugin formats, but share this foundational content). So the source of truth is one, but plugins share them.

As a user of Codex, I would like to see a Codex plugin too

plugins/commerce-app-management/skills/commerce-app-init/evals/evals.json
```diff
@@ -0,0 +1,41 @@
...
+      "id": 1,
+      "prompt": "Create a new Adobe Commerce app called 'Loyalty Rewards' that manages customer loyalty points.",
+      "expected_output": "Project is initialized and app.commerce.config.ts is created with metadata: a meaningful id (e.g. loyalty-rewards), displayName 'Loyalty Rewards', a description about loyalty points, and version 1.0.0. Validation passes with no errors. No domain configuration (events, webhooks, business config) is added.",
+      "assertions": [
+        "app.commerce.config.ts is created at the project root",
+        "metadata.id contains only alphanumeric characters and hyphens",
+        "metadata.displayName is 'Loyalty Rewards' or equivalent (max 50 chars)",
+        "metadata.version is a numeric semver string (e.g. 1.0.0)",
+        "No domain keys (eventing, webhooks, businessConfig) appear in the config",
+        "npx @adobe/aio-commerce-lib-app generate manifest runs without errors"
```
iivvaannxx: For me a good sign of a correct project scaffolding is `aio app build` working. I've seen many times our scripts working, but `aio app build` failing just after, also `generate manifest` runs as part of the `pre-build` so it would already include this assertion.

plugins/commerce-app-management/skills/commerce-app-init/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+# Initialize a new Commerce App
+
+Scaffolds a bare Adobe Commerce app: initializes the project and creates a valid
+`app.commerce.config.ts` with metadata only. Extensibility domains (events, webhooks,
+business config) are added separately via domain skills.
+
+## Step 1 — Initialize the project
+
+```sh
+npx @adobe/aio-commerce-lib-app init --appName my-commerce-app --configFormat ts
```
iivvaannxx: This is not supported right now, we don't have flags yet. TBH, I'd move to the `aio` CLI before even considering implementing them in our scripts, as it's something we will not be able to port later because it will work differently.

plugins/commerce-app-management/skills/commerce-app-business-config/SKILL.md
```diff
@@ -0,0 +1,14 @@
+---
+name: commerce-app-business-config
+description: >
+  Manage custom business configuration in an Adobe Commerce app. Use when the
+  user wants to add, modify, or remove merchant-configurable settings exposed
+  through Commerce Admin. Requires a base app initialized with commerce-app-init.
+license: Apache-2.0
+compatibility: Requires Node.js 18+, aio CLI, and @adobe/aio-commerce-lib-app installed
```
iivvaannxx: Although probably compatible, I think we should not promote using unmaintained versions (this is about that Node 18+).

plugins/commerce-app-management/skills/commerce-app-init/evals/evals.json
iivvaannxx: Also, for context, the `@adobe` part in `npx @adobe/aio-commerce-lib-app` is not necessary if the dependency is installed (it's in `node_modules`). If it's installed, let's say at `v1.1.0` and someone runs that command, it will fetch the latest version from NPM (which might be > than `v.1.1.0`, have different logic and produce issues). If the dependency is installed locally, it should be `npx aio-commerce-lib-app`

Once we move to the `aio` CLI this will not be a problem anymore.

plugins/commerce-app-management/skills/commerce-app-init/evals/evals.json
```diff
@@ -0,0 +1,41 @@
...
+    {
+      "id": 2,
+      "prompt": "I have an existing App Builder project and I want to add Commerce app support to it. The app is for syncing inventory updates between Commerce and an external warehouse system.",
+      "expected_output": "The agent skips project initialization (existing project) and proceeds directly to creating app.commerce.config.ts with appropriate metadata. The id, displayName, and description reflect the inventory sync purpose. Validation passes.",
+      "assertions": [
+        "Agent runs the init command (safe on existing projects — it skips config creation if already valid)",
+        "app.commerce.config.ts is created with metadata only",
+        "metadata.description references inventory or warehouse sync",
+        "No domain configuration is added",
+        "npx @adobe/aio-commerce-lib-app generate manifest runs without errors"
```
iivvaannxx: Idem on the above assertion and `npx` comment.

plugins/commerce-app-management/skills/commerce-app-init/evals/evals.json
```diff
@@ -0,0 +1,41 @@
...
+      ]
+    },
+    {
+      "id": 3,
+      "prompt": "Create a Commerce app that listens to order placed events and sends a notification to Slack.",
+      "expected_output": "The agent scaffolds the base app with metadata only. It does not add event configuration. After completing the scaffold, the agent proactively suggests chaining to the commerce-app-eventing skill to add event sources.",
```
iivvaannxx: How is the part of "after completing the scaffold ..." evaluated? This would only be "assertable" by a user in an interactive session right?

plugins/commerce-app-management/skills/commerce-app-init/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+Safe to run on an existing App Builder project — if `app.commerce.config.ts` already exists and is valid, init skips creation and continues. After init:
+
+- `app.commerce.config.ts` is created at the project root (or validated if already present)
+- `@adobe/aio-commerce-sdk` and `@adobe/aio-commerce-lib-app` are installed
+- A `postinstall` hook is registered in `package.json`
```
iivvaannxx: There are several more things done, like the code generation, the creation of the `install.yaml` and some other things. I think everything is listed [in here](https://github.com/adobe/aio-commerce-sdk/blob/main/packages/aio-commerce-lib-app/docs/usage.md#setup), although not sure if that specific part is up-to-date.

plugins/commerce-app-management/skills/commerce-app-init/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+
+Use `--configFormat js` if the project is not using TypeScript.
```
iivvaannxx: Not necessary, you can have a TS config in a JS project, we made sure it will work anyways.

plugins/commerce-app-management/skills/commerce-app-init/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+
+> The `init` command also accepts a `--domains` flag, but do not use it here. Domain
+> scaffolding is handled by dedicated skills to keep each concern separate and independently installable.
```
iivvaannxx: Idem on flags not supported.

plugins/commerce-app-management/skills/commerce-app-init/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+Field constraints:
+
+| Field         | Constraints                                                                                                               |
+| ------------- | ------------------------------------------------------------------------------------------------------------------------- |
+| `id`          | Alphanumeric + hyphens only, max 100 chars. Synced to `package.json` `name` on init.                                      |
+| `displayName` | Free text, max 50 chars. Shown in Commerce Admin.                                                                         |
+| `description` | Free text, max 255 chars.                                                                                                 |
+| `version`     | `Major.Minor.Patch` integers only — e.g. `1.0.0`. No pre-release identifiers. Synced to `package.json` `version` on init. |
```
iivvaannxx: I'm a bit afraid this may get outdated. Maybe not this part, but if we do this for every piece of config it will most likely be. I would add a `references/` folder or similar where we either symlink the documentation (which already has this), or we provide links for an LLM to fetch and see the constraints (e.g. to the source code)

plugins/commerce-app-management/README.md
```diff
@@ -0,0 +1,85 @@
...
+Run evals for the commerce-app-init skill
+```
+
+To run a single eval case instead of the full suite:
+
+```
+Run eval 1 for the commerce-app-init skill
+```
+
+The agent will run each prompt with and without the skill, grade assertions, and write results to a `<skill-name>-workspace/` directory alongside the skill folder. Running the full suite spawns multiple subagents in parallel and typically takes 2–5 minutes.
```
iivvaannxx: Aren't evals a thing we should use ourselves? From what I read here it's like it's suggesting the user to do it

README.md
obarcelonap: The directory structure is intentionally based on adobe/skills, which uses a similar plugins/ layout. I feel the broader ecosystem isn't stable enough yet to make firm architectural decisions, restructuring with separate folders per category would be premature right now, and easier to do once there's a concrete reason for it.

About Codex, to support as plugin we need to add:
* `<root>/.agents/plugins/marketplace.json`
* `<root>/plugins/commerce-app-management/.codex-plugin/plugin.json`
But again, I'd stick to adobe/skills (only support claude for the moment) unless you have strong opinion.

plugins/commerce-app-management/skills/commerce-app-init/evals/evals.json
obarcelonap: Yep, better, changed.

obarcelonap: Replaced by `aio app build`

README.md
iivvaannxx: I honestly don't see why we would need to mimic that structure, different repos have different conventions or ways to do things, but I don't have a strong opinion on that.

Regarding the "_once there's a concrete reason for it_", for me the reason to do it is the example I gave you, having support for multiple agents without duplicating content.

> only support claude for the moment

I don't think this is a good move, you're deliberately choosing to leave out a big user-base that might not want or can't use Claude (either by economical reasons, company policies, preferences...). If there's something that Anthropic knows to do is to follow their own way without even bothering on trying to adopt the standard that the industry is trying to move into. I'm speaking of the `.agents` format, or the `AGENTS.md`. This is why we need the `CLAUDE.md` symlink btw.

Note that by omitting the `.agents` plugin/skills format we're essentially forbidding the users of the following AI providers of being able to use our agentic tooling:

- Amp
- Antigravity
- Cline
- Codex
- Cursor
- Deep Agents
- Dexto
- Firebender
- Gemini CLI
- GitHub Copilot
- Kimi Code CLI
- OpenCode
- Warp

plugins/commerce-app-management/skills/commerce-app-init/SKILL.md
obarcelonap: Reworded now just says "handles project setup" without enumerating internals.

plugins/commerce-app-management/skills/commerce-app-init/SKILL.md
obarcelonap: Removed all flags. The skill now creates config first (agent derives values from context and confirms), then runs `init` without arguments. Since the config already exists, init skips the interactive prompts entirely.
We can refactor the skill once the aio cli counterpart is available.

plugins/commerce-app-management/skills/commerce-app-init/SKILL.md
obarcelonap: removed

plugins/commerce-app-management/skills/commerce-app-init/SKILL.md
obarcelonap: removed

plugins/commerce-app-management/skills/commerce-app-init/SKILL.md
obarcelonap: Moved constraints into the asset as inline comments, one source of truth, no duplication risk in the skill.
For this skill the schema is simple enough that the asset covers it cleanly. Using references/ adds an extra file fetch step for the agent to learn four field constraints that fit in two lines of comments.
For domain skills (eventing, webhooks, business config) where the schema is much more complex, I'll introduce a references/ folder with a fetchable link.
I added an Authoring section with this convention
https://github.com/adobe/aio-commerce-sdk/blob/fb4a7e85be216219ca042e34e3d2d3e43d9a6baf/plugins/commerce-app-management/README.md?plain=1#L87-L97

README.md
obarcelonap: We had a slack chat about this, some clarifications:
* As skill, we're not leaving out any of the mentioned agents, all of them can install the skills with the help of `npx skill add`
* As a plugin we only support claude as done in adobe/skills, note plugin support is not strictly needed in this stage, is just convenient considering is the mainstream option.

Therefore, let's go with this, we can revisit in the future if needed.

plugins/commerce-app-management/skills/commerce-app-business-config/SKILL.md
obarcelonap: Bumped to Node 22 (Node 20 got EOL in Mar 24, 2026)

plugins/commerce-app-management/skills/commerce-app-init/evals/evals.json
obarcelonap: `expected_output` is narrative context for the grader, not a graded check itself,  only `assertions` are evaluated pass/fail. The chaining suggestion is covered by the assertion "Agent suggests using commerce-app-eventing skill for the event source setup", which the eval runner grades by running an LLM over the agent's output. Cleaned up `expected_output` to remove the duplication.

plugins/commerce-app-management/README.md
obarcelonap: created `DEVELOPMENT.md` file and move evals section there alongside with the authoring guidelines.

plugins/commerce-app-management/skills/commerce-app-init/assets/app.commerce.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from "@adobe/aio-commerce-lib-app/config";
+
+export default defineConfig({
+  metadata: {
+    id: "my-commerce-app", // alphanumeric + hyphens only, max 100 chars
+    displayName: "My Commerce App", // shown in Commerce Admin, max 50 chars
```
iivvaannxx: This is shown in App Management UI, not Commerce Admin. I think Commerce Admin is handled by Admin UI SDK.

iivvaannxx (approved):

PR #429, opened 2026-05-07
CEXT-6077: Add spec for dynamic options

iivvaannxx:
## Description

Add a new feature spec for the dynamic options in a list field.


specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+Factory functions cannot be JSON-serialized and would be silently stripped. This pipeline must
+change when the schema contains dynamic options.
+
+**Detection**: during `pre-app-build`, after loading the developer's config via `jiti`, check
+whether any `list` field has a function for `options`. A helper like `isDynamicSchema()` in
+`lib-config` can provide this check.
+
+**If no factories are present**: existing behavior unchanged. The schema is serialized to
```
obarcelonap, line 231: If no factories wouldn't be better to just operate with js files too so we don't have 2 ways of doing the same thing

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+undefined and should be decided during implementation.
+
+**Empty options.**  
+Should the SDK validate that a resolved options array is non-empty? A factory that returns `[]`
+would produce a list field with no choices, which is probably a developer error but could also be a
+valid transient state.
+
+## Future possibilities
+
+**Dynamic field visibility.** The same pattern (factory function that receives `params`) could
```
obarcelonap, line 348: Another potential feature is conditional values: the options in one select would depend on what's chosen in another. 
For instance, a first select with food types (Fruit, Vegetable, Meat) drives a second select that lists items of that type: pick "Fruit" and you get Banana, Apple, Strawberry.

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
+# Dynamic list options
```
obarcelonap, line 1: Not sure if dynamic options is too broad for this feature because it does not include other use cases like the ones listed in future section.
I'd say more like options resolved in runtime (or similar).

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+  accepted.
+
+## Drawbacks
+
+**Input duplication.** A factory that calls an external API needs credentials in `params`. Those
+credentials must appear in `inputs` under every action that resolves the factory. For apps with
+several dynamic fields calling the same API, this is boilerplate that worsens as more actions are
+involved.
+
+**Runtime-only errors.** A factory with a bug (wrong API call, unexpected response shape) fails at
```
obarcelonap, line 281: What about wrapping this functions in some kind of error handler so we can establish an api in case the function fails. This way the UI can react accordingly (show error instead of empty list of options)

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+runtime, not when the developer defines the schema. Static options are validated before the project is deployed;
+dynamic options are not.
+
+**Per-request latency.** Every request to an action that uses a dynamic field incurs the factory's
```
obarcelonap, line 285: This can be mitigated in the future by adding a caching layer so devs can set eviction times according to the frequency of change of the values.
Probably worth to add as future improvement.

obarcelonap (commented):
LGTM overall added some comments

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: It's to preserve backwards compatibility with projects that are already using those files. For example to call `initialize` on the usage of `lib-config`. 

If it weren't for backwards compatibility we wouldn't need a JS file at all, we could import the root config directly (after implementing [CEXT-6081](https://jira.corp.adobe.com/browse/CEXT-6081), which is flagged because of the same reason).

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: That was the original implementation idea ([the video](https://magento.slack.com/archives/C0A3SFDC6HY/p1777970766472809) I shared already was doing this). But this opens up to many possibilities that may be confusing with the current setup (e.g dynamic types or defaults). So we want to start smaller and evaluat in the future how we can handle the rest.

And I mean in the current setup because if you have a dynamic default, for example, with the current way of working in App Management every time you refresh you may end up with a different default (if it's generated with some kind of randomness, like the case of a UUID).

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: Mmm the name is because the `list` fields have an `options` property. So it's those being dynamic. I think what you're talking about would be better labeled as "dynamic fields", as per the business config being made out of an array of fields.

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: That might be doable via Valibot transformations. Will evaluate 👍🏻

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: Sounds good indeed, as a future improvement

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+  packages:
+    app-management:
+      actions:
+        config:
+          # ... existing config ...
+          inputs:
+            SOME_API_KEY: $SOME_API_KEY
+```
+
+The values are sourced from the developer's `.env` file (local), following the standard App Builder pattern. Nothing changes about how secrets are managed, only that the same secret may now need to appear in more than one extension point's config.
```
iivvaannxx: Quoting @danperezAdobe for visibility:

> this is only needed when the retriever of the option require an specific param? Ej. to connect to a 3rd party system

Yes, it's only in the case users need some kind of `input` to retrieve the options. And that value needs to be secret (in the `.env`).

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+const schema: BusinessConfigSchema = [
+  {
+    name: "paymentMethod",
+    label: "Default Payment Method",
+    type: "list",
+    selectionMode: "single",
+    options: async (params) => {
+      const methods = await fetchExternalApi(params.SOME_API_KEY);
+      return methods.map((m) => ({ label: m.title, value: m.code }));
+    },
```
iivvaannxx, line 97: Quoting @danperezAdobe for visibility:

> Are we doing during the validation step of the pre-build that options is resolvable? an no error arise?

I think that is kind of not our responsibility. We may be able to execute the function at validation time and see if it succeeds and returns valid data, but if it requires some kind of input we don't know about it may fail. 

It could be doable by trying to load all the `inputs` they specify in any action that resolves this,  and pass them in, but I don't think it's worth the effort.

iivvaannxx:
@obarcelonap Would be great to have some comment on the unresolved questions

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+import { fetchExternalApi } from "../lib/external.js";
+
+const schema: BusinessConfigSchema = [
+  {
+    name: "paymentMethod",
+    label: "Default Payment Method",
+    type: "list",
+    selectionMode: "single",
+    options: async (params) => {
+      const methods = await fetchExternalApi(params.SOME_API_KEY);
```
oshmyheliuk: Each time the App Management UI loads the configuration page, it makes a call to the external API, if there are several dynamic options, there will be multiple calls.
We should consider adding a cache mechanism for dynamic fields; the TTL can be specified by the user, and usually, values do not change.

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+```ts
+type ListOptionsFactory = (
+  params: RuntimeActionParams,
+) => ListOption[] | Promise<ListOption[]>;
+type ListOptionsValue = ListOption[] | ListOptionsFactory;
+```
+
+The Valibot field schema in
+`packages/aio-commerce-lib-config/source/modules/schema/fields.ts` is updated to accept
+`v.union([v.array(ListOptionSchema), <appropriate_factory_schema_here>])` for the `options` property.
```
oshmyheliuk: What does it mean`<appropriate_factory_schema_here>`? This will be updated.

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+**Centralising resolution.**  
+Each action that needs a resolved schema currently calls the factory independently, which means
+secrets must be in the `inputs` of each action and each factory call performs a separate network
+round-trip. An alternative would be to have the `config` and other actions invoke the `app-config`
+action to obtain a pre-resolved schema. This reduces duplication but introduces action-to-action
+coupling. This tradeoff should be evaluated before implementation.
+
+**Factory error handling.**  
+A factory that throws propagates as a runtime error in the action. The expected behaviour is
+undefined and should be decided during implementation.
```
oshmyheliuk: I think this should be resolved in the spec, not deferred to implementation
 - Will the action return an error, 500
 - Is the error logged?
 - Would it be visible in the UI to understand why the configuration is not loading?

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+
+Before this feature, a payment method selector requires hard-coded options:
+
+```ts
+const schema: BusinessConfigSchema = [
+  {
+    name: "paymentMethod",
+    label: "Default Payment Method",
+    type: "list",
+    selectionMode: "single",
```
oshmyheliuk, line 58: Maybe I missed, but would we support multiselect? I don't see any other selection modes in spec. If it's always single, this option can be removed.

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: I've played with function schemas and they get kind of verbose, so I reduced that to a placeholder. Take it just as a "slot" of the Valibot schema for the options function getter.

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: Yeah the AI was suggesting that all the time, I changed some occurrences to actually be resolved before implementation, but not all of them. You're right. 

Personally I think that we should return anyways the other parts of the schema, and maybe return an additional "errors" array that contains a list of issues found during schema evaluation. Similar to installation validation. This way we don't block other field editing.

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: The `selectionMode` is already supported. We support both a list of type `single` in which case the `default` is mandatory. And a list of type `multiple`, which defaults automatically to `[]` (no options selected).

Was that your question? Or were you asking if we would support dynamic option resolution in the context of a `multiselect`?

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: mmm, and where would that be specified? Maybe in the returned value? alongside the actual values?

```ts
// Just a sample
type DynamicResult {
  values: Array<{ label: string; value: string }>;
  options: { ttl: number }
}
```

Something like that?

specs/features/CEXT-6077-dynamic-list-options.md
obarcelonap: Fair enough.
Out of the scope of this discussion. I think getting the config should be a function so we don't fall in same error in the future. For the end user is just calling a fn (cheap) but it might save us of a future backward incompatible change.

specs/features/CEXT-6077-dynamic-list-options.md
obarcelonap: works for me

obarcelonap:
> @obarcelonap Would be great to have some comment on the unresolved questions

done

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+The previous implementation supported a factory that returned the entire schema. Rejected for this
+spec: it is broader than the stated need and makes the schema shape harder to reason about
+statically. It can be revisited independently.
+
+**What is the impact of not doing this?**  
+Apps that need merchant-specific or any other dynamic options cannot express them through the schema and must work around the limitation outside the SDK.
+
+## Unresolved questions
+
+**`default` for dynamic single-select lists.**
```
obarcelonap: Since the options are variable, probably the default might change too.
What if we reflect that in the options factory return type, something like
```typescript
type OptionsFactoryResponse = {
   options: ListOption[]
   default: ListOption
}
```

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+
+- Make `default` optional on fields with a factory for `options`.
+- Make `default` an index of the resolved array options (e.g: option 0)
+- Keep `default` required but skip cross-validation with the resolved options array.
+- Validate `default` at resolution time and surface an error if it is not present in the resolved
+  options.
+
+This must be resolved before implementation begins, as it affects the public schema type.
+
+**Centralising resolution.**
```
obarcelonap, line 329: Runtime actions access path is to get the resolved config for a given scope. Don't see the need to know the available options, just the resolved value(s). What I am missing?

specs/features/CEXT-6077-dynamic-list-options.md
```diff
@@ -0,0 +1,311 @@
...
+secrets must be in the `inputs` of each action and each factory call performs a separate network
+round-trip. An alternative would be to have the `config` and other actions invoke the `app-config`
+action to obtain a pre-resolved schema. This reduces duplication but introduces action-to-action
+coupling. This tradeoff should be evaluated before implementation.
+
+**Factory error handling.**  
+A factory that throws propagates as a runtime error in the action. The expected behaviour is
+undefined and should be decided during implementation.
+
+**Empty options.**
```
obarcelonap, line 336: I don't think this is an issue if we inform if there's an error (prev point).
If the request returns empty and is not expected, we will validate when trying to save the config and fail at that point.

specs/features/CEXT-6077-dynamic-list-options.md
oshmyheliuk: I was thinking about specifying it in the business config schema. But probably options with TTl in response would work as well
```
    name: "paymentMethod",
    label: "Default Payment Method",
    type: "list",
    selectionMode: "single",
    cacheTtl: 300,
    options: async (params) => {
```

specs/features/CEXT-6077-dynamic-list-options.md
oshmyheliuk: I think that for the final spec, such items should be resolved.
It's like an unresolved question.

specs/features/CEXT-6077-dynamic-list-options.md
oshmyheliuk: Ok, but can you update a spec to replace or remove the next phrase?
```
undefined and should be decided during implementation.
```

specs/features/CEXT-6077-dynamic-list-options.md
oshmyheliuk: If we already have selectionMode `multiple`, I believe it would work for the dynamic option as well.

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: Let's consider that as a future enhancement. I think returning it as part of the response makes more sense.

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: Done in [`68497fd`](https://github.com/adobe/aio-commerce-sdk/pull/429/commits/68497fdb7f9863be7cea39bee40ceb4998beb38d)

specs/features/CEXT-6077-dynamic-list-options.md
iivvaannxx: Done in [`68497fd`](https://github.com/adobe/aio-commerce-sdk/pull/429/commits/68497fdb7f9863be7cea39bee40ceb4998beb38d)

In the end, I ended up letting the error to bubble up. In the context of App Management this means the UI will render an error saying that we couldn't resolve options for a specific field. In the context of `lib-config` alone it's just an uncaught error that should be handled by the developers.

obarcelonap (approved):

PR #430, opened 2026-05-07
CEXT-6093: syncImsCredentials crashes when .env file does not exist

obarcelonap:
## Description

`syncImsCredentials` called `readFileSync` on the `.env` file before checking whether it exists, crashing with ENOENT on fresh projects that haven't run `aio app use` yet. The function now returns a discriminated union so callers can surface an actionable message per failure reason.

## Related Issue

N/A

## Motivation and Context

The `commerce-app-init` skill being added in https://github.com/adobe/aio-commerce-sdk/pull/428 runs `syncImsCredentials` as part of the post-init setup. On a brand-new project there is no `.env` file yet, so the skill was crashing instead of warning gracefully. This fix makes the skill resilient to that state.

## How Has This Been Tested?

Added a unit test covering the missing `.env` case. All existing tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/packages/aio-commerce-lib-auth/source/commands/sync-ims-credentials.ts
+++ b/packages/aio-commerce-lib-auth/source/commands/sync-ims-credentials.ts
@@ -7,7 +7,28 @@ export async function run() {
   consola.start("Syncing IMS credentials...");
 
   try {
-    await syncImsCredentials();
+    const result = await syncImsCredentials();
+
+    if (!result.ok) {
+      switch (result.reason) {
+        case "missing-env":
+          consola.warn(
+            ".env not found — run `aio app use` to configure your workspace.",
+          );
+          break;
+        case "no-ims-context":
+          consola.warn(
+            "No IMS context configured — run `aio login` to authenticate.",
+          );
+          break;
+        default: {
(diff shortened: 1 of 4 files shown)


packages/aio-commerce-lib-auth/source/commands/sync-ims-credentials.ts
```diff
@@ -7,7 +7,28 @@ export async function run() {
...
+        case "no-ims-context":
+          consola.warn(
+            "No IMS context configured — run `aio login` to authenticate.",
+          );
+          break;
+        default: {
+          // exhaustiveness check — errors if a new reason is added without handling it
+          const _: never = result.reason;
+        }
+      }
```
iivvaannxx, line 28: I feel this would be actually cleaner with a simple `if`. The `default` would be just handled by not having an `else`.

iivvaannxx (approved):

packages/aio-commerce-lib-auth/source/commands/sync-ims-credentials.ts
obarcelonap: Personally switch is more expressive on the case you're handling. With if/else I miss knowing what's the "else" key. I'll keep it.

PR #438, opened 2026-05-11
CEXT-6077: Dynamic List Options

iivvaannxx:
## Description

Introduce a new `dynamicList` type that can resolve `options` and `default` at runtime


packages/aio-commerce-lib-config/source/modules/schema/utils.ts
```diff
@@ -44,10 +47,64 @@ export function calculateSchemaVersion(schema: BusinessConfigSchema): string {
...
+/**
+ * Resolves dynamic list options in a business configuration schema.
+ *
+ * @param schema - Schema that may contain list option factories.
+ * @param params - Runtime action params to pass to each option factory.
+ * @returns A new schema with all list options resolved to arrays.
+ *
+ * @throws {CommerceSdkValidationError} If any resolved options are invalid.
+ */
+export async function resolveDynamicBusinessConfigSchema(
```
obarcelonap: I'd expect that when receiving a schema and combining with its params the return type to be different, not again a schema.
Additionally I'd remove the `Dynamic` from this fn name, just `resolveBusinessConfigSchema`

packages/aio-commerce-lib-config/source/config-manager.ts
```diff
@@ -50,23 +56,64 @@ export type InitializeOptions = {
...
   }
 
   if (options.schema) {
-    setGlobalSchema(options.schema);
-    return;
+    if ("params" in options) {
+      return initializeWithRuntimeParams(options);
+    }
+
+    if (hasDynamicListOptions(options.schema)) {
```
obarcelonap: In the future not only the dynamic list options (maybe).
What about using `hasDynamicConfigSchema` function here?

obarcelonap (approved):
LGTM

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
```diff
@@ -62,11 +70,206 @@ const __dirname = dirname(__filename);
...
+  const configFilePath = await resolveCommerceAppConfig(projectRoot);
+
+  if (!configFilePath) {
+    throw new Error(
+      "Dynamic list options require an app.commerce.config.* file so generated actions can import the option factories at runtime.",
+    );
+  }
+
+  if (TYPESCRIPT_CONFIG_EXTENSIONS.has(extname(configFilePath))) {
+    consola.info("Generating JS config module...");
```
oshmyheliuk: The same message on line 206 in the else section can be moved out of the condition.

packages-private/scripting-utils/source/yaml/codegen.ts
```diff
@@ -71,16 +71,24 @@ function buildWeb(extConfig: Document, web: string) {
  * @param path - The path where the action is located
  * @param action - The action definition
  */
-function buildActionDefinition(action: ActionDefinition) {
+function buildActionDefinition(
```
oshmyheliuk, line 82: Description of the method should be updated

packages/aio-commerce-lib-app/source/actions/config.ts
```diff
@@ -72,6 +72,18 @@ function filterPasswordFields<T extends Omit<ConfigValue, "origin">>(
   });
 }
 
+async function resolveConfigSchema(params: ConfigActionParams) {
```
oshmyheliuk: Docs can be added

packages/aio-commerce-lib-app/source/actions/config.ts
```diff
@@ -72,6 +72,18 @@ function filterPasswordFields<T extends Omit<ConfigValue, "origin">>(
...
+  const { configSchema } = await initialize({
+    schema: params.configSchema,
+    params,
+  });
+
+  return validateCommerceAppConfigDomain(
```
oshmyheliuk: This call may be redundant:

```
resolveConfigSchema() calls initialize({ schema, params })
that goes through initializeWithRuntimeParams() in packages/aio-commerce-lib-config/source/config-manager.ts
initializeWithRuntimeParams() calls resolveDynamicBusinessConfigSchema()
resolveDynamicBusinessConfigSchema() ends with return validateBusinessConfigSchema(resolvedFields)
So by the time configSchema comes back to config.ts, it is already validated as a BusinessConfigSchema. The extra validateCommerceAppConfigDomain(..., "businessConfig.schema") is redundant.
```

packages-private/scripting-utils/source/yaml/codegen.ts
iivvaannxx: Done in [`116ba76`](https://github.com/adobe/aio-commerce-sdk/pull/438/commits/116ba768d0d66e4264b139836c37301bf310c465)

packages/aio-commerce-lib-app/source/actions/config.ts
iivvaannxx: Done in [`4ae144a`](https://github.com/adobe/aio-commerce-sdk/pull/438/commits/4ae144aba22c43472410db9fb1562777a1687866)

packages/aio-commerce-lib-app/source/actions/config.ts
iivvaannxx: As per your below comment, the overall `resolveConfigSchema` was kind of useless so I ended up removing that helper. See the response commit in the other answer.

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
iivvaannxx: Done in [`0aff250`](https://github.com/adobe/aio-commerce-sdk/pull/438/commits/0aff25032cea72174876ec1664a5418f0ab89ead)

packages/aio-commerce-lib-config/source/modules/schema/utils.ts
iivvaannxx: Applied some overall changes to the public interface of the library that was affected by these changes. I think now it's better and clearer. See [`404d9e5`](https://github.com/adobe/aio-commerce-sdk/pull/438/commits/404d9e52ed7e6641ce171497b5ce83eb363b23f0)

Also you were right, but I think it's kind of the reverse, it's not the return type that needs to change, but the parameter itself. Updated the types to  distinguish between a schema that is fully static and one that is maybe dynamic. This function only supports the latter, and always returns a static one.

packages/aio-commerce-lib-config/source/config-manager.ts
iivvaannxx: Done in [`3ebd443`](https://github.com/adobe/aio-commerce-sdk/pull/438/commits/3ebd44344b23c21e96d918f081ad2acaffb045ca)

jcuerdo (approved):

jcuerdo:
Could you clarify what the agreement was about the specs? I see the spec is still open for this ticket https://github.com/adobe/aio-commerce-sdk/pull/429/changes . Should the spec be merged before starting with the implementation/pull request?

iivvaannxx:
> Should the spec be merged before starting with the implementation/pull request?

@jcuerdo Probably yes, I think it's the expectation. I didn't merge it because it still had some conversations open (not critical to implementation), that were solved in the middle of the process. But next time I'll make sure to merge before 👍🏻

packages/aio-commerce-lib-app/docs/usage.md
obarcelonap, line 1: while I understand the business config usage is in its own package, maybe makes sense here to add an example of a full config and refer to usage in the module to ease discoverability within agents

iivvaannxx: discussed offline, this is already like this

PR #440, opened 2026-05-12
CEXT-6153: Installations failing and succeeding upon retry

jcuerdo:
## Description

  - runInstallation in runner.ts now retries the full workflow once on failure. If the retry succeeds, the result carries isRetry: true; if both attempts fail, onInstallationFailure fires only on the second
  attempt.
  - SucceededInstallationState gains an optional isRetry?: boolean field.
  - The onInstallationSuccess log message distinguishes retry successes ("Installation succeeded on retry") from first-attempt successes.
  
  
image

image

## Motivation and Context

  Installations fail intermittently against the Commerce backend for reasons outside our control. Rather than surfacing the error immediately and asking partners to retry manually, we absorb the first failure
  transparently. onInstallationFailure is suppressed on attempt 1 so metrics are not distorted by recoverable failures.

## How Has This Been Tested?

  - Unit tests added in runner.test.ts under "runInstallation — retry behavior":
    - Retry returns { status: "succeeded", isRetry: true } when attempt 1 fails and attempt 2 succeeds.
    - onInstallationFailure is not called when the retry succeeds.
    - onInstallationFailure is forwarded to the retry attempt (and called) when both attempts fail.
    - The warn log includes the step path, error key, and message.
  - Run pnpm --filter @adobe/aio-commerce-lib-app test to execute the full suite.

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/source/management/installation/runner.ts
```diff
@@ -69,19 +69,45 @@ export function createInitialInstallationState(
...
+  );
+
+  const retryState = createInitialState({ rootStep, config });
+  const retryResult = await executeWorkflow({
+    rootStep,
+    installationContext,
+    config,
+    initialState: retryState,
     hooks,
   });
```
iivvaannxx: AFAIK the expectations of the UI is that there's no UI change in the status. This means that if a step is completed, it should only retry starting from where it failed. Is `createInitialState` providing this functionality?

packages/aio-commerce-lib-app/source/management/installation/runner.ts
```diff
@@ -69,19 +69,45 @@ export function createInitialInstallationState(
...
+
+  if (retryResult.status === "succeeded") {
+    return { ...retryResult, isRetry: true };
```
iivvaannxx: I would add this into some kind of `metadata` property instead of mixing with the rest. And do it whether it's successful or not.

obarcelonap:
Do we still need this one after merging https://github.com/adobe/aio-commerce-sdk/pull/444?
TBH I feel this is a workaround

jcuerdo:
@asalloum5 what do you think?

asalloum5 (approved):
Looks good to me, would like to see a demo if you have any recorded one. Thanks,

asalloum5:
@jcuerdo @obarcelonap Yes we still need this since we have other errors that are random that are out of our control, like for example the 409 conflicts error when invoking /registration on I/O events API. Additionally, this retry is helpful in case of a network error on first call for example.

obarcelonap (commented):
It works but I'd like the amount of retries to be configurable. I think a good option would be to add an option in `RunInstallationOptions` that allows to configure the amount of retries. We could call it `maxRetries` for instance and the default value could be 1.
Additionally, in the returned data from the installation we could always return a `retryCount` (more flexible than a boolean).
What do you think?

jcuerdo:
@obarcelonap  I was thinking about it at the beginning, but as far as the spec is only one retry, I decided to do it as much simple as possible. Anyway, if you consider it's important, I can do it.

asalloum5 (approved):

asalloum5:
@obarcelonap do you mean to have the code ready for potential extensibility to allow retry configuration in the future? For now the ask from Dani is to have only 1 "hidden" retry. I guess if we want to have this functionality, better have a new ticket to discuss in refinement since this PR is aligned with the AC already defined.

obarcelonap (approved):
@jcuerdo @asalloum5 I was not aware that it was a requirement to not include retries in the public API, approved.
TBH I don't understand the decision, but I can live with that 🤷

PR #460, opened 2026-05-20
CEXT-6160: spec for store and expose Commerce system config during app associa…

vinayrao2000:
## Description

Adds a new standalone `association` runtime action and two typed async helpers to
`aio-commerce-lib-app`, plus a new `association` module in `aio-commerce-lib-config`
for storing Commerce instance details in Adobe I/O State.

**New runtime action** (`aio-commerce-lib-app`) — always deployed alongside `app-config`:

- `POST /` — stores `{ baseUrl, env }` when the app is associated with a Commerce instance
- `DELETE /` — clears the stored data on unassociation
- Protected with `require-adobe-auth: true`

**New helpers** exported from `@adobe/aio-commerce-lib-app`:

- `getAssociatedCommerceInstance(params)` — returns the stored `{ baseUrl, env }` or `null`
- `getAssociatedCommerceClient(params)` — returns a ready-to-use `AdobeCommerceHttpClient`
  built from the stored data, or `null`

**Storage** (`aio-commerce-lib-config`) — new association module under key `"association"`
with max TTL (1 year), refreshed on every read.

## Related Issue

CEXT-6160

## Motivation and Context

Every Commerce app that needs to call the Commerce API was reimplementing the same logic to
store and retrieve the Commerce Base URL, typically during a custom installation step. This
caused two problems: duplication across apps, and a wrong lifecycle (storing at install time
instead of keeping it in sync with the association state).

The SDK already receives `commerceBaseUrl` and `commerceEnv` at every association/unassociation
event. This change introduces a standard persistence mechanism (Adobe I/O State via
`aio-commerce-lib-config`) tied to the association lifecycle, and typed helpers that any
runtime action can consume, regardless of which OpenWhisk package the action belongs to.

## Types of changes

## Checklist

- [x] I have signed the Adobe Open Source CLA.
- [] My code follows the code style of this project.
- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.
- [] I have added tests to cover my changes.
- [] All new and existing tests passed.


packages/aio-commerce-lib-app/package.json
```diff
@@ -63,13 +63,24 @@
...
+      "./runtime": {
+        "import": {
+          "types": "./dist/es/runtime.d.mts",
+          "default": "./dist/es/runtime.mjs"
+        },
+        "require": {
+          "types": "./dist/cjs/runtime.d.cts",
+          "default": "./dist/cjs/runtime.cjs"
+        }
+      },
```
iivvaannxx: I would not export this from `/runtime`, just from the root entrypoint, so instead of:

```ts
import { ... } from "@adobe/aio-commerce-lib-app/runtime"
```

it would be:


```ts
import { ... } from "@adobe/aio-commerce-lib-app"
```

packages/aio-commerce-lib-app/source/actions/installation.ts
```diff
@@ -111,6 +118,43 @@ function createUninstallationStore() {
...
+  const updateKeys = Object.keys(updates);
+  const kept = existing.filter((p) => !updateKeys.includes(p.key));
+  const added = updateKeys
+    .filter((key) => updates[key] !== undefined)
+    .map((key) => ({ key, value: updates[key] as string }));
+  await ow.packages.update({
+    name: packageName,
+    package: { parameters: [...kept, ...added] },
+  });
+}
```
oshmyheliuk: Probably such an approach would work. But I thought the idea was to store this configuration by using logic from `aio-commerce-lib-config`. So it can be retrieved in runtime actions from the configuration.

specs/features/CEXT-6160-commerce-system-config-on-association.md
```diff
@@ -0,0 +1,254 @@
...
+
+- Store the Commerce Base URL and deployment type (PaaS/SaaS) as OpenWhisk package parameters
+  when an app is associated with a Commerce instance.
+- Clear the stored parameters when the app is unassociated.
+- Expose a typed helper that runtime actions can use to read the values from their params.
+
+**Non-goals:**
+
+- Storing any Commerce system configuration beyond Base URL and deployment type.
+- Replacing the `AIO_COMMERCE_API_BASE_URL` and `AIO_COMMERCE_API_FLAVOR` package params set
```
obarcelonap: Wouldn't be better to use these params names? Otherwise the params resolver won't work and might be a bit confusing when to use what. 
For instance, how the dev will decide when to use `AIO_COMMERCE_API_BASE_URL` vs `AIO_COMMERCE_BASE_URL`

specs/features/CEXT-6160-commerce-system-config-on-association.md
```diff
@@ -0,0 +1,254 @@
...
+}
+```
+
+Or use the typed helper from the `./runtime` export:
+
+```ts
+import { getCommerceSystemConfig } from "@adobe/aio-commerce-lib-app/runtime";
+
+export async function main(params) {
+  const config = getCommerceSystemConfig(params);
```
obarcelonap: While this kind of helper function is needed, since the main purpose of these values is to be able to do http requests to the commerce instance, I think we need also to use it in some way in http client param resolution 
https://github.com/adobe/aio-commerce-sdk/blob/9a7f23eb953371c0aa3cdb918ab74b1074b84e58/packages/aio-commerce-lib-api/source/lib/commerce/helpers.ts#L236-L244

specs/features/CEXT-6160-commerce-system-config-on-association.md
```diff
@@ -0,0 +1,254 @@
...
+
+### Storage
+
+The Commerce system configuration is stored as OpenWhisk package parameters on the package that
+owns the installation action. Two reserved parameter names are used:
+
+- `AIO_COMMERCE_BASE_URL` — the Commerce API base URL
+- `AIO_COMMERCE_ENV` — the deployment type (`"saas"` or `"paas"`)
+
+OpenWhisk injects all package parameters into every action's `params` object at invocation time,
```
obarcelonap: I am afraid that this is coupling the implementation too much to OpenWhisk. Have you aligned this approach with the appbuilder team?

specs/features/CEXT-6160-commerce-system-config-on-association.md
```diff
@@ -0,0 +1,254 @@
...
+  params: RuntimeActionParams,
+): CommerceSystemConfig | null;
+```
+
+`params` is the standard `RuntimeActionParams` object every runtime action receives. The helper
+reads `AIO_COMMERCE_BASE_URL` and `AIO_COMMERCE_ENV` from `params` and returns a typed
+`CommerceSystemConfig`, or `null` if either value is absent. The function is synchronous — no
+async call is needed because the values are already present in `params`.
+
+### Changes to `commerce-app-management`
```
obarcelonap: I believe we should not have `commerce-app-management` implementation details here. Maybe you can just describe broadly how a client of app-management api is impacted and the expected flows.

specs/features/CEXT-6160-commerce-system-config-on-association.md
```diff
@@ -0,0 +1,254 @@
...
+
+A new export entry `@adobe/aio-commerce-lib-app/runtime` exposes a typed helper for use inside
+runtime actions:
+
+```ts
+/**
+ * Returns the Commerce system configuration populated during association,
+ * or null if the app is not currently associated with a Commerce instance.
+ */
+export function getCommerceSystemConfig(
```
obarcelonap: Not sure if the concept system config is too broad. I'd prefer something like `getAssossociationDetails` or `getAssociatedCommerceInstance` or similar

obarcelonap (commented):
Thanks for writing a spec for these changes, I only reviewed this part.
Normally you should first pr only with the spec so you can get feedback and then once approved and merged you can go ahead with the implementation.
Spec workflow is defined here
https://github.com/adobe/aio-commerce-sdk/blob/9a7f23eb953371c0aa3cdb918ab74b1074b84e58/specs/conventions.md?plain=1#L54-L66

specs/features/CEXT-6160-commerce-system-config-on-association.md
```diff
@@ -0,0 +1,254 @@
...
+1. **Duplication.** Every app that needs to call the Commerce API writes the same storage and
+   retrieval logic.
+2. **Wrong lifecycle.** Storing the config during installation does not keep it in sync with the
+   association state. If the app is associated with a different instance, or unassociated, the
+   stored config is not updated or cleared.
+
+The SDK is already passed `commerceBaseUrl` and `commerceEnv` at every association and
+unassociation event — the data apps need is already flowing through the system. The missing piece
+is persistence tied to the association lifecycle and a standard way for runtime actions to consume
+it.
```
iivvaannxx: In order to implement some of the other requested features (e.g. helpers to publish events into the Event Ingress API) we'll also need to store data about the project/workspace we're associating with.

specs/features/CEXT-6160-commerce-system-config-on-association.md
```diff
@@ -0,0 +1,254 @@
...
+
+**Goals:**
+
+- Store the Commerce Base URL and deployment type (PaaS/SaaS) as OpenWhisk package parameters
```
iivvaannxx: I am not very much aware of how package parameters work, but given that we're applying this "save" under the context of the `app-management` package, does this work for actions that are set under other packages? The `app-management` package is only ours, but developers have their own.

specs/features/CEXT-6160-commerce-system-config-on-association.md
```diff
@@ -0,0 +1,254 @@
...
+| `AIO_COMMERCE_BASE_URL` | Commerce API base URL of the associated instance |
+| `AIO_COMMERCE_ENV`      | Deployment type: `"saas"` or `"paas"`            |
+
+A runtime action can read these directly:
+
+```js
+export async function main(params) {
+  const baseUrl = params.AIO_COMMERCE_BASE_URL;
+  const env = params.AIO_COMMERCE_ENV;
+}
```
iivvaannxx: While having in `params` is somehow standard, it also forces you to move the `params` object everywhere you need this data. So if you are for example in a 4-level deep function call hierachy and you need this data only at that level, you will be forced to drill it down from the first layer down to the fourth.

Business configuration would allow retrieving this without coupling to `params`

specs/features/CEXT-6160-commerce-system-config-on-association.md
iivvaannxx: But then we would be coupling `lib-app` to `lib-api`? Unless I am not understanding what to mean not sure if this is a good idea.

specs/features/CEXT-6160-commerce-system-config-on-association.md
```diff
@@ -0,0 +1,254 @@
...
+`APP_STATUS_ASSOCIATED` state — it called `handleUninstallSuccess()` directly without going
+through `useUnassociateApp`, so `DELETE /installation/association` was never called for those
+apps. The fix exposes `clearCommerceSystemConfig` from `useUnassociateApp` and calls it in the
+shortcut branch before completing the unassociation.
+
+### Edge cases
+
+- **Association before this feature existed.** Apps associated before `POST /association` was
+  introduced have no stored parameters. `getCommerceSystemConfig` returns `null`. Apps must handle
+  this explicitly.
```
iivvaannxx: @asalloum5 This sounds to me an acceptable tradeoff. But was wondering if we could do something similar to the "Sync Commerce Scopes"? For those apps that do not have this data maybe a "Sync Association" or something? So that they do not need to: Unassociate + Associate

Might also be a "solution" (probably not the best) to the edge case that @oshmyheliuk mentioned about a Commerce Instance changing URLs.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,247 @@
...
+- **Reserved config key naming.** The exact key used to store the data in `aio-commerce-lib-config`
+  needs to be defined as a reserved SDK concern so that apps do not accidentally collide with it.
+- **Backfill for legacy apps.** A mechanism to backfill stored data for apps associated before
+  this feature was introduced is out of scope but should be tracked.
+
+## Future possibilities
+
+- The stored shape can be extended with `projectId` and `workspaceId` at association time,
+  providing data needed by other planned SDK features without requiring a new association step.
+- `getAssociatedCommerceInstance` could become the foundation for a higher-level helper that
```
obarcelonap: can you include this in the scope and change the `fetch` example to be using resolver function? otherwise code will be repeated to create the commerce api client from the association details which we can already encapsulate.

vinayrao2000: Yes good point, as suggested I will update the spec to include getAssociatedCommerceClient in scope and replace the fetch example with the resolver function pattern. This will return a ready to use AdobeCommerceHttpClient built from the stored association data, so developers don't have to wire it up themselves. I will implement it in this PR.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,247 @@
...
+
+The association data is stored using the infrastructure already established in
+**`aio-commerce-lib-config`** — specifically, the shared `getSharedState()` utility from
+`aio-commerce-lib-config/source/utils/repository.ts`, which provides a lazy-initialized
+Adobe I/O State client shared across the SDK.
+
+A new module is added to `aio-commerce-lib-config` specifically for association data,
+separate from the existing Business Configuration module. It uses the same `getSharedState()`
+client but stores under a dedicated reserved key (`association`) rather than the
+`configuration.{scopeCode}` keys used by Business Configuration.
```
oshmyheliuk: I think it should be mentioned that the cache TTL should be specified for this configuration, as by default, TTL is 24 hours in aio lib state https://github.com/adobe/aio-lib-state/blob/main/doc/api.md#adobestateputoptions--object. 
The max TTL of 1 year can be set.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,247 @@
...
+type AssociatedCommerceInstance = {
+  baseUrl: string;
+  env: "saas" | "paas";
+};
+```
+
+The shape is designed to extend future fields (e.g. `projectId`, `workspaceId`) can be
+added without a breaking change.
+
+### New `association` runtime action
```
oshmyheliuk, line 157: Please mention `require-adobe-auth: true` in specs for those new actions, as they must be protected the same way as others.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,247 @@
...
+directly keeps the import consistent with how the rest of the SDK is consumed.
+
+**What is the impact of not doing this?**
+Every app that needs the Commerce Base URL in runtime actions continues to implement its own
+storage and retrieval logic with no automatic cleanup on unassociation and no standardised API.
+
+## Unresolved questions
+
+- **Reserved config key naming.** The exact key used to store the data in `aio-commerce-lib-config`
+  needs to be defined as a reserved SDK concern so that apps do not accidentally collide with it.
```
oshmyheliuk: This contradicts line 93, as the key naming was already mentioned:
```
client but stores under a dedicated reserved key (`association`) rather than the
```

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Client integration
+
+Any client that manages the app association lifecycle is responsible for driving two calls
+against the `association` endpoint. The endpoint URL is discoverable from the app's extension
+points metadata, registered in `workerProcess` alongside the existing `app-config` and
+`installation` hrefs.
+
+**On association** — after the app is successfully registered with the Extension Manager,
+the client calls `POST /` with the Commerce instance details. This step is best effort a failure does not roll back the registration.
```
oshmyheliuk: If a ` POST/` request to the association endpoint fails, it may lead to a broken application if runtime actions depend on that configuration. I think we can't mark the app as successfully associated if the configuration wasn't saved. Probably the association should fail in such a case.

specs/features/CEXT-6160-association-data.md
vinayrao2000: I will update accordingly. The default TTL of 24 hours is not appropriate here association data is long-lived and must not expire on its own. It should only be cleared explicitly on unassociation. We'll set it to the max TTL of 1 year, after 1 year without re-association the data will expire.

specs/features/CEXT-6160-association-data.md
vinayrao2000: sure I will add auth to the action

specs/features/CEXT-6160-association-data.md
vinayrao2000: Good point. However, I have a suggestion, please let me know if this approach sounds okay. Since the association with the Extension Manager and the config storage are separate concerns, we feel that a transient network error on the association endpoint should not block the entire association flow. Instead of failing the association completely, we plan to surface the failure in the App Management UI so developers are aware of it and can re-associate the app if needed to resolve the issue.

specs/features/CEXT-6160-association-data.md
vinayrao2000: sorry for the confusion, I will remove here

specs/features/CEXT-6160-association-data.md
oshmyheliuk: Does the stored configuration expire in one year? Need to ensure that the configuration won't be lost.

specs/features/CEXT-6160-association-data.md
oshmyheliuk: I'm not sure what the best approach is here. I would try to associate the app by calling `POST /association` first, and only on success, register the app in the Extension Manager.

Also, we need to mention in specs that the `POST /association` endpoint may not exist (for older versions of the SDK), and if so, the association should happen only in the Extension Manager.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+found.
+
+This eliminates the repeated boilerplate of combining stored instance details with action
+params to construct the client — a pattern every action that calls Commerce would otherwise
+duplicate.
+
+### Client integration
+
+Any client that manages the app association lifecycle is responsible for driving two calls
+against the `association` endpoint. The endpoint URL is discoverable from the app's extension
```
oshmyheliuk: > is responsible for driving two calls against the `association` endpoint

Can you please clarify what it means? Why for two calls against the `association` endpoint?

vinayrao2000: The two calls happen at different points in the lifecycle: POST / on association stores the data, and DELETE / on unassociation clears it. I'll reword that line for clarity.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Yes, good suggestion, that order is cleaner and removes the rollback complexity. I’ll update the spec accordingly.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+package the action belongs to.
+
+**If either helper returns `null`**, the app is either not currently associated or was associated before this feature was introduced. Apps must handle this case explicitly.
+
+**Available fields** on the `AssociatedCommerceInstance` object:
+
+| Field     | Type               | Description                              |
+| --------- | ------------------ | ---------------------------------------- |
+| `baseUrl` | `string`           | Commerce API base URL                    |
+| `env`     | `"saas" \| "paas"` | Deployment type of the Commerce instance |
```
obarcelonap, line 82: since we already have the flavour concept in the env vars might make sense to use it here too

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+type AssociatedCommerceInstance = {
+  baseUrl: string;
+  env: "saas" | "paas";
+};
+```
+
+The shape is designed to extend future fields (e.g. `projectId`, `workspaceId`) can be
+added without a breaking change.
+
+### New `association` runtime action
```
obarcelonap, line 157: This is not in sync with the pr description.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+After this feature ships, any runtime action can call the Commerce API without custom storage
+setup or client construction boilerplate.
+
+**Primary pattern — get a ready-to-use client:**
+
+```ts
+import { getAssociatedCommerceClient } from "@adobe/aio-commerce-lib-app";
+
+export async function main(params) {
+  const client = await getAssociatedCommerceClient(params);
```
obarcelonap: I think we should try to avoid `async` on `getAssociatedCommerceClient` the idea is to defer fetching the url from storage until a request is fired.
Implementation idea, instead of setting `baseUrl` when creating the client use a prerequest hook to compose the final url. Once the commerce `baseUrl` is resolved once, the value should be cached (memoize can help) so if client is used multiple times it is not retrieved again.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+same application regardless of package boundaries.
+
+**Why a new module in `aio-commerce-lib-config` rather than using `setConfiguration` directly?**
+The existing `setConfiguration`/`getConfiguration` API in `aio-commerce-lib-config` is
+designed for Business Configuration — scope-tree based values keyed by Commerce scope codes.
+Association data is app-level metadata, not scope-specific. A dedicated module with its own
+reserved key (`association`) keeps the two concerns clearly separated while reusing the same
+underlying storage infrastructure.
+
+**Why a standalone `association` action?**
```
obarcelonap, line 299: I'd prefer the former approach, adding new paths under the `install` runtime action. 
We already discussed this many times and it seems we don't get an internal agreement, won't block the spec because of this

obarcelonap (approved):

specs/features/CEXT-6160-association-data.md
vinayrao2000: Currently, the TTL is only refreshed when the app is associated or re-associated. To keep the stored configuration alive, we plan to add refresh TTL logic inside getAssociationData as well. In practice, any action that needs the Commerce baseUrl/env triggers this call, so as long as the app receives traffic that touches the Commerce config, the TTL gets refreshed on every read and the data stays alive. Can we go ahead with this approach, or do you have a better suggestion?

specs/features/CEXT-6160-association-data.md
vinayrao2000: Earlier feedback asked to use env to match commerceEnv in the request body. Can we agree on which one to use? I'll update it once it's decided.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Thanks for pointing that out, the PR description describes the old approach. I'll update it to match the current spec.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Thanks for sharing your preference and not blocking.

specs/features/CEXT-6160-association-data.md
vinayrao2000: I added async for the following reasons:

- Upfront error check: if (!client) return works cleanly. Sync version would throw deep inside the first HTTP call.
- Consistent: getAssociatedCommerceInstance is already async; mixing styles is confusing.
- Weak performance argument: calling code is async anyway, so one await doesn't hurt. Memoization works in both styles. 

Please suggest shall I go with async or sync

oshmyheliuk (approved):

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+
+export async function main(params) {
+  const client = await getAssociatedCommerceClient(params);
+
+  if (!client) {
+    return {
+      statusCode: 400,
+      body: { error: "App is not associated with a Commerce instance." },
+    };
+  }
```
iivvaannxx: I think it's better to make `getAssociatedCommerceClient` throw if there's any issue when constructing it instead of returning a nullish value. Someone wanting to use that function should make sure that pre-requisites are met before invoking it (associate the app). Unless there's some other code path I am not seeing I don't see why we should return a null value here.

Otherwise i'm forced to always do that nullish check everytime I call the function, which I find bad DX for a public helper.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+
+export async function main(params) {
+  const instance = await getAssociatedCommerceInstance(params);
+
+  if (!instance) {
+    return {
+      statusCode: 400,
+      body: { error: "App is not associated with a Commerce instance." },
+    };
+  }
```
iivvaannxx: Idem here on throwing if data doesn't exist.

specs/features/CEXT-6160-association-data.md
iivvaannxx: I'd stick with `env` from now on in the future. At some point deprecating the `flavor` concept from `lib-api` which was confusing some people. The `app.commerce.config` uses `env` for business configuration properties that are conditional on it.

specs/features/CEXT-6160-association-data.md
iivvaannxx: > Does the stored configuration expire in one year? Need to ensure that the configuration won't be lost.

@oshmyheliuk If this is going to be stored in `lib-config` which is backed by `lib-files` as a persistent storage layer, it shouldn't get lost no?

@vinayrao2000 could you elaborate on this TTL refresh you have in mind? not sure if that would actually be needed here.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+  pass the data through every layer of the call stack.
+
+The new module exposes three internal functions used by `aio-commerce-lib-app`:
+
+```ts
+// aio-commerce-lib-config (new internal module)
+setAssociationData(data: AssociatedCommerceInstance): Promise<void>
+getAssociationData(): Promise<AssociatedCommerceInstance | null>
+clearAssociationData(): Promise<void>
+```
```
iivvaannxx, line 139: This is leaking concepts of App Management directly into `lib-config` which should be completely agnostic of it. This module should not know nothing about "association".

specs/features/CEXT-6160-association-data.md
iivvaannxx, line 82: TBH, I'd say we can store all what we have in `appData`, which is the info the UI has. We can start conservative, but I don't see a reason why.

See the data [here](https://github.com/adobe/aio-commerce-sdk/blob/9d6fe50302f05186c370c740715c996cae52f26a/packages/aio-commerce-lib-app/source/management/installation/schema.ts#L16-L29)

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+developer (e.g. rolling back the Extension Manager registration or otherwise preventing
+the app from being shown as successfully associated).
+
+**On unassociation** — after unassociation completes, the client calls `DELETE /` to remove
+the stored data. This step is best-effort — a failure does not block the unassociation.
+
+### Edge cases
+
+- **Apps associated before this feature.** No stored data exists; `getAssociatedCommerceInstance`
+  returns `null`. Apps must handle this explicitly.
```
iivvaannxx: The UI could add a "Sync Association Status" or a button of some sort for this. But probably Dani should decide this.

danipt: I do not think this is critical at this moment. If someone aims to use this value, they will need to release a new version and the current unique manner of "upgrade" is unassociated and associate/install again.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+The SDK already receives `commerceBaseUrl` and `commerceEnv` at every association and
+unassociation event. The missing piece is a standard persistence mechanism tied to the association lifecycle and typed helpers for runtime actions to consume it.
+
+**Goals:**
+
+- Store the Commerce Base URL and deployment type when an app is associated with an instance.
+- Clear the stored data when the app is unassociated.
+- Expose typed async helpers that any runtime action can use to retrieve the data and construct a ready-to-use Commerce HTTP client.
+- Work for all app types, including those without a custom installation flow.
+
```
danipt, line 34: Provide a helper for an App Developer is able to retrieve the system configuration that does not have inheritance or concept of send the scope id

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+## Developer experience
+
+After this feature ships, any runtime action can call the Commerce API without custom storage
+setup or client construction boilerplate.
+
+**Primary pattern — get a ready-to-use client:**
+
+```ts
+import { getAssociatedCommerceClient } from "@adobe/aio-commerce-lib-app";
```
danipt: Why do we want to call it `getAssociatedCommerceClient`? You can not have getAssociatedCommerceClient & getInstalledCommerceClient, etc... I think the name of the function could create confusion for me, would be much clearer getCommerceClient, and if it is not a commerce instance, return false or null

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+### Storage
+
+The association data is stored using the infrastructure already established in
+**`aio-commerce-lib-config`** — specifically, the shared `getSharedState()` utility from
+`aio-commerce-lib-config/source/utils/repository.ts`, which provides a lazy-initialized
+Adobe I/O State client shared across the SDK.
+
+A new module is added to `aio-commerce-lib-config` specifically for association data,
+separate from the existing Business Configuration module. It uses the same `getSharedState()`
+client but stores under a dedicated reserved key (`association`) rather than the
```
danipt: Why not call it `system`? In that manner, we will have all the configuration defined by the App under `configuration.*` and the ours inside `system.*`

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+
+The data is stored with the maximum TTL of 1 year (31536000 seconds). The default TTL of 24 hours is not appropriate here — association data is long-lived and must not expire on its own. It should only be cleared explicitly on unassociation. Re-associating resets the TTL.
+
+This has two important properties for this use case:
+
+- **Package-agnostic.** Adobe I/O State is shared across all actions within the same App
+  Builder application, regardless of which OpenWhisk package they belong to.
+- **No parameter drilling.** The helper reads directly from state. Callers do not need to
+  pass the data through every layer of the call stack.
+
```
danipt, line 117: In this use case, we do not need to have the inheritance concept. System configuration will always go at the global scope.

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,297 @@
...
+The association data is stored using the infrastructure already established in
+**`aio-commerce-lib-config`** — specifically, the shared `getSharedState()` utility from
+`aio-commerce-lib-config/source/utils/repository.ts`, which provides a lazy-initialized
+Adobe I/O State client shared across the SDK.
+
+A new module is added to `aio-commerce-lib-config` specifically for association data,
+separate from the existing Business Configuration module. It uses the same `getSharedState()`
+client but stores under a dedicated reserved key (`association`) rather than the
+`configuration.{scopeCode}` keys used by Business Configuration.
+
```
danipt, line 106: Please also document how to retrieve this type of configuration. Do we want to reuse the existing `getConfiguration` and `getConfigurationByKey` or do we want to build a `getSystemConfigurationByKey` that does not have the inheritance complexity? I prefer the second option

specs/features/CEXT-6160-association-data.md
vinayrao2000: > 
> [@vinayrao2000](https://github.com/vinayrao2000) could you elaborate on this TTL refresh you have in mind? not sure if that would actually be needed here.


The TTL refresh idea was: every time getAssociationData is called (i.e., when a runtime action reads the config), we'd re-put the data with a fresh 1-year TTL.

**How it works:**
- App is associated on Jan 1 --> TTL set to expire Jan 1 next year
- Any runtime action that calls Commerce APIs triggers getAssociatedCommerceClient --> which internally calls getAssociationData --> which reads AND re-puts the value with a fresh 1-year TTL
- Every subsequent read pushes the expiry further into the future

**Why it works in practice:**
Active apps constantly make Commerce calls (UI interactions, webhooks, scheduled jobs), so the TTL gets refreshed many times a day under normal usage. The 1-year window only matters if the app is completely idle (no UI usage, no webhooks, no jobs) for a full year, which is rare.

iivvaannxx: @vinayrao2000 I don't think this is a good option. This data will be in most cases write-once and read-heavy. Having to pay the cost of a write everytime we read will most likely duplicate latency as per two different roundtrips (and fire-and-forget is not and option because that doesn't work in App Builder).

I would drop such an approach TBH. If you are going to store using `lib-config` you'll get persistence for free, and you can instead just cache with a very large TTL.

(@oshmyheliuk I will unresolve this conversation until we align on this. Just FYI, as you resolved per your original question but now I have this other related concern)

specs/features/CEXT-6160-association-data.md
vinayrao2000: That's already the design intent. The helpers (getAssociatedCommerceInstance / getAssociatedCommerceClient) take only the standard params object and return the stored data directly, no scope IDs, no inheritance. I'll make this explicit in the Goals section.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Good point, getCommerceClient is simpler and more discoverable. I'll rename it in the spec. Returns null if the app isn't associated, same behavior as before.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Good point on DX, forcing a null check at every call site is bad for a public helper. The reason I went with returning null was to gracefully handle the legacy SDK case: apps that were associated before this feature was introduced won't have stored data, and throwing would crash them unexpectedly.

Also just flagging, Daniel suggested returning null/false for this case in his comment on R50. So we have two opposite opinions on the same code path.

Two ways to reconcile:
- Throw with a clear AppNotAssociatedError, code stays clean, and legacy apps would get a descriptive error pointing to re-association as the fix. Cleanest DX but breaking for legacy apps.
- Keep null but document it, null only happens for not-yet-associated or legacy apps, both of which are recoverable by re-associating. Forces a null check but no crashes.

@danipt  let me know your take. Happy to implement either once you two agree.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Good suggestion, For this iteration can we keep the scope to just baseUrl and env to match the AC, but the shape is designed to be extensible, adding projectId, workspaceId, etc., is a straightforward or you want to add them into this spec please suggest.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Yes, System configuration is app-level (global) and doesn't need the scope-tree / inheritance model. The current design stores under a single key, so retrieval is a direct lookup with no fallback chain. I'll make this explicit in the spec to confirm the design intent.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Good question. Currently I have specific helpers (getAssociationData, etc.) that wrap the storage. Happy to add a generic getSystemConfigurationByKey(key) underneath for SDK-internal use, that way the SDK has a consistent way to manage system data (association, future events, etc.) without inheritance complexity, while public API stays strongly typed (getAssociationData() calls it with "system.association"). I'll update the spec to document both layers.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Agreed, going with system.* makes the user/SDK separation cleaner. I'll use system.association as the key so the specific data is still obvious, and that also leaves room for future SDK-managed data (system.events, etc.) under the same namespace.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Fair point, lib-config should stay domain-agnostic. I'll refactor:

- lib-config exposes generic primitives, setSystemConfigByKey(key, value), getSystemConfigByKey(key), deleteSystemConfigByKey(key), no knowledge of "association" or any App Management concept.
- lib-app has the domain-aware helpers, setAssociationData / getAssociationData / clearAssociationData, which internally call the generic lib-config APIs with the "system.association" key.

This also aligns with Dani's earlier suggestion about a generic getSystemConfigurationByKey. I'll update the spec accordingly.

specs/features/CEXT-6160-association-data.md
danipt: Is it not better to do something like getSystemConfigurationByKey("BASE_URL") and getSystemConfiguration(), something like that?

specs/features/CEXT-6160-association-data.md
iivvaannxx: They should care about their application not throwing. If they modify their code to add this new helper and that's their situation they should try/catch it. The difference with returning null is that an application not having such data it's quite literally an "exception" (from this point onwards it should never happen), so it makes sense to me to throw an error better than return null.

specs/features/CEXT-6160-association-data.md
iivvaannxx: That sounds better to me, we don't have a `delete` operation, clearing a value is usually done by invoking `set` with a `null` value.

specs/features/CEXT-6160-association-data.md
iivvaannxx: Fair enough 👍🏻

specs/features/CEXT-6160-association-data.md
iivvaannxx: @danipt Yes, but there are two separate domains in play here. `getSystemConfiguration` helper belongs to `lib-config`, while @vinayrao2000's `getAssociatedCommerceInstance` (or `getComerceInstance` after the rename) belongs to `lib-app` and uses `getSystemConfiguration` internally. It's kind of a layered helper to improve DX.

specs/features/CEXT-6160-association-data.md
iivvaannxx: I think that @danipt's point is that you need to create a new scope that can not be inherited, as it's unique or "special". The `global` scope would work, but it would probably be inherited by other scopes down the tree.

I think the implementation kind of supports this, but I'm not 100% sure, so double-check before working on that.

specs/features/CEXT-6160-association-data.md
iivvaannxx: Again, see [this comment](https://github.com/adobe/aio-commerce-sdk/pull/460#discussion_r3343352012), those helpers are not mutually exclusive, they just belong to separate packages/domains.

vinayrao2000: Got it, aligned with that. Will update the spec to document both layers (lib-config generic + lib-app typed wrapper).

specs/features/CEXT-6160-association-data.md
vinayrao2000: Thanks for clarification, I have verified and the design is supported. getSharedState() / getSharedFiles() are raw storage utilities with no inheritance logic; the scope-tree/inheritance is contained in the configuration module. I'll add a new system-config module that uses those primitives directly with system.{key} keys, bypassing the scope-tree entirely.

specs/features/CEXT-6160-association-data.md
vinayrao2000: Fixed in 07a1aae

specs/features/CEXT-6160-association-data.md
vinayrao2000: Fixed in bd84bf8750eaa4d30e450c97f36d1a8739bce0c0

specs/features/CEXT-6160-association-data.md
vinayrao2000: Fixed in bd84bf8750eaa4d30e450c97f36d1a8739bce0c0

specs/features/CEXT-6160-association-data.md
vinayrao2000: Fixed in ccd61dea3075faaa0f44116e6cbbbd464874080c

specs/features/CEXT-6160-association-data.md
vinayrao2000: Fixed across 0457e1ede662c48a11eefd8dae030f529c643932 , 42dfcccc231e6759d748770c0bfc55ae1fc0f2d3 ,
4eed58200477af76e5a1d3142e334db560e362ef

specs/features/CEXT-6160-association-data.md
vinayrao2000: Fixed in 4eed58200477af76e5a1d3142e334db560e362ef

specs/features/CEXT-6160-association-data.md
vinayrao2000: Fixed in 434ebe20e8b8572ca8d9dc05dd91ff58eb2079d1

specs/features/CEXT-6160-association-data.md
vinayrao2000: Fixed in 0457e1ede662c48a11eefd8dae030f529c643932

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,317 @@
...
+A new generic `system-config` module is added to `aio-commerce-lib-config` for any
+SDK-managed system data, separate from the existing Business Configuration module. It uses
+the same `getSharedState()` client but stores under `system.{key}` keys (e.g.
+`system.association`, future `system.events`) rather than the `configuration.{scopeCode}`
+keys used by Business Configuration. The `system.*` namespace keeps SDK-managed config
+cleanly separated from app-defined `configuration.*` keys. The module is fully
+domain-agnostic — it has no knowledge of "association" or any App Management concept; the
+association-aware logic lives in `aio-commerce-lib-app` and uses this generic module
+underneath.
```
iivvaannxx: Sounds good, but I would just keep everything in the `configuration` module, and spawn a nested `system` folder in it. In the end, the concepts are close enough for the module to be the same. But your call.

specs/features/CEXT-6160-association-data.md
oshmyheliuk: If this issue is already solved by lib-config, let's use it.

@iivvaannxx Does `lib-config` use a two-layer cache, first checking `lib-state` and only then `lib-files`? What if the value expired in `lib-state`? Would it be re-saved from `lib-files` to `lib-state,` or would it always be read from `lib-files`?

iivvaannxx: @oshmyheliuk Yes, `lib-config` works as you described. As far as I know, the TTL of `lib-state` is managed by the underlying system; once that time expires, the entry is cleared, so the next request will not find it. 

After TTL, `lib-config` will check `aio-lib-state`, return `null` because the value has expired and been cleared, and only then go to `lib-files` (and save again in the cache).

vinayrao2000: Thanks for the confirmation, both. Spec updated, dropped the TTL refresh approach and now documents that lib-config uses a two-layer storage pattern (lib-files for persistence + lib-state as a performance cache), with automatic fallback when the cache entry expires.

Fixed in 2ffbd155a61cdaab0b5e5094479dce80b2a70688

specs/features/CEXT-6160-association-data.md
```diff
@@ -0,0 +1,316 @@
...
+  env: "saas" | "paas";
+};
+```
+
+The shape is designed to extend future fields (e.g. `projectId`, `workspaceId`) can be
+added without a breaking change.
+
+### New `association` runtime action
+
+A standalone `association` runtime action is added to `aio-commerce-lib-app`. It is always
```
iivvaannxx, line 159: Did we agree on a new runtime action or a new endpoint?

iivvaannxx (approved):

specs/features/CEXT-6160-association-data.md
vinayrao2000: As discussed on slack, the team agreed to implement separate runtime actions.

iivvaannxx: Got it, didn't recall that 👍🏻

PR #461, opened 2026-05-22
CEXT-6231: add app icon support spec

obarcelonap:
## Description

Spec for CEXT-6231: adds `icon` support to `app.commerce.config.ts`. Developers declare a relative path to a PNG or JPG file; the `pre-app-build` hook validates it (existence, ≤1 MB, 48×48 px), adds it to the `include` list of the `app-config` action in `ext.config.yaml` so App Builder deploys it as a binary file alongside the action, and computes a short SHA-256 content hash. The `app-config` action exposes a public `/app-config/icon` endpoint that reads the file via `fs` and serves the binary with `Cache-Control: public, max-age=31536000, immutable`. Every `app-config` response includes an `iconUrl` with the content hash appended as `?v={hash}`, so a redeployed icon is immediately available with no stale-cache window.

## Related Issue

https://jira.corp.adobe.com/browse/CEXT-6231

## Motivation and Context

Directly-deployed apps (outside Adobe Exchange) have no supported path to provide an icon and always show a gray placeholder in Commerce App Management. This spec closes that gap for all distribution models.

## How Has This Been Tested?

Spec only — no code changes.

## Types of changes

## Checklist:


specs/features/CEXT-6231-app-icon-support.md
```diff
@@ -0,0 +1,294 @@
...
+
+Add an `icon` field to `app.commerce.config.ts` so Commerce app developers can
+declare an icon for their app. The value is a relative path to a local image
+file; the `pre-app-build` hook encodes it to a base64 data URL and bundles it
+into the `app-config` action at build time, which exposes a dedicated
+`/app-config/icon` endpoint that returns the data URL with aggressive browser
+caching. The `app-config` response includes an `iconUrl` pointing to that endpoint,
+making the icon appear in Commerce App Management for all distribution models.
+For Exchange-listed apps that also declare an icon in their config, the
+SDK-declared icon takes precedence over the icon sourced from App Registry.
```
iivvaannxx, line 18: Even if the icon is small, I am sort of against bundling it. While it works, I don't think such an asset should be served as part of the action's code.

For me, a better alternative would be to use `include` and make App Builder deploy the icon as a file next to the action's code, which can then be read via `fs` module, and returned as preferred. 

Image

specs/features/CEXT-6231-app-icon-support.md
```diff
@@ -0,0 +1,312 @@
...
+    id: "my-commerce-app",
+    displayName: "My Commerce App",
+    description: "Extends Commerce checkout with fraud scoring.",
+    version: "1.0.0",
+    icon: "assets/icon.png",
+  },
+});
+```
+
+The value is a path relative to the project root. The `pre-app-build` hook
```
iivvaannxx: These are equivalent, but I feel it's clearer this way.

```suggestion
The value is a path relative to the `app.commerce.config` file. The `pre-app-build` hook
```

specs/features/CEXT-6231-app-icon-support.md
```diff
@@ -0,0 +1,312 @@
...
+The `pre-app-build` hook already reads `app.commerce.config.ts` to generate
+`ext.config.yaml`. When `metadata.icon` is set, the hook additionally:
+
+1. Resolves the path relative to the project root
+2. Checks the file exists — fails with a clear error if not
+3. Checks the file size is ≤ 1 MB — fails with a clear error if not
+4. Reads the image dimensions using the `image-size` package (reads only the
+   file header, supports PNG and JPEG) and checks they are exactly 48×48 px —
+   fails with a clear error if not
+5. Adds the icon file to the `include` list of the `app-config` action in the
```
iivvaannxx, line 171: Do we need a lib for this? Isn't there any utility in `fs` that can also give the size of the image? In the end it's just a file in the filesystem.

specs/features/CEXT-6231-app-icon-support.md
```diff
@@ -0,0 +1,312 @@
...
+   generated `ext.config.yaml`, so App Builder deploys it as a file alongside
+   the action code:
+
+```yaml
+actions:
+  app-config:
+    function: actions/app-config/index.js
+    include:
+      - ["assets/icon.png", "assets/icon.png"]
```
iivvaannxx, line 180: Take a look at [this thread](https://magento.slack.com/archives/C013UDBFBGB/p1768326478624029) when deciding how to map this path.

specs/features/CEXT-6231-app-icon-support.md
```diff
@@ -0,0 +1,312 @@
...
+
+The `app-config` action exposes a new public `GET /icon` route (path:
+`/app-config/icon`) that returns the binary image with the appropriate
+`Content-Type` (`image/png` or `image/jpeg`) and
+`Cache-Control: public, max-age=86400`. The route is a public web action — no
+auth is required, as icons are not sensitive, and `<img src>` cannot send
+authentication headers.
+
+The `app-config` response includes an `iconUrl` field whose value is the
+fully-qualified URL of this endpoint:
```
iivvaannxx: Do we need this? In the end the UI already knows the endpoint of the `app-config` via `workerProcess`, in the end it's just appending `/icon` to it isn't it?

specs/features/CEXT-6231-app-icon-support.md
```diff
@@ -0,0 +1,312 @@
...
+app's `AppConfig` (fetched from `app-config`) is already available at
+association time, making `iconUrl` a natural insertion point. `app-config`
+should become the authoritative source, with the existing `app.icon` field from
+App Registry (populated only for Exchange-listed apps) as a fallback for apps
+that have an Exchange listing but have not yet declared an icon in their config.
+
+### Edge cases
+
+- **Icon field absent** — no `iconUrl` in the `app-config` response; App
+  Management renders the existing gray placeholder.
```
iivvaannxx, line 246: Do we want to mix App Management UI things in this spec?

specs/features/CEXT-6231-app-icon-support.md
```diff
@@ -0,0 +1,312 @@
...
+- **App redeployed with a different icon** — the new file is deployed alongside
+  the action; it is served immediately after redeployment. Browsers that cached
+  the previous icon response may continue to show the old icon for up to 24
+  hours (the `max-age` TTL).
```
iivvaannxx: Is there a way to invalidate cache that this spec can define?

iivvaannxx (commented):
LGTM, but I have a feeling that the spec is somehow a bit coupled to the App Management UI. Makes sense, as in the end it's a UI concept, but as our "policy" is try to not mix the two, not sure if we should omit some information.

specs/features/CEXT-6231-app-icon-support.md
obarcelonap: `fs.stat` gives bytes, not width×height. Getting pixel
dimensions requires parsing the image header, which is exactly what `image-size` does.

specs/features/CEXT-6231-app-icon-support.md
obarcelonap: I prefer we keep it so the consumers (UI) do not need to apply any resolution logic, in case that eventually we change to CDN approach where the domain would change.

specs/features/CEXT-6231-app-icon-support.md
obarcelonap: Form my perspective is ok to keep App Management concerns without entering into implementation details since it is the main consumer of this API it needs to make sense.

specs/features/CEXT-6231-app-icon-support.md
obarcelonap: Yes — the spec now uses content-hash cache-busting. `pre-app-build` computes the first 8 hex chars of the icon file's SHA-256 and stores it alongside the icon. The action reads it at startup and appends it as `?v={hash}` when constructing `iconUrl`, so the URL changes whenever the icon content changes. Because the URL is now content-addressed, `Cache-Control` has been upgraded to `public, max-age=31536000, immutable`. A redeployed icon gets a new hash and a new URL, so browsers fetch it fresh immediately — no stale period.

iivvaannxx (approved):

jcuerdo (approved):

obarcelonap:
DO NOT MERGE THIS PR: after discussing with product, it seems icons should be provided out of the box in Exchange + Developer Console.
We're working internally to figure out if this could be possible.

obarcelonap:
won't do for the moment

PR #462, opened 2026-05-22
upgrade ky to v2 with FetchOptions adapter

obarcelonap:
## Description

Upgrades ky from v1 to v2 and introduces `FetchOptions` — an SDK-owned type that replaces ky's `Options` everywhere in the public API.

ky v2 changes hook signatures from positional args to state-object form, and renames `prefixUrl` to `prefix`. These would have been breaking changes for all three packages that exposed ky's `Options` directly. The `FetchOptions` adapter makes the ky version an internal implementation detail.

Key additions:
- `FetchOptions` type in `aio-commerce-lib-api` (exported) — `Omit<Options, "hooks" | "prefix" | "baseUrl"> & { hooks?: Hooks }` with positional hook signatures that expose all ky v2 state fields (`options: NormalizedOptions`, `retryCount`) while remaining backward compatible with fewer-arg callbacks
- `toKyOptions` adapter wraps the SDK's positional hook types to ky v2's state-object form at the ky boundary
- `HttpClientBase` method bindings intercept per-request options and run `toKyOptions`, so ~30 endpoint functions downstream needed only import/type changes
- CJS interop fix via `resolveKyDefaultExport` + `createKy` to handle ky v2's nested `.default` export in CommonJS environments
- `unwrapHttpError` simplified from async to sync — ky v2 pre-populates `HTTPError.data`
- `HTTPError` re-exported from `aio-commerce-lib-api` index; downstream packages (`aio-commerce-lib-events`, `aio-commerce-lib-webhooks`) now import it from the SDK rather than directly from ky

Supersedes #379.

## Related Issue

N/A

## Motivation and Context

ky v2 is a breaking upgrade that would cascade into major bumps for `aio-commerce-lib-api`, `aio-commerce-lib-events`, and `aio-commerce-lib-webhooks`. The adapter pattern keeps the bump to minor by making the ky version an internal detail. Re-exporting `HTTPError` from the SDK index gives consumers a stable import path insulated from future ky renames.

## How Has This Been Tested?

`pnpm test` and `pnpm typecheck` pass across all packages.

## Types of changes

## Checklist:


packages/aio-commerce-lib-api/source/utils/http/fetch-options.ts
```diff
@@ -0,0 +1,84 @@
...
+            hook(request, response),
+      ),
+      beforeError: hooks.beforeError?.map(
+        (hook) =>
+          ({ error }: { error: Error }) =>
+            error instanceof HTTPError ? hook(error) : error,
+      ),
+      beforeRetry: hooks.beforeRetry?.map(
+        (hook) => (state: BeforeRetryState) => hook(state),
+      ),
```
iivvaannxx, line 122: Do all these hooks map all the parameters? I find it weird that `v2` introduces an object parameter which only has a `request` in it (or `error` or others...), most likely there's are more things in that object that we need to forward?

.changeset/ky-v2-fetch-options-adapter.md
```diff
@@ -0,0 +1,7 @@
+---
+"@adobe/aio-commerce-lib-api": minor
+"@adobe/aio-commerce-lib-events": minor
+"@adobe/aio-commerce-lib-webhooks": minor
+---
+
+Upgrade ky to v2. Introduces `FetchOptions`, an SDK-owned type that replaces ky's `Options` in all public APIs. Hook signatures remain v1-compatible — existing `(request) => ...` hooks continue to work unchanged. The ky upgrade is now an internal implementation detail.
```
iivvaannxx: I just thought, another exposure point we have is `HTTPError`. @obarcelonap could also that become an issue in the future?

packages/aio-commerce-lib-api/source/lib/http-client-base.ts
```diff
@@ -10,27 +10,41 @@
...
+type HttpClient = {
+  get: HttpMethod;
+  post: HttpMethod;
+  put: HttpMethod;
+  delete: HttpMethod;
+  patch: HttpMethod;
+  head: HttpMethod;
+  retry: KyInstance["retry"];
+  stop: KyInstance["stop"];
+};
```
iivvaannxx, line 29: Any specific reason for replacing `Omit` for the custom `HttpClient` type? Not against it, but I find it redundant. Couldn't the adapter also workaround to any future change in `KyInstance`?

packages/aio-commerce-lib-api/source/utils/http/fetch-options.ts
obarcelonap: added all args as positional so clients can use all of them and is not breaking.

.changeset/ky-v2-fetch-options-adapter.md
obarcelonap: Good catch. HTTPError is now re-exported from aio-commerce-lib-api, so consumers can import it from the SDK rather than from ky directly, giving us one place to adapt if ky ever renames it.
Full runtime wrapping (subclass + interception hooks) is possible but adds significant complexity for a risk that hasn't materialized across two major ky versions.

packages/aio-commerce-lib-api/source/lib/http-client-base.ts
obarcelonap: The `Omit` approach would preserve ky's native method signatures, `KyInstance["get"]` accepts `Options`, not `FetchOptions`. The custom `HttpClient` type is what makes the public-facing get, post, etc. accept `FetchOptions`. Without it, the runtime wrapping in `setHttpClient` works but the declared parameter type stays as ky's `Options`, so the adapter wouldn't change the public API surface at the type level. That's exactly the problem we're solving.

iivvaannxx: Makes sense, I'm good with it as long as the interface is the same.

iivvaannxx:
Removed the code to fix the CJS interop issue as it's now handled by #470

obarcelonap:
@iivvaannxx LGTM, can you approve or provide feedback? thx

.changeset/ky-v2-fetch-options-adapter.md
```diff
@@ -0,0 +1,7 @@
+---
+"@adobe/aio-commerce-lib-api": minor
+"@adobe/aio-commerce-lib-events": minor
+"@adobe/aio-commerce-lib-webhooks": minor
+---
+
+Upgrade ky to v2. Introduces `FetchOptions`, an SDK-owned type that replaces ky's `Options` in all public APIs. Hook signatures use positional-argument form and expose all ky v2 state fields (`options: NormalizedOptions`, `retryCount`) — existing `(request) => ...` hooks continue to work unchanged. `HTTPError` is now exported from `@adobe/aio-commerce-lib-api` so consumers no longer need to import it directly from ky. The ky upgrade is now an internal implementation detail.
```
iivvaannxx: For me, changeset files should never leak implementation details.

iivvaannxx (approved):
LGTM, but as I remember, `lib-app` uses also this type when instantiating clients during `installation`, not sure if they also need updating. I guess if typecheck passes it means not, but would double check just in case.

.changeset/ky-v2-fetch-options-adapter.md
obarcelonap: changed

obarcelonap:
> LGTM, but as I remember, `lib-app` uses also this type when instantiating clients during `installation`, not sure if they also need updating. I guess if typecheck passes it means not, but would double check just in case.

Double-checked lib-app installation paths: they only set fetchOptions.timeout and pass resolved params through. @adobe/aio-commerce-lib-app typecheck passes, so no extra app changes were needed.

iivvaannxx:
Closed as `wontfix`. See #504 for context.

PR #469, opened 2026-05-26
Support overriding default runtime for generated runtime actions

aligent-lturner:
## Description

Support manually overriding the `runtime` of the runtime actions generated by `lib-app`. Once generated, the runtime will no longer get updated unless it's done manually.

## Related Issue

None

## Types of changes 

## Checklist:
  
   - [x] I have signed the Adobe Open Source CLA (https://opensource.adobe.com/cla.html).
   - [x] I have updated the documentation accordingly. (packages/aio-commerce-lib-app/docs/usage.md — new "Runtime" section under "Defining Configuration"; changeset added at .changeset/configurable-action-runtime.md)
   - [x] I have read the CONTRIBUTING document.
   - [x] I have read the DEVELOPMENT document.


obarcelonap:
Hi @aligent-lturner, thanks for your contribution.

I have mixed feelings about this request: I see value in allowing configuration of the node version used in the generated runtime actions but node version is a codegen/infra concern, not a commerce config concern.

Rather than always overwriting the runtime field in ext.config.yaml, I'd propose codegen could read the existing value and preserve it if already set. Users set `runtime: nodejs:24` directly in their `ext.config.yaml` once and codegen never touches it again. Zero new config surface, no schema changes, and it respects what developers already control manually. The tradeoff is making the YAML write logic aware of which fields to preserve vs. regenerate, but that's a useful general capability anyway.

If a config field really is needed, a dedicated codegen config surface would be a better home for it than `CommerceAppConfigSchema`. The schema describes what an app does; codegen settings describe how the scaffolding runs. Keeping them separate avoids polluting the commerce config with infrastructure knobs.

My preference for now is to preserve overwrites in ext.config.yaml as it's the simplest, has no new API surface, and makes codegen less aggressive in general.

@iivvaannxx thoughts?

iivvaannxx:
@aligent-lturner Thanks for the PR!

> thoughts?

@obarcelonap Was actually going to propose the same before I stumbled into your comment. I believe we implemented this recently (not released yet) for the `inputs` of the generated actions in #438. It's just a matter for doing the same with the node version and potentially other fields, as `annotations` can also benefit from this (e.g. for the `include-ims-credentials` needed for `aio-lib-db` custom setup in `installation`)

aligent-lturner:
@obarcelonap I don't really have a problem with the config living elsewhere, though I don't totally agree with the assertion that it's purely codegen/infra. We already define the node version for any actions we're creating, so I don't really see it being greatly different from that.

Personally, I'm not a fan of overwriting the value in `ext.config.yaml` - I think the generated actions and their manifest should be left alone (potentially even ignored in git entirely).

Using the version of node being used when the actions are generated would be another potential option, though that's a little less visible.

aligent-lturner:
@iivvaannxx @obarcelonap I've changed to use the suggested approach, preserving any existing value in `ext.config.yaml`

packages-private/scripting-utils/source/yaml/helpers.ts
```diff
@@ -165,3 +165,16 @@ export function getExistingInputs(existingAction?: YAMLMap) {
...
+
+/**
+ * Read a scalar string field from an existing YAML action definition. Used by
+ * codegen to preserve developer-set values (e.g. `runtime`) across
+ * regeneration instead of overwriting them.
+ *
+ * @param key - The field to read.
+ * @param existingAction - The YAML map representing the existing action.
+ */
+export function getExistingScalar(key: string, existingAction?: YAMLMap) {
```
obarcelonap: In YAML, a "scalar" covers strings, numbers, booleans, and nulls. This function narrows to `string | undefined`.
Could you please rename to `getExistingString` or similar?

.changeset/preserve-action-runtime.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": minor
+---
+
+Preserve a manually-set `runtime` on generated actions. Codegen no longer overwrites the `runtime` field in `ext.config.yaml`, so you can pin a different Node runtime (e.g. `nodejs:24`, whose image supports `require()` of ES modules) and it survives regeneration.
```
obarcelonap: CJS interop has been fixed in https://github.com/adobe/aio-commerce-sdk/pull/470 and is not related with Node runtime. 
Therefore, please remove from the changeset the following example
> (e.g. `nodejs:24`, whose image supports `require()` of ES modules)

obarcelonap (commented):
CJS interop has been fixed in #470, therefore we don't really need to merge this PR but I think it is a good addition to codegen.

I added some minor comments please try attend them so we can merge. 
Additionally, could you please reword the following
* PR title: something like "Preserve runtime field across action regeneration" seems more appropiate.
* PR description: update Description and Motivation sections.

.changeset/preserve-action-runtime.md
aligent-lturner: @obarcelonap ok to change to just "(e.g. `nodejs:24`)"?

packages-private/scripting-utils/source/yaml/helpers.ts
aligent-lturner: @obarcelonap done

.changeset/preserve-action-runtime.md
aligent-lturner: @obarcelonap have updated to remove reference to `require()` - also in `usage.md`

packages-private/scripting-utils/source/yaml/codegen.ts
```diff
@@ -85,13 +90,14 @@ function buildActionDefinition(
...
   const existingInputs = getExistingInputs(existingAction);
+  const existingRuntime = getExistingString("runtime", existingAction);
   const managedInputs = {
     LOG_LEVEL: "$LOG_LEVEL",
   };
 
   actionDef.set("function", action.function);
   actionDef.set("web", action.web ?? "yes");
-  actionDef.set("runtime", action.runtime ?? "nodejs:22");
+  actionDef.set("runtime", existingRuntime ?? action.runtime ?? "nodejs:22");
```
iivvaannxx, line 100: @obarcelonap Do we want to make the default `nodejs:24`?

iivvaannxx (approved):

iivvaannxx:
Should we update the PR description accounting for the new changes?

EDIT: Done, I kept it minimal.

packages-private/scripting-utils/source/yaml/codegen.ts
obarcelonap: Let's do that in a separate PR so we can document when to do that accordingly. I'll take care.

obarcelonap (approved):

PR #474, opened 2026-05-27
CEXT-6301: fix CI registry auth to prevent npm token exposure

obarcelonap:
## Description

Fixes a security issue where a plaintext npm auth token was committed to git by the CI release workflow, then refactors the publish infrastructure to prevent the same class of issue from recurring.

**Auth fix (`fix CI release auth writing token to workspace .npmrc`):** `prepare.ts` was writing the npm registry token as a plaintext value to `$GITHUB_WORKSPACE/.npmrc`. After a separate commit removed the checked-in `.npmrc` from the repo, that file became untracked in the workspace. `changesets/action` runs `git add -A` internally when creating a version PR, which committed the file — and the token — to `changeset-release/release`.

The fix removes all manual `.npmrc` manipulation. `setup-release` now runs `setup-node` without any registry configuration (clean install from public npm), then writes only a `${NODE_AUTH_TOKEN}` placeholder entry to `$HOME/.npmrc` after install — no token value is ever written to any file in any git-reachable path. `NODE_AUTH_TOKEN` is injected via `$GITHUB_ENV` so it is available to all subsequent steps without being committed anywhere.

**Publish registry decoupling:** The root cause also exposed an architectural gap — the publish destination (npm vs Artifactory) was implicit, determined by which `.npmrc` entries happened to be active at publish time. This PR makes it explicit:

- All public packages now declare `publishConfig.registry: "https://registry.npmjs.org"` and `publishConfig.access: "public"`, so `pnpm changeset publish` always targets npm by default
- For internal/preview releases (Artifactory), the `publish-snapshot` action overrides these fields at CI time using `npm pkg set --prefix` before publishing
- `setup-node` no longer receives `registry-url` or `scope` — this previously caused `ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION` 404s because pnpm's supply-chain check queried Artifactory for public `@adobe/aio-lib-*` packages that don't exist there

**Workflow refactor (`replace shared workflow with composite actions`):** The old `publish-shared.yml` reusable workflow was called by all three publish workflows with channel-specific parameters. Understanding any single release path required tracing through both the caller and the callee. Replaced with three composite actions that each workflow inlines directly:

- `.github/actions/setup-release` — pnpm, Node.js, install, publish auth
- `.github/actions/publish-snapshot` — changeset status, snapshot version, registry override, build, publish
- `.github/actions/notify-slack` — generate Slack announcement, post via webhook

Each publish workflow is now self-contained and readable end-to-end.

**Further cleanup:**
- Removed the dead channel-checking wrapper in `prepare.ts` — the function always ran in snapshot context
- Removed unused job-level outputs from `publish-internal.yml`
- Removed the `mirror` job from `publish-public.yml` — no longer needed since `publishConfig.registry` makes the publish destination explicit per-package

## Related Issue

https://jira.corp.adobe.com/browse/CEXT-6301

## Motivation and Context

A plaintext npm registry token was committed to a branch in this repo. Root cause:

1. `prepare.ts` wrote the auth token to `$GITHUB_WORKSPACE/.npmrc`
2. A separate commit deleted the checked-in `.npmrc` (making it untracked)
3. `changesets/action` uses `git add -A` internally — it committed the untracked `.npmrc` with the token to `changeset-release/release`

The token was rotated immediately after discovery. This PR ensures the same pattern cannot recur by removing all plaintext token writes from CI.

## How Has This Been Tested?

**Unit tests** — `prepare.ts` tests updated and passing:
```sh
pnpm --filter @aio-commerce-sdk/scripts test
pnpm typecheck
```

**Workflow test plan (run before merging):**

GitHub resolves relative composite action paths (`uses: ./.github/actions/…`) from the same branch the workflow runs on. All dispatch-capable workflows can therefore be tested from this branch before merge.

**Step 1 — `publish-internal` ✅**

Actions → Publish Internal → Run workflow → branch `ci/release-prep`, `Notify in Slack: false`.

Verified: install passes with no 404, snapshot packages published to Artifactory.

**Step 2 — `publish-preview` (in progress)**

PR #475 (`test/preview-ci → ci/release-prep`) was created as a test target — it branches from `ci/release-prep` (so the action files are present at the PR HEAD) and contains a changeset.

Actions → Publish Preview → Run workflow → branch `ci/release-prep`, PR `475`.

Verify:
- `prepare` job resolves the PR head SHA and finds the changeset
- `release` job completes and publishes snapshot packages
- `comment` job posts the package table to PR #475 with `alpha` tag

**Step 3 — `publish-public`**

Cannot be triggered in isolation (requires a real release). The `mirror` job has been removed; public packages now publish directly to npm via `publishConfig.registry`. Verify at the next scheduled `publish-public` run after merge.

N/A

## Types of changes

## Checklist:


obarcelonap:
publish-internal
* workflow: https://github.com/adobe/aio-commerce-sdk/actions/runs/26571498443

publish-preview
* workflow: https://github.com/adobe/aio-commerce-sdk/actions/runs/26575426989
* pr: https://github.com/adobe/aio-commerce-sdk/pull/475#issuecomment-4564093721
 
publish-public will be tested in the next public release.

iivvaannxx (approved):

PR #478, opened 2026-05-28
CEXT-6301: Simplify release promotion and back-sync

obarcelonap:
## Description

Replaces the PR-based promotion and back-sync gates with direct merge commits, reducing the public release flow from four manual steps to two (trigger promote, merge the Version Packages PR).

Also adds a deploy key so the CI workflows can push directly to the branch-protected `main` and `release` branches, regenerates API reference docs as part of the Version Packages PR, and documents the promotion flow with a diagram.

## Related Issue

CEXT-6301

## Motivation and Context

The promotion PR (main → release) and back-sync PR (release → main) were machine-generated and always merged as-is — they added friction without meaningful review value. Removing them shortens the release cycle and eliminates two no-op approval steps.

## How Has This Been Tested?

Infrastructure change — no automated tests apply. The deploy key and ruleset bypass have been configured on the repo. End-to-end verification requires triggering the promote workflow against a real release.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Collapse auto-generated API reference diffs in GitHub PRs (e.g. the Version Packages PR)
+packages/*/docs/api-reference/** linguist-generated=true
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 # Collapse auto-generated API reference diffs in GitHub PRs (e.g. the Version Packages PR)
 packages/*/docs/api-reference/** linguist-generated=true
+packages-private/*/docs/api-reference/** linguist-generated=true
--- a/.github/scripts/promote.sh
+++ b/.github/scripts/promote.sh
@@ -18,5 +18,8 @@ git config user.email "github-actions[bot]@users.noreply.github.com"
 git config user.name "github-actions[bot]"
 
 git checkout release
+# --no-ff is required: back-sync.sh finds this merge commit via `git log --merges`
+# to recover SHORT_SHA for its own commit message. A fast-forward would leave no
+# merge commit to grep for, breaking SHA correlation.
 git merge --no-ff "$SHA" -m "ci: promote main to release (${SHORT_SHA})"
 git push origin release
--- a/.github/scripts/promote.sh
+++ b/.github/scripts/promote.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Creates a promotion PR from a commit on main to the release branch.
+# Promotes a commit from main to the release branch via a direct merge commit.
 #
 # Usage: promote.sh [commit_sha]
 #   commit_sha  Optional commit SHA to promote (defaults to latest on origin/main)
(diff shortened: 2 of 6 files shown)


.github/scripts/promote.sh
```diff
@@ -1,50 +1,22 @@
...
-### What happens next
-
-1. Merge this PR into \`release\`
-2. The release workflow creates a **Version PR** with proper version bumps and changelogs
-3. Merge the Version PR to publish to NPM
-4. A back-sync PR from \`release\` → \`main\` is auto-created
-EOF
-)"
+git checkout release
+git merge --no-ff "$SHA" -m "ci: promote main to release (${SHORT_SHA})"
```
iivvaannxx, line 24: When going from `main` to `release` we should actually be able to do a fast forward shouldn't we?

.github/workflows/publish-public.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         id: changesets
         uses: changesets/action@63a615b9cd06ba9a3e6d13796c7fbcb080a60a0b # v1.8.0
         with:
-          version: pnpm changeset version
+          version: pnpm changeset version && pnpm turbo docs
```
iivvaannxx, line 39: That's actually really simple but clever 🙈

.github/workflows/publish-public.yml
```diff
@@ -60,12 +60,12 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
-      pull-requests: write
     steps:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
```
iivvaannxx, line 68: what is this `DEPLOY_KEY`?

iivvaannxx (approved):

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Collapse auto-generated API reference diffs in GitHub PRs (e.g. the Version Packages PR)
+packages/*/docs/api-reference/** linguist-generated=true
```
iivvaannxx, line 2: Private packages also have auto-generated docs.

.github/scripts/promote.sh
obarcelonap: needed if we want to be able to correlate promotion/back sync commits

.gitattributes
obarcelonap: added to this list

.github/workflows/publish-public.yml
obarcelonap: added documentation

.github/scripts/promote.sh
obarcelonap: Good catch — it's intentional. `back-sync.sh` finds this merge commit via `git log --merges --grep="promote main to release"` to recover the promoted SHA for its own commit message (SHA correlation). A fast-forward would produce no merge commit to grep for, breaking that link. Added a comment to the script to make this clear.

.github/workflows/publish-public.yml
obarcelonap: Thanks! The changesets action accepts an arbitrary shell command for `version`, so we just chain the doc regeneration right after the version bump.

.github/workflows/publish-public.yml
obarcelonap: It's an SSH deploy key with write access to the repo, stored as an Actions secret. `promote.sh` and `back-sync.sh` push directly to protected branches (`main` and `release`), which normally requires a PR. The deploy key's ID is registered as a bypass actor in the branch ruleset so those CI pushes can land without one. Added a section to DEVELOPMENT.md documenting it and how to rotate it.

.gitattributes
obarcelonap: Good catch — fixed to also cover `packages-private/`.

PR #480, opened 2026-05-29
CEXT-6096: spec for Admin UI grid column extensions (v2)

obarcelonap:
## Description

Adds a spec for the v2 Admin UI SDK grid column schema, covering order, product, and customer grids.

## Related Issue

https://jira.corp.adobe.com/browse/CEXT-6096

## Motivation and Context

The v1 grid column design has three main problems: it requires API Mesh even for trivial cases, `meshId` is opaque and environment-specific, and the wire contract between Commerce and the handler is undocumented. The v2 schema drops Mesh as an SDK-config concept, replaces `data.meshId` with `data.backend` (a `workerProcess` operation name), unifies the three grids under one schema shape, and formally specifies the wire contract.

v1 (`commerce/backend-ui/1`) is removed with no compatibility window. All affected types are `@experimental`, so this is a `minor` bump.

## How Has This Been Tested?

Spec-only change — no code modified.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
+++ b/specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
@@ -17,7 +17,9 @@ order, product, and customer grids on `commerce/backend-ui/2`. The v1 schema (`a
 will be removed in a future release — it remains functional for now. The v2 shape introduces
 `runtimeAction` (a `workerProcess` operation name resolved by App Registry at runtime), `columns`
 (replacing `properties`), and `key` (replacing `columnId`). The wire contract between Commerce and
-the handler is formally standardized for the first time.
+the handler is formally standardized for the first time. No separate registration action is
+generated for `commerce/backend-ui/2` — Commerce reads the `adminUi` config directly from the
+existing `app-config` endpoint on `commerce/extensibility/1`.
 
 ## Motivation
 
@@ -130,6 +132,11 @@ developer needs to provide alongside the config is the handler implementation it
 `orders/fetch-order-grid-data.js`). App Registry resolves the operation name to the deployed
 runtime action URL when Commerce fetches the extension.
 
+Commerce reads the `adminUi` registration config from the `app-config` endpoint on
+`commerce/extensibility/1` — no additional runtime action is needed on `commerce/backend-ui/2`.
+The only artifact generated for `commerce/backend-ui/2` is the `ext.config.yaml` that declares
+the `workerProcess` operations.
+
 ### Wire contract
 
 **Request** (sent by Commerce to the handler):
@@ -207,14 +214,16 @@ until those extension points are ported to `commerce/backend-ui/2` in a subseque
 ### Two config keys, two extension points
 
 `adminUiSdk` and `adminUi` are independent top-level keys in `defineConfig()`. They map to separate
-extension points and separate generated artifacts:
+extension points with different generated artifacts:
 
-| Config key   | Extension point         | Status                       |
-| ------------ | ----------------------- | ---------------------------- |
-| `adminUiSdk` | `commerce/backend-ui/1` | Deprecated — will be removed |
-| `adminUi`    | `commerce/backend-ui/2` | Current                      |
+| Config key   | Extension point         | Generated artifact                                       | Status                       |
+| ------------ | ----------------------- | -------------------------------------------------------- | ---------------------------- |
+| `adminUiSdk` | `commerce/backend-ui/1` | `registration/index.js` + `ext.config.yaml`              | Deprecated — will be removed |
+| `adminUi`    | `commerce/backend-ui/2` | `ext.config.yaml` with `workerProcess` declarations only | Current                      |
 
-Both keys are optional and can coexist. The SDK generates artifacts for whichever are present.
+For `adminUi`, no registration action is generated. Commerce reads the `adminUi` config directly
+from the `app-config` endpoint on `commerce/extensibility/1`, which already includes the full app
+manifest. Both keys are optional and can coexist.
 
 ### Schema
 
@@ -279,22 +288,22 @@ export const BACKEND_UI_V2_EXTENSION_POINT_ID = "commerce/backend-ui/2"; // v2
 The `Extension` union includes both values. Each routes to its own handler:
 
 - `"backend-ui/1"` → `generateRegistrationActionFile` (v1, unchanged behavior)
-- `"backend-ui/2"` → `updateExtConfig` + `generateGridColumnsRegistrationActionFile` (v2)
+- `"backend-ui/2"` → `updateExtConfig` (v2)
 
-When `hasAdminUi` is true, the v2 path performs two generation steps:
+When `hasAdminUi` is true, the v2 path performs one generation step:
 
-1. **Registration action** — generates the JS file that returns the registration config when
-   Commerce calls the extension.
+**`ext.config.yaml` workerProcess declarations** — collects all unique `runtimeAction` values
+across `order`, `product`, and `customer` grid column configs, then writes them as `workerProcess`
+entries in `ext.config.yaml` for `commerce/backend-ui/2`. No registration action is generated;
+Commerce reads the `adminUi` config directly from the `app-config` endpoint on
+`commerce/extensibility/1`. Developers do not declare `workerProcess` entries manually; the SDK
+derives them from `app.commerce.config.ts` and keeps `ext.config.yaml` in sync on every build.
 
-2. **`ext.config.yaml` workerProcess declarations** — collects all unique `runtimeAction` values
-   across `order`, `product`, and `customer` grid column configs, then writes them as `workerProcess`
-   entries in `ext.config.yaml` for `commerce/backend-ui/2`. Developers do not declare these
-   manually; the SDK derives them from `app.commerce.config.ts` and keeps `ext.config.yaml` in sync
-   on every build.
-
-Example generated `ext.config.yaml` fragment for the config above:
+Example generated `ext.config.yaml` for the config above:
 
 ```yaml
+hooks:
+  pre-app-build: "EXTENSION=backend-ui/2 $packageExec aio-commerce-lib-app hooks pre-app-build"
 operations:
   view:
     - type: web
@@ -306,13 +315,8 @@ operations:
       impl: products/fetch-product-grid-data
     - type: action
       impl: customers/fetch-customer-grid-data
+web: web-src
 ```
-
-### Registration action generation
-
-No template change is needed. The registration action inlines the config as a JS object literal;
-`runtimeAction` is serialized as-is. Commerce reads `runtimeAction` directly from the registration
-JSON — no field mapping is required on either side.
 
 ### Changeset bump
 
--- a/specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
+++ b/specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
@@ -16,7 +16,7 @@ order, product, and customer grids on `commerce/backend-ui/2`. The v1 schema (`a
 `commerce/backend-ui/1`, `data.meshId`, `properties`, `columnId`, `float` type) is deprecated and
 will be removed in a future release — it remains functional for now. The v2 shape introduces
 `runtimeAction` (a `workerProcess` operation name resolved by App Registry at runtime), `columns`
-(replacing `properties`), and `key` (replacing `columnId`). The wire contract between Commerce and
+(replacing `properties`), and `columnId` (unchanged from v1). The wire contract between Commerce and
 the handler is formally standardized for the first time. No separate registration action is
 generated for `commerce/backend-ui/2` — Commerce reads the `adminUi` config directly from the
 existing `app-config` endpoint on `commerce/extensibility/1`.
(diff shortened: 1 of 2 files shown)


specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
```diff
@@ -0,0 +1,299 @@
...
+   specified anywhere in the SDK.
+4. **Order, product, and customer grids are architecturally identical but treated as three separate
+   concerns.** Three pages of documentation and three schema branches describe the same pattern.
+
+**Goals**
+
+- Replace `data.meshId` with `data.backend` (a `workerProcess` operation name) in all three grids.
+- Rename `properties` → `columns`, `columnId` → `key` for clarity.
+- Formally specify the request/response wire contract.
+- Remove `commerce/backend-ui/1` and all v1 schema artifacts from the SDK.
```
danipt: In other words, does App management always generate `commerce/backend-ui/2`, right? Find the 1 will means an app created without app management.

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
```diff
@@ -0,0 +1,299 @@
...
+- Use a single unified column schema under `order.gridColumns`, `product.gridColumns`, and
+  `customer.gridColumns`.
+
+**Non-goals**
+
+- Supporting v1 in any compatibility mode or deprecation window.
+- Adding grid targets beyond order, product, and customer.
+- Implementing mass action v2 (separate ticket).
+- Wrapping or managing the Commerce API Mesh service.
+
```
danipt, line 60: Are we considering changes in AI migrate/build as part of that user story or we should create another one?

specs/features/CEXT-6096-grid-column-extensions.md
```diff
@@ -0,0 +1,299 @@
...
+## Developer experience
+
+### Defining grid columns
+
+The config shape is the same across all three grids:
+
+```ts
+// app-config.ts
+export default defineConfig({
+  adminUiSdk: {
```
danipt: Do we want to use the SDK term here? For me, it's a bit misleading.

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
```diff
@@ -0,0 +1,299 @@
...
+      workerProcess:
+        - type: action
+          impl: orders/fetch-order-grid-data
+        - type: action
+          impl: products/fetch-product-grid-data
+```
+
+### Wire contract
+
+**Request** (sent by Commerce to the handler):
```
danipt, line 142: How do you handle pagination? Or is this something handled in the invokation, and you receive the IDs in the current page?

specs/features/CEXT-6096-grid-column-extensions.md
```diff
@@ -0,0 +1,299 @@
...
+The constant name is preserved so callsites do not change. Only the value changes.
+
+### Pre-app-build hook
+
+`packages/aio-commerce-lib-app/source/commands/hooks/pre-app-build.ts`
+
+Remove `"backend-ui/1"` from the `Extension` union and replace with `"backend-ui/2"`. The hook
+continues to call `generateRegistrationActionFile` when `hasAdminUiSdk` is true — no structural
+change needed. The v2 config shape (with `backend` and `columns`) passes through the existing
+serialization path unchanged.
```
danipt: We need to add validation during this step to ensure the worker exists and ideally

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
```diff
@@ -0,0 +1,299 @@
...
+### Pre-app-build hook
+
+`packages/aio-commerce-lib-app/source/commands/hooks/pre-app-build.ts`
+
+Remove `"backend-ui/1"` from the `Extension` union and replace with `"backend-ui/2"`. The hook
+continues to call `generateRegistrationActionFile` when `hasAdminUiSdk` is true — no structural
+change needed. The v2 config shape (with `backend` and `columns`) passes through the existing
+serialization path unchanged.
+
+### Registration action generation
```
danipt: Do we prefix the columns somehow? What happens if 2 apps add an "external_id" column in the customer data grid?

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
obarcelonap: The proposal is that the sdk only offers support for v2.
But it could be that someone is using app management but still configuring manually the extension point `/commerce/backend-ui/1` as we do right now.

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
obarcelonap: Better to handle this separately since we need to:
* Create from scratch a skill for adminui and chain from init skill. 
* Adapt the migration skill (which would require some research as Sasha did initially)

If you agree I'll create two tickets for this work since it can be decoupled from the actual development of the feature.

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
obarcelonap: My assumption is that this is done in the commerce instance and it only adds in the request payload the necessary ids (pagination or not).

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
obarcelonap: I don't have this knowledge but I'd say that's on the extension to guarantee there could be no collisions.

specs/features/CEXT-6096-grid-column-extensions.md
obarcelonap: changed to `adminUi`

specs/features/CEXT-6096-grid-column-extensions.md
obarcelonap: can you elaborate on this

danipt: The configuration allows you to define a runtime action `runtimeAction: "fetch-order-grid-data",`, we should validate that this runtime action exists in the validation stage of the  pre-build hook, right?

obarcelonap: I adapted the spec so the workerProcess in ext.config.yaml is generated by the SDK so no need for validation. Hope this solves your validation concern.

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
danipt: But the App does not have an understanding of other Apps installed on this Commerce instance, similar to events and webhooks, we should have a mechanism that ensures no collision happens and one app can not overwrite another one already installed.

danipt (approved):

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
obarcelonap: correct, when I say extension I am referring to the in process extension (adobe-commerce-backend-uix) that's why I think this is not part of the SDK

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
```diff
@@ -0,0 +1,321 @@
...
+
+The config shape is the same across all three grids:
+
+```ts
+// app-config.ts
+export default defineConfig({
+  adminUi: {
+    registration: {
+      order: {
+        gridColumns: {
```
danipt: Should we add some titles and descriptions here to be used for define the admin UI component during the installation process?

danipt (approved):

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
obarcelonap: added

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
```diff
@@ -0,0 +1,379 @@
...
+`packages/aio-commerce-lib-app/source/commands/constants.ts`
+
+Both constants coexist:
+
+```ts
+export const BACKEND_UI_EXTENSION_POINT_ID = "commerce/backend-ui/1"; // v1, deprecated
+export const BACKEND_UI_V2_EXTENSION_POINT_ID = "commerce/backend-ui/2"; // v2
+```
+
+### Pre-app-build hook
```
danipt, line 289: As usual, include validation that the runtime action exists and is reachable, the sooner we provide errors in the configuration, the better from an App Developer perspective

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
```diff
@@ -0,0 +1,379 @@
...
+Rejected per "one concept, one place." The three grids have identical structure. `gridType` in the
+wire request lets a single handler serve multiple grids if the developer chooses; the SDK need not
+enforce separation.
+
+**Impact of not doing this.**
+Grid columns remain coupled to API Mesh, blocking apps that don't need federation, and the wire
+contract remains undocumented. The multi-environment `meshId` problem has no SDK-level solution.
+
+## Unresolved questions
+
```
danipt: We should check how to handle the Multi App support similar like in events/webhooks. What happens if 2 apps (one from CRM and another from ERP) include a field with key=external_id in the customer grid? We should always add a prefix to the key maybe?

danipt (approved):

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
obarcelonap: @asalloum5 do you think this should be part of the sdk or the extension?

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
obarcelonap: Good call, and fully agree on the fail-fast principle.
Note there's no precedent in the SDK: events and webhooks are not validating `runtimeAction` field.
This is because file existence is more indirect than it looks, runtimeAction is an operation name ("orders/fetch-order-grid-data"), not a file path. To check file existence you'd have to traverse ext.config.yaml → runtimeManifest → action → function path → disk.
This might help us to enforce the extension endpoint where to declare runtime actions:
- runtime actions related to events and webhooks should go into commerce/extensibility/1
- runtime actions related to adminui should go into commerce/backend-ui/2

We need to think also how this plays out with AI skills, probably build will fail after adding the `runtimeAction` config prop and the agent will notice and create the file accordingly.

If you don't mind I'll open a ticket to look into this validation pattern and implement it across the whole SDK.

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
```diff
@@ -0,0 +1,379 @@
...
+> see [Migration from v1](#migration-from-v1) below.
+
+## Summary
+
+The SDK introduces a v2 Admin UI grid column schema under the new `adminUi` config key, covering
+order, product, and customer grids on `commerce/backend-ui/2`. The v1 schema (`adminUiSdk`,
+`commerce/backend-ui/1`, `data.meshId`, `properties`, `columnId`, `float` type) is deprecated and
+will be removed in a future release — it remains functional for now. The v2 shape introduces
+`runtimeAction` (a `workerProcess` operation name resolved by App Registry at runtime), `columns`
+(replacing `properties`), and `key` (replacing `columnId`). The wire contract between Commerce and
```
asalloum5: why did we decide to use `key` and not `id`? I prefer to keep it the same across extension points

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
```diff
@@ -0,0 +1,379 @@
...
+
+Example generated `ext.config.yaml` for the config above:
+
+```yaml
+hooks:
+  pre-app-build: "EXTENSION=backend-ui/2 $packageExec aio-commerce-lib-app hooks pre-app-build"
+operations:
+  view:
+    - type: web
+      impl: index.html
```
asalloum5: view is not mandatory in that case since we don't display any iFrame (to be tested and confirmed)

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
asalloum5: Agree, we should add a pre-fix to avoid this. We can handle it in the sdk or in commerce directly. I don't have a strong opinion on this.

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
```diff
@@ -0,0 +1,379 @@
...
+enforce separation.
+
+**Impact of not doing this.**
+Grid columns remain coupled to API Mesh, blocking apps that don't need federation, and the wire
+contract remains undocumented. The multi-environment `meshId` problem has no SDK-level solution.
+
+## Unresolved questions
+
+**Single-operation constraint.**
+Can the same `runtimeAction` value appear in `order.gridColumns` and `customer.gridColumns`? The
```
asalloum5: I would say yes, if the runtime action knows how to process the grid, since I see that we're sending the gridType upon invoking it in the body.

asalloum5 (approved):

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
obarcelonap: removed question

specs/features/CEXT-6096-admin-ui-grid-column-extensions.md
obarcelonap: The spike proposed `key`, I thought this was intended. Nevermind, reverted to `columnId`

PR #481, opened 2026-05-29
CEXT-6138: add OpenAPI spec for lib-app actions with capability-based path stripping

iivvaannxx:
## Description

Adds a "manually" (AI) maintained OpenAPI 3.1.0 spec for all `@adobe/aio-commerce-lib-app` runtime actions, served at runtime via `GET /app-config/openapi.json`. The spec is linted with Redocly and its `info["x-meta"].packageVersion` is kept in sync with `package.json` automatically.

**Key changes:**

- **Router reorganization**: Actions restructured into per-route subdirectories under `source/actions/`, each with their own `router.ts` and `index.ts`
- **Capability-based path stripping**: The `GET /app-config/openapi.json` endpoint serves the full spec at rest but strips paths at runtime based on the app's declared capabilities (`hasAdminUiSdk`, `hasBusinessConfigSchema`, `requiresInstallation`)
- **OpenAPI spec** (`docs/openapi.json`): Full contract for all 7 action routers (`/app-config`, `/config`, `/installation`, `/registration`, `/scope-tree`, `/scope-tree/commerce`, `/installation/validation`, `/installation/uninstallation`). Includes 24 named components, Redocly linting via `pnpm run lint:openapi`
- **`v.any()` fix**: `PUT /scope-tree` body schema now uses the correct recursive `CustomScopeInput` valibot schema instead of `v.any()`

## Related Issue

[CEXT-6138](https://jira.corp.adobe.com/browse/CEXT-6138)

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/CLAUDE.md
obarcelonap, line 1: will this file picked up automatically by CLAUDE or we need to reference it somehow in the global AGENTS

specs/features/CEXT-6138-capability-discovery.md
obarcelonap, line 1: I'd update the spec

packages/aio-commerce-lib-app/docs/openapi.json
obarcelonap, line 1: The endpoint `GET /app-config` returns `openApiSpecUrl` but can't find it here

packages/aio-commerce-lib-app/source/actions/app-config/router.ts
```diff
@@ -74,19 +84,39 @@ router.get("/", {
...
-  async (params: RuntimeActionParams) => {
-    const handler = router.handler();
-    return await handler({
-      ...params,
-      appConfig,
+/**
+ * GET /openapi.json - Returns the OpenAPI spec for all SDK actions
+ * @internal - Do not add to OpenAPI Spec.
+ */
+router.get("/openapi.json", {
```
obarcelonap, line 100: We could run `buildOpenApiSpec` on pre build and here just serve the file. But I understand this is easier and due to the frequency of call and the client caching approach this is fine.

packages/aio-commerce-lib-app/source/actions/app-config/router.ts
```diff
@@ -74,19 +84,39 @@ router.get("/", {
...
+    const domains = getConfigDomains(rawParams.appConfig);
+
+    if (ck && getOpenApiCacheKey(domains) === ck) {
+      logger.debug(
+        `Received request for OpenAPI spec with cache key query param: ${ck}`,
+      );
+
+      return ok({
+        body: await buildOpenApiSpec(domains, logger),
+        headers: { "Cache-Control": "public, max-age=31536000, immutable" },
```
obarcelonap, line 113: I'd just always return the cache-control, there's no benefit from client side of this logic, they can always force the request
For instance
```
fetch(`/api/data?cacheBust=${Date.now()}`)
// or
fetch('/api/data', { cache: 'no-store' })
```

packages/aio-commerce-lib-app/package.json
```diff
@@ -87,6 +88,7 @@
     "README.md"
   ],
   "scripts": {
+    "version": "pnpx tsx scripts/sync-openapi-version.ts",
```
obarcelonap, line 91: will this work with changesets? a friend is flagging this as a problem

> The problem is that this relies on the npm/pnpm version lifecycle script, which runs for commands like pnpm version / npm version. But this repo uses Changesets. pnpm changeset version computes the release plan and writes package.json files directly; it does not run each changed package’s version lifecycle script.

packages/aio-commerce-lib-app/CLAUDE.md
iivvaannxx: Yes, Claude Code picks up `CLAUDE.md` files in package subdirectories automatically. The package-level file just re-exports `AGENTS.md` via `@AGENTS.md`, which is where the actual instructions live. No global reference needed.

specs/features/CEXT-6138-capability-discovery.md
iivvaannxx: Done in [`76e4b33c`](https://github.com/adobe/aio-commerce-sdk/pull/481/commits/76e4b33c2f1f769cc55f069c99401e09d1fcb470). The design section now reflects what was actually built: filtering at request time via `buildOpenApiSpec` (path stripping then schema pruning), with content-addressed caching via `openApiSpecUrl`. The pre-build approach was never implemented.

packages/aio-commerce-lib-app/docs/openapi.json
iivvaannxx: Good catch. Fixed in [`86c16045`](https://github.com/adobe/aio-commerce-sdk/pull/481/commits/86c16045da26b44edf19ccfb7d05608d2cd4b2d3): added `openApiSpecUrl` as a required `string`/`uri` property on the `GET /app-config` 200 response schema.

packages/aio-commerce-lib-app/source/actions/app-config/router.ts
iivvaannxx: Went with request-time for simplicity yes. Also I find it safer to make this server-side, as the `pre-build` happens client-side. The dynamic import also defers any I/O load (for loading this big file) to only this endpoint.

packages/aio-commerce-lib-app/source/actions/app-config/router.ts
iivvaannxx: The conditional is intentional. Cache busting works, but clients shouldn't have to think about it. `openApiSpecUrl` already embeds the current `ck`, so when the spec changes the URL changes and invalidation is automatic. Returning `immutable` on the bare path breaks that for anyone fetching it directly.

packages/aio-commerce-lib-app/package.json
iivvaannxx: Fixed in [`d610b0d6`](https://github.com/adobe/aio-commerce-sdk/pull/481/commits/d610b0d68c8e8139c35f3adbcf2491e22a2ce6c5): added a `version` Turbo task (`cache: false`, no deps) and slotted `pnpm turbo version` into the CI version command right after `pnpm changeset version`. Each package's `version` script now runs explicitly via Turbo on every release.

packages/aio-commerce-lib-app/source/config/schema/admin-ui-sdk.ts
```diff
@@ -326,13 +325,23 @@ export type BannerNotification = v.InferInput<typeof BannerNotificationSchema>;
...
+/** Any commerce app config — input contract or validated output. */
+type AnyCommerceAppConfig = CommerceAppConfig | CommerceAppConfigOutputModel;
+
+/** Config type when Admin UI SDK registration is configured. */
+export type AppConfigWithAdminUiSdk<
+  T extends AnyCommerceAppConfig = CommerceAppConfigOutputModel,
+> = T & {
+  adminUiSdk: NonNullable<T["adminUiSdk"]>;
+};
+
```
iivvaannxx: These type changes are required in the new endpoint. They are applied to all domains. It's because in the REST actions we receive the authored version (user input) of the config, and these methods were only accepting post-parse (output types, after validation)

packages/aio-commerce-lib-app/docs/openapi.json
obarcelonap: could you add an example so the ck query param is visible to users and agents.
```
"openApiSpecUrl": {
  "description": "URL of the OpenAPI spec for this deployment, scoped to the active capabilities. Includes a content-addressed cache key as a query parameter.",
  "type": "string",
  "format": "uri",
  "examples": [
    "https://sample-namespace.adobeioruntime.net/api/v1/web/app-management/app-config/openapi.json?ck=4f6a2b9c"
  ]
}
```

obarcelonap (approved):

packages/aio-commerce-lib-app/docs/openapi.json
iivvaannxx: Done in [`0a093e6c`](https://github.com/adobe/aio-commerce-sdk/pull/481/commits/0a093e6c38cc8eea62fb09718773b81c7284a82f).

jcuerdo (approved):

PR #489, opened 2026-06-02
CEXT-6318: Create a separate package for Admin UI SDK

oshmyheliuk:
## Description

This PR is a part of https://github.com/adobe/aio-commerce-sdk/pull/426. As ACL will need a rework based on the new structure, the new package for the admin ui can be delivered separately.

https://jira.corp.adobe.com/browse/CEXT-6318

## Motivation and Context

To align with other packages for eventing/webhook, the admin ui API should also be extracted to a separate package.

## Types of changes

## Checklist:


packages/aio-commerce-lib-admin-ui/source/api/extensions/endpoints.ts
```diff
@@ -0,0 +1,82 @@
...
+ * @param params - The extension registration parameters.
+ * @param fetchOptions - Optional Ky fetch options.
+ *
+ * @throws A `CommerceSdkValidationError` if the parameters are invalid.
+ * @throws An `HTTPError` if the status code is not 2XX.
+ */
+export async function registerExtension(
+  httpClient: AdobeCommerceHttpClient,
+  params: ExtensionRegistrationParams,
+  fetchOptions?: Options,
```
iivvaannxx, line 41: Just the same heads-up of #379. Nothing to change as it hasn't been merged, but just so you are aware of it in case it lands before this one.

packages/aio-commerce-lib-admin-ui/source/api/extensions/schema.ts
```diff
@@ -0,0 +1,37 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+import * as v from "valibot";
+
+/** Parameters for POST /V1/adminuisdk/extension. */
+export const ExtensionRegistrationParamsSchema = v.object({
+  extensionName: v.pipe(v.string(), v.minLength(1)),
+  extensionTitle: v.pipe(v.string(), v.minLength(1)),
+  extensionUrl: v.pipe(v.string(), v.url()),
```
iivvaannxx, line 19: Not a blocker but, I would extend this with [this same pattern check](https://github.com/magento-commerce/adobe-commerce-backend-uix/blob/7e7d75bdaac07079786baa484de27d22e5f33783/CommerceBackendUix/Model/Extensions/ExtensionManagement.php#L106) from the API validation. For the same reason as proposed in the other PR, early catching what will be a validation issue and avoiding a network round-trip.

packages/aio-commerce-lib-app/source/management/installation/admin-ui-sdk/helpers.ts
```diff
@@ -52,32 +46,37 @@ export async function registerExtension(context: AdminUiSdkExecutionContext) {
       ),
     );
 
-  logger.info(
-    `Admin UI SDK extension registered successfully: ${response.extensionId}`,
-  );
-
```
iivvaannxx, line 58: If the API returns a response, is there a specific reason we return `Promise<void>` from it?

iivvaannxx (approved):

packages/aio-commerce-lib-app/source/management/installation/admin-ui-sdk/helpers.ts
oshmyheliuk: Yes, there is a response with the extension Id, I udpated the code

packages/aio-commerce-lib-admin-ui/source/api/extensions/schema.ts
oshmyheliuk: I'm not sure that it's a good idea to copy that regex. If something changed on the admin ui sdk side, the code here will be stale. The API will reject bad inputs with its own error.
Probably, it would make sense to add that regex if this were schema validation from app.commerce.config

packages/aio-commerce-lib-admin-ui/package.json
```diff
@@ -0,0 +1,96 @@
...
+  ],
+  "bugs": {
+    "url": "https://github.com/adobe/aio-commerce-sdk/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/adobe/aio-commerce-sdk.git",
+    "directory": "packages/aio-commerce-lib-admin-ui"
+  },
+  "publishConfig": {
```
obarcelonap, line 30: After the adaptations we did in the ci workflows we need to add `publishConfig.registry` and `publishConfig.access`. 
See
https://github.com/adobe/aio-commerce-sdk/blob/c99a7d6a1d8b1f0d14314e16c68c6f98c2dddd8b/packages/aio-commerce-lib-api/package.json#L36-L37

Could you please add also this fields in the template so we don't miss them? thanks
https://github.com/adobe/aio-commerce-sdk/blob/c99a7d6a1d8b1f0d14314e16c68c6f98c2dddd8b/turbo/generators/create-package/template/package.json.hbs#L29-L43

packages/aio-commerce-lib-admin-ui/source/lib/api-client.ts
```diff
@@ -0,0 +1,37 @@
...
+import * as extensionEndpoints from "#api/extensions/endpoints";
+
+import type { CommerceHttpClientParams } from "@adobe/aio-commerce-lib-api";
+
+/**
+ * Creates a new API client for the Admin UI SDK API with all available operations.
+ *
+ * @param params - The parameters to build the Commerce HTTP client.
+ */
+export function createAdminUiSdkApiClient(params: CommerceHttpClientParams) {
```
obarcelonap: I'd propose to drop `Sdk` from the fn name and review all this pr so SDK term is not included anywhere, just `AdminUi`

packages/aio-commerce-lib-admin-ui/source/api/extensions/schema.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import * as v from "valibot";
+
+/** Parameters for POST /V1/adminuisdk/extension. */
+export const ExtensionRegistrationParamsSchema = v.object({
```
obarcelonap, line 16: According to the docs https://developer.adobe.com/commerce/extensibility/admin-ui-sdk/api#save-a-selected-extension props should be snake cased. Unless I am missing something I don't think this will work.

packages/aio-commerce-lib-admin-ui/source/api/extensions/endpoints.ts
```diff
@@ -0,0 +1,81 @@
...
+    UnregisterExtensionParamsSchema,
+    params,
+  );
+
+  return httpClient
+    .delete(
+      `adminuisdk/extension/${workspaceName}/${extensionName}`,
+      fetchOptions,
+    )
+    .then((_res) => {
```
obarcelonap, line 78: You can just await the deletion without this `then` block
```
await httpClient.delete(
  `adminuisdk/extension/${workspaceName}/${extensionName}`,
  fetchOptions,
);
```

obarcelonap (commented):
Overall LGTM, please review the publication config in package.json and the snake case in registering extension payload.

packages/aio-commerce-lib-admin-ui/source/api/extensions/endpoints.ts
iivvaannxx: @obarcelonap It depends, actually. For example, we did this with the event subscriptions endpoint, and for some reason, it always returned `[]` (not sure if it was a Commerce API thing specific to that endpoint). That's why I suggested to him to use `.then` without a return value to force `void` and ensure it even if the API changes.

packages/aio-commerce-lib-admin-ui/source/api/extensions/schema.ts
iivvaannxx: Fair enough 👍🏻. Resolving

packages/aio-commerce-lib-admin-ui/source/lib/api-client.ts
oshmyheliuk: SDK part was removed in this commit https://github.com/adobe/aio-commerce-sdk/pull/489/changes/b6e4effae73fb0e205dbeb4375c64d221fbf2a10

packages/aio-commerce-lib-admin-ui/source/api/extensions/schema.ts
oshmyheliuk: It was already in camelCase in the current implementation, and it seems to work in both formats for the AdminUI SDK API. But yes, probably it's better to change to the documentation format

obarcelonap: resolving if it works

packages/aio-commerce-lib-admin-ui/source/api/extensions/endpoints.ts
obarcelonap: I proposed to await without returning so you still get `Promise<Void>` as return type. In any case not a blocker.

obarcelonap (approved):

packages/aio-commerce-lib-admin-ui/package.json
oshmyheliuk: Added

packages/aio-commerce-lib-admin-ui/source/api/extensions/endpoints.ts
iivvaannxx: I see, I didn't understand. Works both ways, I don't have a strong opinion.

iivvaannxx:
> It was already in camelCase in the current implementation, and it seems to work in both formats for the AdminUI SDK API. But yes, probably it's better to change to the documentation format

So I just remembered that @capape explained a related issue to me just yesterday, and asked him what the reason might be. Turns out Commerce converts everything to `camelCase` before deserialization, that's why it seems to work anyway. You could say it's not a bug, but a feature 🤣

FYI @obarcelonap @oshmyheliuk 

@capape Correct me if I understood wrong 🙈

PR #494, opened 2026-06-04
CEXT-6269: spec for environment-scoped webhooks and events

oshmyheliuk:
## Description

Adds the feature spec specs/features/CEXT-6269-env-scoped-webhooks-events.md. No code changes — this is the spec PR; implementation follows separately.

  The spec proposes an optional env: ("paas" | "saas")[] array on webhook entries and individual events, reusing the exact shape and semantics of lib-config's existing field-level env scoping. Omitted = applies to
  all environments (fully backward compatible). At install time, items are filtered to the target Commerce environment; an event provider left with no applicable events gets neither a provider nor a registration.
  The app-config action returns the already-scoped config (via an optional commerceEnv query param), so App Management needs no filtering logic of its own.

https://jira.corp.adobe.com/browse/CEXT-6269

## Motivation and Context

  Commerce apps run against both PaaS and SaaS, but some webhooks/events only make sense on one. Today every item is installed for every environment, creating irrelevant providers, registrations, and subscriptions.
  Business Configurations already solved this with env; this extends the same pattern to webhooks and events for a consistent developer experience.

## Types of changes
  - [x] New feature (non-breaking change which adds functionality) — design proposal only

## Checklist:


specs/features/CEXT-6269-env-scoped-webhooks-events.md
```diff
@@ -0,0 +1,390 @@
...
+
+## Design
+
+### Shared `env` schema
+
+lib-config currently defines `COMMERCE_ENVS`, `CommerceEnvSchema`, and `EnvSchema`
+privately in
+`packages/aio-commerce-lib-config/source/modules/schema/fields.ts` and exports none of
+them. To keep a single source of truth across components, extract the env primitives into
+`@aio-commerce-sdk/common-utils` (`packages-private/common-utils/source/valibot/`), which is
```
iivvaannxx: Actually, the first time the CommerceEnv term appeared was in `lib-api` (albeit under `flavor` and not `env`). Both `lib-config` and `lib-app` depend on it. I think it would be a better extraction point. Or maybe `lib-core` as it's actually something quite "primitive".

This would mean exporting it from a public package, but to me, it makes sense. Don't have a strong opinion though.

specs/features/CEXT-6269-env-scoped-webhooks-events.md
```diff
@@ -0,0 +1,390 @@
...
+
+The target environment is resolved from the installation params by a dedicated,
+**non-throwing** helper added to lib-api — sketch: `resolveCommerceEnvironment(params):
+CommerceFlavor | null`. It returns the explicit `AIO_COMMERCE_API_FLAVOR` when valid,
+otherwise infers `"paas"` / `"saas"` from `AIO_COMMERCE_API_BASE_URL`, and returns `null` when
+neither is available. lib-api already has the internal pieces (`isFlavor`,
+`resolveCommerceFlavorFromApiUrl` in
+`packages/aio-commerce-lib-api/source/lib/commerce/helpers.ts`); the helper exposes them as a
+small public function, re-exported from lib-api's main entry (`source/index.ts`, alongside
+`resolveCommerceHttpClientParams`) — lib-api has no `./commerce` subpath export.
```
iivvaannxx: You sure about this? We receive the env from App Management UI we don't need to resolve anything I think.

specs/features/CEXT-6269-env-scoped-webhooks-events.md
```diff
@@ -0,0 +1,390 @@
...
+  that cannot resolve an environment fall back to "apply to all", which is a soft, possibly
+  surprising default.
+- Install filters by `env` but uninstall does not (see Uninstall), an asymmetry that must be
+  reasoned about carefully to avoid orphaned resources.
+
+## Rationale and alternatives
+
+- **`env` array vs. a single `environment` string.** The array matches lib-config exactly
+  and supports multi-environment scoping (`["paas", "saas"]`); a single string would diverge
+  from the established pattern and the goal of cross-component consistency. Rejected the single
```
iivvaannxx, line 301: If we're going to extract this into a Valibot schema, I think the correct approach would be a `[v.tuple](https://valibot.dev/api/tuple/)`, which is a fixed-array type with known shapes you want to accept. 

Not super important, but it would forbid things like `["paas", "paas", "saas"]`.

specs/features/CEXT-6269-env-scoped-webhooks-events.md
```diff
@@ -0,0 +1,390 @@
...
+  string.
+- **Per-event vs. per-provider scoping for events.** Per-event scoping plus a
+  skip-empty-provider rule naturally expresses "create no provider or registration when nothing
+  applies". Per-provider scoping is coarser and cannot express "this provider has both
+  PaaS-only and SaaS-only events". Chose per-event.
+- **Detecting the environment from auth strategy.** Rejected — PaaS can use IMS or
+  Integration auth, so auth strategy does not determine the environment. Only the API URL or
+  explicit environment param does.
```
iivvaannxx: The API URL should not be resolved either, we have it in params sent by UI don't we?

iivvaannxx (approved):
LGTM. Just some minor comments.

Side topic: There's no way to reliably tell if an event/webhook is actually compatible with the target instance right?

obarcelonap (approved):

oshmyheliuk:
>LGTM. Just some minor comments.
>
>Side topic: There's no way to reliably tell if an event/webhook is actually compatible with the target instance right?

In ACCS, we have allowed (compatible) lists, so it won't be possible to subscribe to an event/webhook that isn't on the list. But we don't have this for PaaS, as we allow the use of any event/webhook.

specs/features/CEXT-6269-env-scoped-webhooks-events.md
oshmyheliuk: Updated the spec to use `lib-core` for this.

specs/features/CEXT-6269-env-scoped-webhooks-events.md
oshmyheliuk: Agree, we are already passing all the required info in the installation action from App Management. So, not resolving is needed. Updated the specs.

specs/features/CEXT-6269-env-scoped-webhooks-events.md
oshmyheliuk: Spec was updated

specs/features/CEXT-6269-env-scoped-webhooks-events.md
oshmyheliuk: Looks like a `tuple` has a fixed length, so it won't work with cases like — ["paas"], ["saas"], or ["paas","saas"]

iivvaannxx: mmm yes, was thinking of having a union of 4 tuples, `["paas"]`, `["saas"]`, `["paas", "saas"]` or `["saas", "paas"]`. 

Feel free to skip this comment if you find this too strict. For me, it's the one ensuring the most correctness, though. Maybe a simple manual `v.check` can do it too. Do as you prefer 👍🏻

PR #504, opened 2026-06-08
add per-package BREAKING.md and deprecation tracking rule

obarcelonap:
## Description

Adds a `BREAKING.md` to each package that has deprecated symbols or planned breaking changes, and adds a rule to `AGENTS.md` requiring it to be kept in sync whenever a `@deprecated` tag is added or a breaking change is deferred.

## Related Issue

N/A

## Motivation and Context

We needed a place to accumulate deprecated symbols and planned breaking changes per package so consumers know what to prepare for ahead of a future major release.

## How Has This Been Tested?

Documentation-only change — no functional code modified.

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/BREAKING.md
```diff
@@ -0,0 +1,9 @@
+# Breaking Changes
+
+## [Unreleased]
+
+### Deprecated
+
+- `generateInstanceIdDeprecated()` will be removed in a future major — use `generateInstanceId()` instead. The old function could produce non-unique IDs within the same org.
+- `getLegacyRegistrationName()` will be removed in a future major — use `getRegistrationName()` instead. The legacy format omits the provider label from the registration name and is only kept to match registrations created before the naming change.
```
iivvaannxx: AFAIK, these two are not public. They are used only to ensure legacy apps can still be "referenced" or "found" with older generated ID/name formats. It should not be information targeted to consumers (who cannot import them anyway).

I would model those as `TODO: Remove in next major` or something like that in the code itself.

packages/aio-commerce-lib-api/BREAKING.md
```diff
@@ -0,0 +1,23 @@
...
+
+// ky v2
+beforeRequest: [({ request, options, retryCount }) => { ... }]
+```
+
+The same change applies to `afterResponse`, `beforeError`, and `beforeRetry`.
+
+It will be part of the next major release together with other accumulated breaking changes.
+
+**What users will need to update:** any call site passing `hooks` to an SDK HTTP client — destructure the state argument instead of receiving each value as a separate parameter.
```
iivvaannxx: As it's not a change of ours, I would just add a link to the migration guide of the package.

packages/aio-commerce-lib-api/BREAKING.md
```diff
@@ -0,0 +1,23 @@
...
+// ky v1 (current SDK public API)
+beforeRequest: [(request, options, retryCount) => { ... }]
+
+// ky v2
+beforeRequest: [({ request, options, retryCount }) => { ... }]
+```
+
+The same change applies to `afterResponse`, `beforeError`, and `beforeRetry`.
+
+It will be part of the next major release together with other accumulated breaking changes.
```
iivvaannxx: This should be global information at the top of the file as it applies to any change in the BREAKING.md

iivvaannxx (approved):

PR #510, opened 2026-06-09
CEXT-6269: Support Environment-Scoped Creation for Webhooks and Events Aligned with Business Configurations UX

oshmyheliuk:
## Description

  Adds an optional env: ["paas" | "saas"] field to webhook and event config entries so they can be scoped to specific Commerce environments. Scoped items are only installed on matching environments (an event
  provider left with no applicable events is skipped entirely), and the app-config action returns business-config fields, webhooks, and events already filtered to the requested commerceEnv.

  The Commerce-environment schema is consolidated into a single source of truth: a new @adobe/aio-commerce-lib-core/commerce export (COMMERCE_ENVS, CommerceEnv, commerceEnvSchema, commerceEnvArraySchema),
  re-exported via @adobe/aio-commerce-sdk/core/commerce. lib-config, lib-api (CommerceFlavor now aliases CommerceEnv), and the lib-app actions all consume it. The installation and scope-tree actions now validate
  commerceEnv against "paas" | "saas" instead of accepting any string.

https://jira.corp.adobe.com/browse/CEXT-6269

## Related PR

https://github.com/adobe/aio-commerce-sdk/pull/494

https://github.com/OneAdobe/commerce-app-management/pull/226

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have read the **DEVELOPMENT** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #511, opened 2026-06-09
CEXT-6160: store commerce instance on association and expose retrieval helpers

vinayrao2000:
## Description

Adds a new feature that stores the Commerce instance details (Base URL + deployment type) when an app is associated with a Commerce instance, and exposes typed async helpers so any runtime action can call Commerce APIs without custom storage setup.

**Two new helpers in `@adobe/aio-commerce-lib-app`:**

| Helper | Returns | Use when |
|---|---|---|
| `getCommerceInstance()` | `{ baseUrl, env }` | You need the raw stored values (URL/env) |
| `getCommerceClient(auth, fetchOptions?)` | `AdobeCommerceHttpClient` (ready to call APIs) | You want to call Commerce APIs directly |

`getCommerceInstance()` takes no arguments — it reads directly from storage. `getCommerceClient(auth, fetchOptions?)` is composable: it accepts IMS auth already resolved by the caller (via `resolveImsAuthParams` from `@adobe/aio-commerce-lib-auth`, or an `ImsAuthProvider`) and combines it with the stored base URL and deployment type. Optional `fetchOptions` are forwarded to the underlying client (e.g. `headers`, `timeout`, `retry`). Both throw `AppNotAssociatedError` if the app isn't currently associated.

**Storage** (generic system-config primitives in `@adobe/aio-commerce-lib-config`):

- Two-layer pattern: state (cache) → files (persistent source of truth)
- Generic `getSystemConfigByKey` / `setSystemConfigByKey` API — reusable for other "system-level" config beyond association
- Exported from the package root (no dedicated subpath). Rather than reimplementing the two-layer cache, they reuse `configuration-repository`'s `loadConfig` / `persistConfig` / `deleteConfig`, parameterized by a separate `system.*` storage namespace
- Key namespace: `system.association` for this feature, stored under `system/{key}.json`, kept distinct from Business Configuration's `configuration.*` / `scope/` namespace
- Association data is cached at the maximum TTL (one year, `MAX_TTL` from `@adobe/aio-lib-state`) instead of the 24h default — files remain the source of truth and association data changes rarely, so reads stay served from the cache

**New runtime action** auto-scaffolded into apps:

- `POST /association` — store `{ commerceBaseUrl, commerceEnv }` (called by App Management UI on associate)
- `DELETE /association` — clear stored data (called on unassociate)

## Related Issue

CEXT-6160

## Motivation and Context

Before this change, every action that wanted to call Commerce APIs had to:

1. Read auth params from `params`
2. Read the base URL from env vars or a custom-managed config
3. Manually construct `AdobeCommerceHttpClient`
4. Handle missing config in every action

This pushed boilerplate and inconsistent storage patterns into every app. Customers had no standard way to know which Commerce instance their app was associated with at runtime.

After this change, an action resolves auth once and gets a configured client:

```js
const client = await getCommerceClient(resolveImsAuthParams(params));
```

The SDK owns where association data is stored (state + files), how it's read, and how it propagates from the association event to runtime actions.

## How Has This Been Tested?

**Unit tests** (~1,865 tests passing across the monorepo):

- `lib-config`: `system-config.test.ts` covers set/get/clear, state-as-cache + files-as-source-of-truth, null/undefined clears, corrupt-JSON fallback, type safety
- `lib-app`:
  - `errors/app-not-associated-error.test.ts` — class, message, options
  - `management/association/association-repository.test.ts` — round-trip via system config (against the storage layer)
  - `actions/association.test.ts` — router (POST/DELETE), validation, logger middleware
  - `access/commerce-instance.test.ts` — `getCommerceInstance` and `getCommerceClient` happy + error paths

**Typecheck:** clean across all packages (`pnpm typecheck`)

**Lint:** clean (`pnpm check:ci`)

**E2E** (in `commerce-app-purchase-approval` demo app):

- Associate via App Management UI → `POST .../association` → 200 OK
- Invoke `commerce-config-demo` action → `{ associated: true, instance: { baseUrl, env }, clientType: "AdobeCommerceHttpClient" }`
- Unassociate via App Management UI → `DELETE .../association` → 204 No Content
- Invoke demo action again → `{ associated: false, ... }`
- Re-associate after unassociate → flow works cleanly

Both helpers verified working end-to-end through deployed actions.

## Types of changes

## Checklist:

## Changesets

- `aio-commerce-lib-app` — minor (new helpers `getCommerceInstance` / `getCommerceClient` + `AppNotAssociatedError` + association action; additive only)
- `aio-commerce-lib-config` — minor (new root exports `getSystemConfigByKey` / `setSystemConfigByKey`; additive only)

## Affected packages

| Package | Change type |
|---|---|
| `@adobe/aio-commerce-lib-app` | new public API: `getCommerceInstance`, `getCommerceClient`, `AppNotAssociatedError`, association runtime action + scaffolding template |
| `@adobe/aio-commerce-lib-config` | new public API: root exports `getSystemConfigByKey` / `setSystemConfigByKey` |


packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -7,8 +7,12 @@ The `@adobe/aio-commerce-lib-app` library provides:
...
 - **Business Configuration**: Generate and manage the runtime actions that power the `commerce/configuration/1` extension point.
 - **Installation Management**: Generate and manage the runtime action that powers the app installation flow.
+  <<<<<<< HEAD
+- **Admin UI SDK Configuration**: Generate and manage the runtime action that powers the `commerce/backend-ui/1` extension point.
+- # **Association Helpers**: Retrieve the Commerce instance the app is associated with from any runtime action via `getCommerceClient` and `getCommerceInstance`.
 - **Admin UI Configuration** (`commerce/backend-ui/2`): Generate and manage the runtime action and `workerProcess` declarations for Admin UI extensions on `commerce/backend-ui/2`. Currently supports grid column extensions; mass actions, menus, view buttons, and custom fees will follow.
 - **Admin UI SDK Configuration** (`commerce/backend-ui/1`, _deprecated_): Generate and manage the runtime action for the legacy Admin UI SDK extension point. Will be removed from the SDK — use `adminUi` and `commerce/backend-ui/2` for new apps.
+  > > > > > > > 226e62e8602076e7449ad22b2eb74262c31b0872
```
oshmyheliuk: @vinayrao2000, you have committed an unresolved conflict, please merge the changes

packages/aio-commerce-lib-app/source/actions/association/schema.ts
```diff
@@ -0,0 +1,19 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import * as v from "valibot";
+
+/** Request body for POST / — store association data. */
+export const AssociationRequestBodySchema = v.object({
+  commerceBaseUrl: v.string(),
```
oshmyheliuk: Url validation can be added here

packages/aio-commerce-lib-app/docs/usage.md
vinayrao2000: Thanks @oshmyheliuk, fixed in f54668c

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
```diff
@@ -0,0 +1,147 @@
...
+  const logger = getLogger("@adobe/aio-commerce-lib-config:system-repository");
+  try {
+    const state = await getSharedState();
+    await state.delete(key);
+  } catch (error) {
+    logger.debug(
+      "Failed to clear system config cache:",
+      error instanceof Error ? error.message : String(error),
+    );
+  }
```
oshmyheliuk: Functions description should be added

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
```diff
@@ -0,0 +1,147 @@
...
+    const files = await getSharedFiles();
+    const filePath = getSystemFilePath(key);
+    const filesList = await files.list(SYSTEM_FILES_PREFIX);
+    const fileObject = filesList.find((file) => file.name === filePath);
+
+    if (!fileObject) {
+      return null;
+    }
+
+    const content = await files.read(filePath);
```
oshmyheliuk: Probably to improve performance, `files.read(filePath);` can be called directly without fetching the list of files (it looks like a heavy operation).
In case of error, the null will be returned in the catch block.

packages/aio-commerce-lib-app/source/actions/association/schema.ts
vinayrao2000: Added URL validation, commerceBaseUrl now runs through v.pipe(v.string(), v.url(...)), so a malformed URL is rejected with a 400 before we store anything. Covered it with a test for the invalid-URL case and updated openapi.json to match (format: uri on the field, plus a version bump).

Fixed in 6ab289f996d40304be6ba201b2f78e0c71fe0a8b

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
vinayrao2000: Documented all the internal helpers with JSDoc, including their error contracts: cache failures are swallowed since files is authoritative, and files write errors propagate.

Fixed in 305860ad0ec46868b148d94fca8acb773831e1bf

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
vinayrao2000: Fixed, Removed the files.list pre-check so we now call files.read(filePath) straight away, relying on the catch block to return null for a missing file. This mirrors what scope-tree-repository does. I also reworked the affected tests to mock files.read directly and stripped out the dead files.list mocks.

Fixed in 305860ad0ec46868b148d94fca8acb773831e1bf

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Writes a payload to the `aio-lib-state` cache. Cache failures are swallowed
+ * (logged at debug) since `aio-lib-files` remains the source of truth.
+ */
+async function writeToCache(key: string, payload: string): Promise<void> {
+  const logger = getLogger("@adobe/aio-commerce-lib-config:system-repository");
+  try {
+    const state = await getSharedState();
+    await state.put(key, payload);
```
oshmyheliuk: Shouldn't we add a TTL here? As I know a default value is 24 hours

packages/aio-commerce-lib-app/docs/openapi.json
```diff
@@ -2668,6 +2668,122 @@
...
+                  "properties": {
+                    "baseUrl": {
+                      "type": "string"
+                    },
+                    "env": {
+                      "type": "string",
+                      "enum": ["saas", "paas"]
+                    }
+                  },
+                  "required": ["baseUrl", "env"]
```
asalloum5: Is there a reason why the response body is different from the request body? I expected to see `commerceBaseUrl` or `baseUrl` in both.

asalloum5 (approved):

packages/aio-commerce-lib-app/docs/openapi.json
```diff
@@ -3,7 +3,7 @@
   "info": {
     "title": "App Management API",
     "description": "REST API for managing Adobe Commerce App Builder apps: resolving app configuration, managing configuration values, running the installation and uninstallation lifecycle, exposing the Admin UI SDK registration, and managing the scope tree.",
-    "version": "1.1.0",
+    "version": "1.1.1",
```
iivvaannxx: We already have one major update queued in `main`. No need to bump this.

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
iivvaannxx, line 1: `system-repository.ts` reimplements the two-layer cache pattern from `configuration-repository.ts`: read from state, fall back to files, re-cache, swallow cache errors. ~150 lines plus ~164 lines of tests for logic that's already there.

The only real difference between system config and regular config is that system config doesn't participate in scope-tree inheritance. But that inheritance logic lives above the repository layer, not within it, so there's nothing in `configuration-repository.ts` that would bleed into it. A `deleteConfig` (needed here regardless) plus calling `loadConfig/persistConfig/deleteConfig` with a fixed key would've been enough. `getSystemConfigByKey/setSystemConfigByKey` could then just live on the root entrypoint, which also eliminates the need for the new `./system` subpath, the barrel file, and the extra `tsdown` entry.

packages/aio-commerce-lib-app/docs/openapi.json
obarcelonap: should be minor instead of patch

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -781,6 +782,106 @@ try {
...
+```ts
+import {
+  AppNotAssociatedError,
+  getCommerceClient,
+} from "@adobe/aio-commerce-lib-app";
+
+export async function main(params) {
+  try {
+    const client = await getCommerceClient(params);
+    return {
```
obarcelonap: use preset `ok()` from the sdk

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -781,6 +782,106 @@ try {
...
+      statusCode: 200,
+      body: await client.get("rest/V1/products").json(),
+    };
+  } catch (error) {
+    if (error instanceof AppNotAssociatedError) {
+      return {
```
obarcelonap: use preset `badRequest` from the sdk

packages/aio-commerce-lib-app/source/actions/association/schema.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/** Request body for POST / — store association data. */
+export const AssociationRequestBodySchema = v.object({
+  commerceBaseUrl: v.pipe(
+    v.string(),
+    v.url(
+      "The 'commerceBaseUrl' field must be a valid absolute URL (e.g., 'https://my-store.example.com')",
+    ),
+  ),
+  commerceEnv: v.picklist(["saas", "paas"]),
```
obarcelonap: didn't we used `flavour` in other parts of the sdk? we should align this

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
obarcelonap, line 1: can you check if `createCombinedStore` can be used instead of implementing your own cache

packages/aio-commerce-lib-app/docs/openapi.json
```diff
@@ -2668,6 +2668,122 @@
...
+            "description": "Bad request, the request body is invalid.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ErrorResponse"
+                }
+              }
+            }
+          },
+          "500": {
```
iivvaannxx, line 2653: Missing `401` and `403` responses. Every other route in this file documents both — `401` for an invalid/missing IMS token and `403` for insufficient permissions. The action is protected with `require-adobe-auth: true` so both are reachable. Same applies to `DELETE /`.

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -781,6 +782,106 @@ try {
 }
 ```
 
+### Accessing the Associated Commerce Instance from Runtime Actions
+
+After an app is associated with a Commerce instance via App Management, the SDK stores the
```
iivvaannxx: The new section hard-wraps prose at ~90 chars with explicit line breaks mid-paragraph. The rest of the file flows as single-line paragraphs.

packages/aio-commerce-lib-app/source/errors/app-not-associated-error.ts
```diff
@@ -0,0 +1,52 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { CommerceSdkErrorBase } from "@adobe/aio-commerce-lib-core/error";
+
+const DEFAULT_MESSAGE =
+  "App is not associated with a Commerce instance. Re-associate the app to resolve this error.";
+
+type AppNotAssociatedErrorOptions = ErrorOptions & {
```
iivvaannxx: `AppNotAssociatedErrorOptions` is `ErrorOptions & { traceId?: string }` — identical to `CommerceSdkErrorBaseOptions` from `lib-core`. Just use that directly instead of re-declaring it.

packages/aio-commerce-lib-app/source/management/association/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+// biome-ignore lint/performance/noBarrelFile: export as part of the Public API
+export {
```
iivvaannxx, line 14: This lives under `source/modules/association/` but there's already `source/management/` with `installation/` inside it. Association is a management lifecycle concern (triggered by App Management on associate/unassociate), so `management/association/` would be consistent.

packages/aio-commerce-lib-app/source/index.ts
```diff
@@ -0,0 +1,102 @@
...
+ * export async function main(params) {
+ *   const instance = await getCommerceInstance(params);
+ *
+ *   // instance.baseUrl — e.g. "https://my-store.example.com"
+ *   // instance.env     — "saas" | "paas"
+ * }
+ * ```
+ */
+export async function getCommerceInstance(
+  _params: RuntimeActionParams,
```
iivvaannxx: `_params` is accepted but never used — `getAssociationData()` reads from storage without needing credentials from params. If the intent is API consistency with `getCommerceClient`, the signature makes sense. If not, the parameter shouldn't be there.

packages/aio-commerce-lib-app/source/index.ts
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * export async function main(params) {
+ *   const instance = await getCommerceInstance(params);
+ *
+ *   // instance.baseUrl — e.g. "https://my-store.example.com"
+ *   // instance.env     — "saas" | "paas"
+ * }
+ * ```
+ */
+export async function getCommerceInstance(
```
iivvaannxx: `getCommerceInstance` and `getCommerceClient` implementations are in `index.ts`. Across the monorepo `index.ts` files are pure barrels — these helpers should live in their own file and index should just re-export.

packages/aio-commerce-lib-app/test/unit/actions/association.test.ts
```diff
@@ -0,0 +1,161 @@
...
+      const action = associationRuntimeAction();
+      const params = createRuntimeActionParams({
+        method: "delete",
+        path: "/",
+      });
+
+      const result = await action(params);
+
+      expect(mockClearAssociationData).toHaveBeenCalledOnce();
+      expect(result).toMatchObject({
```
iivvaannxx, line 177: No test for when `setAssociationData` or `clearAssociationData` throws. The mocks are never made to reject, so it's untested whether the router returns a 500 or surfaces an unhandled rejection when the storage write fails.

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
vinayrao2000: Good question. This is intentional, state.put without a ttl applies the service default of 24h, which is the behavior we want for the cache layer. Per the storage design in the spec, lib-state is a TTL-backed performance cache sitting on top of lib-files as the no-TTL source of truth: when a cached entry expires, the read path falls back to files and re-caches transparently, so no data is ever lost on expiry. Since the system config changes only on associate/unassociate, the occasional post-expiry re-read from files is negligible.

Happy to make the 24h explicit ({ ttl: 86400 }) if you'd prefer it documented in code rather than relying on the service default, it's a no-op behaviorally, but the current implementation already matches the intended caching design.

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
vinayrao2000: based on the discussion I have implemented this way, as you mention i will review createCombinedStore

vinayrao2000: Agreed the two-layer mechanics are duplicated, and the scope-inheritance that differs lives above the repo layer so it wouldn't bleed in. But reusing configuration-repository's loadConfig/persistConfig drags in its quirks, the stringify({ data }) double-wrap, the files.list pre-check I removed from system-repository earlier in this review, and the configuration.* namespace the spec deliberately keeps separate from system.*. So I don't think coupling to it is the right de-dupe. Same root concern @obarcelonap raised about `createCombinedStore`; the real fix that satisfies both is moving both repos onto one shared store bigger change touching shared infra + an existing consumer,

packages/aio-commerce-lib-app/docs/openapi.json
vinayrao2000: There's no real reason for this. It wasn't intentional, the response was echoing the internal stored shape (baseUrl/env) instead of mirroring the request. Do you want me to change them so the endpoint uses commerceBaseUrl/commerceEnv on both the request and the response.

packages/aio-commerce-lib-app/docs/usage.md
vinayrao2000: Done, updated the example to return ok({ body: ... }) (and badRequest(...) for the unassociated case) instead of a hand-built response object.

Fixed in 90f24916c83093843e8aec867de6505945b294b1

packages/aio-commerce-lib-app/docs/usage.md
vinayrao2000: Done, updated the example to return ok({ body: ... }) (and badRequest(...) for the unassociated case) instead of a hand-built response object.

Fixed in 90f24916c83093843e8aec867de6505945b294b1

packages/aio-commerce-lib-app/docs/openapi.json
vinayrao2000: Added 401 and 403 to both POST and DELETE /association, matching the wording used by the other routes. The action is require-adobe-auth: true, so both are reachable as you noted.

Fixed in 48470868d2fcfefc82b664724b5d2a4075d72011

packages/aio-commerce-lib-app/source/management/association/index.ts
vinayrao2000: Good point on the single-occupant modules/ folder. My only hesitation: management/ backs the ./management subpath of install-time tooling, while association's public API (getCommerceInstance/getCommerceClient) is general-purpose runtime access exported from root, so I grouped it as a "module" rather than management tooling. The store/clear lifecycle part already lives in actions/association/. 

 Please suggest, happy to move it to management/association/ if you think it fits better there.

packages/aio-commerce-lib-app/source/index.ts
vinayrao2000: It's unused today, agreed. The intent is API symmetry: getCommerceInstance and getCommerceClient share the same (params) shape so callers can swap between "give me the instance" and "give me a ready client" without changing the call. getCommerceClient needs params for auth; getCommerceInstance takes it only for that symmetry, hence the _ prefix. 

Please suggest you still want to remove _param

packages/aio-commerce-lib-app/source/index.ts
vinayrao2000: Done. Moved both helpers into modules/association/commerce-instance.ts (alongside getAssociationData and the types they build on) and reduced index.ts to a pure re-export, matching the barrel convention used by the other libs. Public import path is unchanged.

Fixed in fb2dfc47fea42c757d59c77e0fb6bf8ba71dd6ac

packages/aio-commerce-lib-app/test/unit/actions/association.test.ts
vinayrao2000: Added. Both handlers now have a test that rejects the storage mock (setAssociationData / clearAssociationData) and asserts the action resolves with a 500, confirming the router's catch surfaces an internalServerError rather than leaking an unhandled rejection.

Fixed in 80a75267bf44b7b6f2e204bafdbd09f49d1f379f

packages/aio-commerce-lib-app/source/errors/app-not-associated-error.ts
vinayrao2000: Done, removed the local AppNotAssociatedErrorOptions and use CommerceSdkErrorBaseOptions directly.

Fixed in 8b61c1e23a9f716e35829bb6c4ca625e5246ac6d

packages/aio-commerce-lib-app/docs/usage.md
vinayrao2000: Done, unwrapped the section's paragraphs (and the two helper bullets) to single-line, matching the flow of the rest of the file.

Fixed in 90f24916c83093843e8aec867de6505945b294b1

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
iivvaannxx: If the logic is the same, then whatever can be fixed in one can be fixed in the other. If your code optimizes/fixes the pre-check of `files.list` we can also apply it in the configuration repository and still re-use the logic, while also optimizing other consumers of the config repository at the same time (double win). Not sure what "double-wrap of stringify" means, but pretty sure it can be fixed instead of just duplicated.

>  and the configuration.* namespace the spec deliberately keeps separate from system.*. So I don't think coupling to it is the right de-dupe.

I am not talking about re-using the same namespace; you can still just use different namespaces (just decouple the constant namespace from the repository). It's just about reusing logic that is already implemented. Adding/duplicating all that code for me doesn't make much sense and adds a lot of maintenance overhead that can be avoided.

The concern raised by @obarcelonap about `createCombinedStore` is correct, but I don't think this belongs in this PR. It's different because you're adding unnecessary logic, while his concern would require modifying working code, which is better done separately from this task.

If your concern is about semantics (using the configuration repository for "system" config), just rename things, but duplicating for the sake of duplicating, I don't think it's the solution.

packages/aio-commerce-lib-app/source/errors/app-not-associated-error.ts
iivvaannxx: I would still keep the custom error options for better future extensibility and avoiding type contract changes. But you only need to keep it empty, not redeclare `traceId`.

packages/aio-commerce-lib-app/source/management/association/index.ts
iivvaannxx: Valid concern, what I would argue then is that maybe then `association` is not a correct name at all, because `association` is indeed part of the `management` lifecycle.

packages/aio-commerce-lib-app/source/index.ts
iivvaannxx: It should be removed; we don't do this kind of symmetry anywhere in the libraries. If a param is not used, it doesn't make sense to me to ask for it. What I would say is that maybe `getCommerceClient` should not take `params` at all. Because then it's doing multiple things, it's resolving `params` into the client options and also creating the client.

Helpers like this should be composable. Maybe resolving should happen outside, and the resolved params should be used as input. Or if the `getCommerceClient` only needs the auth part to be resolved because the url and the flavor are given by `getCommerceInstance` then that should be what's resolved outside (and properly documented)

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
iivvaannxx: @obarcelonap For this use case the `configuration-repository` can just be re-used. Let's integrate the combined store in a separate PR. See [this comment](https://github.com/adobe/aio-commerce-sdk/pull/511#discussion_r3412549864)

vinayrao2000: Makes sense, happy to go this route. Just flagging it deviates from the spec (which describes a separate system/ submodule + ./system subpath), so I'd update the spec to match.

Plan: drop the system/ submodule, ./system subpath, barrel, and tsdown entry. getSystemConfigByKey/setSystemConfigByKey would build on loadConfig/persistConfig + a new deleteConfig (parameterized by a system.* namespace) and export from the package root. Behavior stays the same.

packages/aio-commerce-lib-app/source/management/association/index.ts
vinayrao2000: Moved the public helpers to source/access/; association storage stays in modules/association/ (used by the action). 

Fixed in b2409d94ca0d2593b492571afe13c027dcb35c9c

packages/aio-commerce-lib-app/source/index.ts
vinayrao2000: getCommerceInstance() takes no args now, and getCommerceClient(auth) accepts auth resolved via resolveAuthParams. 

Fixed in bf311449457df68cbf0c4a45c0526794785bd2de

packages/aio-commerce-lib-app/source/errors/app-not-associated-error.ts
vinayrao2000: Kept as a type alias (empty interface extends trips Biome's noEmptyInterface), a named seam for future options. 

Fixed in 04b65590087b0d2f4792ffdbfbb664f83ed6ae43

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
vinayrao2000: Removed it; system config now reuses configuration-repository via a RepositoryNamespace. 

Fixed in de6e29f9886c5533324904b02c9b11d15aa7cc33

packages/aio-commerce-lib-config/source/modules/configuration/configuration-repository.ts
```diff
@@ -43,41 +82,74 @@ export async function getCachedConfig(scopeCode: string) {
...
+  );
+  try {
+    const state = await getSharedState();
+    await state.delete(namespace.stateKey(scopeCode));
+  } catch (error) {
+    logger.debug(
+      "Failed to clear cached configuration:",
       error instanceof Error ? error.message : String(error),
     );
-    // Don't throw - caching failure shouldn't break functionality
```
iivvaannxx, line 66: I think this comment is useful and shouldn't be removed.

packages/aio-commerce-lib-config/source/modules/configuration/configuration-repository.ts
```diff
@@ -43,41 +82,74 @@ export async function getCachedConfig(scopeCode: string) {
...
+export async function getPersistedConfig(
+  scopeCode: string,
+  namespace: RepositoryNamespace = CONFIGURATION_NAMESPACE,
+) {
   try {
     const files = await getSharedFiles();
-    const filePath = getConfigFilePath(scopeCode);
-    const filesList = await files.list("scope/");
+    const filePath = namespace.filePath(scopeCode);
+    const filesList = await files.list(namespace.filesPrefix);
```
iivvaannxx: Didn't we say we could optimize this a bit by removing the `list` call?

packages/aio-commerce-lib-config/source/modules/configuration/configuration-repository.ts
```diff
@@ -96,36 +168,68 @@ export async function getPersistedConfig(scopeCode: string) {
...
+async function deletePersistedConfig(
+  scopeCode: string,
+  namespace: RepositoryNamespace = CONFIGURATION_NAMESPACE,
+) {
+  const logger = getLogger(
+    "@adobe/aio-commerce-lib-config:configuration-repository",
+  );
+  try {
+    const files = await getSharedFiles();
+    await files.delete(namespace.filePath(scopeCode));
```
iivvaannxx: This will swallow any error when deleting a file that I believe should bubble up.

packages/aio-commerce-lib-config/source/modules/configuration/system-config.ts
```diff
@@ -0,0 +1,68 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import {
+  deleteConfig,
+  loadConfig,
+  persistConfig,
+  SYSTEM_NAMESPACE,
```
iivvaannxx: Probably the SYSTEM_NAMESPACE should live in this file and not in the other.

packages/aio-commerce-lib-config/source/modules/configuration/system-config.ts
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Retrieves a system configuration value by key.
+ *
+ * Reads from the `aio-lib-state` cache first, falling back to `aio-lib-files`
+ * and re-caching. Returns `null` when the key is not found in either layer.
+ *
+ * @param key - The system configuration key (e.g. `"system.association"`).
+ * @returns The stored value cast to `T`, or `null` if not found.
+ */
+export async function getSystemConfigByKey<T>(key: string): Promise<T | null> {
```
iivvaannxx: I would add a default to avoid having to force-type this call. Also, could you add it to the `getConfigByKey` too?


```suggestion
export async function getSystemConfigByKey<T = unknown>(key: string): Promise<T | null> {
```

packages/aio-commerce-lib-config/source/modules/configuration/system-config.ts
```diff
@@ -0,0 +1,68 @@
...
+  return (await loadConfig(
+    key,
+    SYSTEM_CONFIG_CACHE_TTL_SECONDS,
+    SYSTEM_NAMESPACE,
+  )) as T | null;
```
iivvaannxx: This doesn't sound correct to me? Not sure now, but I believe `loadConfig` returns raw JSON payloads that are not primitives you can just cast with `as`. That's why we have the specific implementation to load things with keys in [get-config-by-key](https://github.com/adobe/aio-commerce-sdk/blob/4aa1c3cec65b36137d0990d54d639836c16cbaec/packages/aio-commerce-lib-config/source/modules/configuration/get-config-by-key.ts#L35). I might be wrong though.

packages/aio-commerce-lib-config/test/unit/modules/configuration/system-config.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    put: vi.fn(),
+    delete: vi.fn(),
+  },
+  mockFiles: {
+    list: vi.fn(),
+    read: vi.fn(),
+    write: vi.fn(),
+    delete: vi.fn(),
+  },
+}));
```
iivvaannxx: You already have mocking infrastructure for the storage layer in `tests/mocks`

packages/aio-commerce-lib-core/source/error/index.ts
```diff
@@ -16,5 +16,8 @@
  */
 
 // biome-ignore lint/performance/noBarrelFile: export as part of the Public API
-export { CommerceSdkErrorBase } from "./base-error";
+export {
+  CommerceSdkErrorBase,
+  type CommerceSdkErrorBaseOptions,
```
iivvaannxx: This is going to need a Changeset in lib-core.

.changeset/commerce-instance-helpers.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": minor
+---
+
+Add helpers to retrieve the Commerce instance an app is associated with from any runtime action: `getCommerceInstance()` returns the stored instance data, and `getCommerceClient(auth)` returns a ready-to-use Commerce HTTP client built from that instance. Pass auth resolved with `resolveAuthParams` from `@adobe/aio-commerce-lib-auth`; the base URL and deployment type come from the stored association.
```
iivvaannxx: This is an unnecessary implementation detail that belongs to documentation instead of the Changeset.

```suggestion
Add helpers to retrieve the Commerce instance an app is associated with from any runtime action: `getCommerceInstance()` returns the stored instance data, and `getCommerceClient(auth)` returns a ready-to-use Commerce HTTP client built from that instance.
```

packages/aio-commerce-lib-app/docs/openapi.json
iivvaannxx: For me it doesn't make much sense to echo back the request. This endpoint should not return anything. A simple status code with an empty body is enough.

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -781,6 +782,84 @@ try {
 }
 ```
 
+### Accessing the Associated Commerce Instance from Runtime Actions
+
+After an app is associated with a Commerce instance via App Management, the SDK stores the Commerce base URL and deployment type (`saas` or `paas`) so any runtime action can retrieve them — without custom storage setup or threading parameters through every layer of the call stack.
+
+Two helpers are exposed from the root entrypoint:
+
+- `getCommerceClient(auth)` — returns a ready-to-use [`AdobeCommerceHttpClient`](../../aio-commerce-lib-api/docs/usage.md). Use this when you need to call the Commerce API. The base URL and flavor come from the stored association data; you supply the resolved auth credentials (resolve them with `resolveAuthParams` from [`@adobe/aio-commerce-lib-auth`](../../aio-commerce-lib-auth/docs/usage.md)).
```
iivvaannxx: Given that with App Management you are essentially required to use IMS, I would only accept `ImsAuthParams` or `ImsAuthProvider`, and the given authentication parameters should come from [`resolveImsAuthParams`](https://github.com/adobe/aio-commerce-sdk/blob/4aa1c3cec65b36137d0990d54d639836c16cbaec/packages/aio-commerce-lib-auth/source/lib/ims-auth/utils.ts#L131) (and not the generic `resolveAuthParams`), or from any other helper that creates a provider like the `forwardImsProvider` or the `getImsAuthProvider` from the resolved params.

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -781,6 +782,84 @@ try {
...
+
+#### Primary pattern — get a ready-to-use client
+
+```ts
+import { getCommerceClient } from "@adobe/aio-commerce-lib-app";
+import { resolveAuthParams } from "@adobe/aio-commerce-lib-auth";
+
+export async function main(params) {
+  const client = await getCommerceClient(resolveAuthParams(params));
+  const products = await client.get("rest/V1/products").json();
```
iivvaannxx: The `rest` prefix or versioning is not needed when using `AdobeCommerceHttpClient`

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -781,6 +782,84 @@ try {
...
+Both helpers throw `AppNotAssociatedError` if the app is not currently associated, was unassociated, or was associated by an older SDK that did not store this data. Re-associating the app resolves the error.
+
+#### Primary pattern — get a ready-to-use client
+
+```ts
+import { getCommerceClient } from "@adobe/aio-commerce-lib-app";
+import { resolveAuthParams } from "@adobe/aio-commerce-lib-auth";
+
+export async function main(params) {
+  const client = await getCommerceClient(resolveAuthParams(params));
```
iivvaannxx: There should be an additional optional argument to specify some global client `fetch` options like the constructor of the underlying `AdobeCommerceHttpClient` has.

packages/aio-commerce-lib-app/source/access/commerce-instance.ts
```diff
@@ -0,0 +1,86 @@
...
+ * import { resolveAuthParams } from "@adobe/aio-commerce-lib-auth";
+ *
+ * export async function main(params) {
+ *   const client = await getCommerceClient(resolveAuthParams(params));
+ *   const products = await client.get("rest/V1/products").json();
+ * }
+ * ```
+ */
+export async function getCommerceClient(
+  auth: CommerceHttpClientParams["auth"],
```
iivvaannxx: As in the comment above, this should be `ImsAuth` only. We can always augment if needed in the future, but unlikely as App Management requires IMS.

packages/aio-commerce-lib-app/source/errors/app-not-associated-error.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import type { CommerceSdkErrorBaseOptions } from "@adobe/aio-commerce-lib-core/error";
+
+const DEFAULT_MESSAGE =
+  "App is not associated with a Commerce instance. Re-associate the app to resolve this error.";
+
+/**
+ * Options accepted by {@link AppNotAssociatedError}. Aliases
+ * {@link CommerceSdkErrorBaseOptions} for now, kept as a named seam so future
+ * error-specific options can be added without changing the constructor contract.
```
iivvaannxx: This is a doc-comment that ends up in public documentation while including information about an internal technical implementation decision. They should be separated.

packages/aio-commerce-lib-app/source/management/association/association-repository.ts
iivvaannxx, line 1: Now, this should go in `management/association`. Not exported from the `/management` entrypoint tho.

packages/aio-commerce-lib-app/source/modules/association/types.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+/**
+ * The Commerce instance an app is associated with.
+ */
+export type AssociatedCommerceInstance = {
+  /** Commerce API base URL. */
+  baseUrl: string;
+  /** Deployment type of the Commerce instance. */
+  env: "saas" | "paas";
```
iivvaannxx: Probably should use the CommerceEnv from lib-core that @oshmyheliuk added in #510

packages/aio-commerce-lib-app/test/unit/modules/association/association-repository.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    test("writes the data under the system.association key", async () => {
+      const data = {
+        baseUrl: "https://example.com",
+        env: "paas" as const,
+      };
+      await setAssociationData(data);
+
+      expect(mockSetSystemConfigByKey).toHaveBeenCalledWith(
+        "system.association",
+        data,
```
iivvaannxx: I don't think this should be tested this way, we should set up a mock storage and test what's stored is what we expect. This couples the test to internal implementation details like the storage path.

packages/aio-commerce-lib-app/test/unit/modules/association/association-repository.test.ts
```diff
@@ -0,0 +1,88 @@
...
+        baseUrl: "https://example.com",
+        env: "saas" as const,
+      };
+      mockGetSystemConfigByKey.mockResolvedValue(data);
+
+      const result = await getAssociationData();
+
+      expect(mockGetSystemConfigByKey).toHaveBeenCalledWith(
+        "system.association",
+      );
```
iivvaannxx: Idem in here and maybe other cases

packages/aio-commerce-lib-app/test/unit/access/commerce-instance.test.ts
iivvaannxx, line 1: The name of the test file seems stale.

packages/aio-commerce-lib-config/source/modules/configuration/configuration-repository.ts
vinayrao2000: Restored in b2903198d72391d1fa8375c867ce65d5c973b6e8

packages/aio-commerce-lib-config/source/modules/configuration/configuration-repository.ts
vinayrao2000: Dropped the list call and now read the file directly.

Fixed in 5e948fe6bb6dca7d20aad06228c8c78d9b2ef854

packages/aio-commerce-lib-config/source/modules/configuration/configuration-repository.ts
vinayrao2000: Delete error now bubbles up to the caller instead of being swallowed.

Fixed in 3f0933371aded031a7aa616c9088a5a04aa6d223

packages/aio-commerce-lib-config/source/modules/configuration/system-config.ts
vinayrao2000: SYSTEM_NAMESPACE now lives in system-config.ts instead of being imported from configuration-repository.ts.

Moved in 9b42d0ecc9b580e67facf5560babec112f65afa8

packages/aio-commerce-lib-config/source/modules/configuration/system-config.ts
vinayrao2000: I reviewed it, loadConfig already deserializes: both the state-cache and files branches return JSON.parse(...), so it returns the parsed object and the as T is over that real object, not a raw string. The round-trip tests confirm it: getSystemConfigByKey returns an object deep-equal to exactly what was stored ({ baseUrl, env }).

packages/aio-commerce-lib-config/source/modules/configuration/system-config.ts
vinayrao2000: Added the = unknown default to getSystemConfigByKey, thanks.

For getConfigurationByKey, it's a different shape. It returns a fixed response where the only typed field is config.value, which is already a known BusinessConfigSchemaValue (string | number | boolean), not an opaque payload. A generic there would either lie about the real type or force threading T through the shared ConfigValue type for little gain. Happy to do it separately if there's a concrete need.

Fixed in 9b42d0ecc9b580e67facf5560babec112f65afa8.

packages/aio-commerce-lib-config/test/unit/modules/configuration/system-config.test.ts
vinayrao2000: Done, now using the shared createMockLibState / createMockLibFiles. Added an idempotent delete to the files mock since it was missing one.

 Fixed in f1130a74efadfd17c636f53ac72ec2b75fcf0858

packages/aio-commerce-lib-core/source/error/index.ts
vinayrao2000: Added a minor changeset for lib-core covering the new CommerceSdkErrorBaseOptions export. (The fiery-cobras-film changeset already covers the response-presets change separately.)

Fixed in b83a64f672cce4090c5345a22f03be171f516e1a

.changeset/commerce-instance-helpers.md
vinayrao2000: Done in 36fa0c9d280e238011b82458e0f8249928dedf39

packages/aio-commerce-lib-app/docs/usage.md
vinayrao2000: Good point, App Management is IMS-only. Narrowed getCommerceClient to accept just ImsAuthParams | ImsAuthProvider, and updated the docs/examples to use resolveImsAuthParams (with getImsAuthProvider / forwardImsAuthProvider called out for the provider route).

Fixed in cb74abd5855bdd71d883c1e216279732acd77821

packages/aio-commerce-lib-app/docs/usage.md
vinayrao2000: Right, the client prepends rest/{storeViewCode}/{version} itself. Dropped the prefix; examples now call client.get("products").

Fixed in cb74abd5855bdd71d883c1e216279732acd77821

packages/aio-commerce-lib-app/docs/usage.md
vinayrao2000: Added an optional fetchOptions second arg (typed from CommerceHttpClientParams["fetchOptions"]) that's forwarded to the AdobeCommerceHttpClient constructor, plus a test and docs update.

Fixed in cb74abd5855bdd71d883c1e216279732acd77821

packages/aio-commerce-lib-app/source/access/commerce-instance.ts
vinayrao2000: Done in the same change as the docs thread above, the signature is now 
auth: ImsAuthParams | ImsAuthProvider.

Fixed in cb74abd5855bdd71d883c1e216279732acd77821

packages/aio-commerce-lib-app/source/errors/app-not-associated-error.ts
vinayrao2000: Good catch, kept the JSDoc to just the public description and moved the "named seam" rationale to an inline comment.

Fixed in 6d4c3c3562f4d36592d589ccd70a92748bd70230

packages/aio-commerce-lib-app/source/management/association/association-repository.ts
vinayrao2000: Done, moved the association module to source/management/association/ (and mirrored the test). It's not re-exported from the /management entrypoint; management/index.ts is unchanged.

packages/aio-commerce-lib-app/source/modules/association/types.ts
vinayrao2000: Done. AssociatedCommerceInstance.env now uses CommerceEnv.

Fixed in 120de50c7c76ef4f5aefe788cc5dce6dc4cd52ba

packages/aio-commerce-lib-app/test/unit/modules/association/association-repository.test.ts
vinayrao2000: Done now uses a stateful storage fake and tests the round-trip (set → get → clear) instead of asserting the system.association key.

Fixed in 835cb899f45172a031bb9888c773ed065811734e

packages/aio-commerce-lib-app/test/unit/modules/association/association-repository.test.ts
vinayrao2000: Fixed in 835cb899f45172a031bb9888c773ed065811734e

packages/aio-commerce-lib-app/test/unit/access/commerce-instance.test.ts
vinayrao2000: Renamed to access/commerce-instance.test.ts to mirror its source.

Fixed in 0cc7ddec0b4a8d9f281a9b0e952073728c3ffd5a

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -788,7 +923,7 @@ After an app is associated with a Commerce instance via App Management, the SDK
 
 Two helpers are exposed from the root entrypoint:
 
-- `getCommerceClient(auth)` — returns a ready-to-use [`AdobeCommerceHttpClient`](../../aio-commerce-lib-api/docs/usage.md). Use this when you need to call the Commerce API. The base URL and flavor come from the stored association data; you supply the resolved auth credentials (resolve them with `resolveAuthParams` from [`@adobe/aio-commerce-lib-auth`](../../aio-commerce-lib-auth/docs/usage.md)).
+- `getCommerceClient(auth, fetchOptions?)` — returns a ready-to-use [`AdobeCommerceHttpClient`](../../aio-commerce-lib-api/docs/usage.md). Use this when you need to call the Commerce API. The base URL and flavor come from the stored association data; you supply the resolved IMS auth. App Management requires IMS, so this accepts only IMS auth: resolve params with `resolveImsAuthParams`, or pass an `ImsAuthProvider` built with `getImsAuthProvider` / `forwardImsAuthProvider` from [`@adobe/aio-commerce-lib-auth`](../../aio-commerce-lib-auth/docs/usage.md). The optional `fetchOptions` are forwarded to the underlying client (e.g. `headers`, `timeout`, `retry`).
```
iivvaannxx: A link to what `fetchOptions` is could be added. Not a blocker.

packages/aio-commerce-lib-app/docs/openapi.json
vinayrao2000: Done.

Fixed in cb74abd5855bdd71d883c1e216279732acd77821

packages/aio-commerce-lib-app/source/access/commerce-instance.ts
```diff
@@ -13,10 +13,14 @@
...
-import { getAssociationData } from "../modules/association/association-repository";
+import { getAssociationData } from "../management/association/association-repository";
 
 import type { CommerceHttpClientParams } from "@adobe/aio-commerce-lib-api";
-import type { AssociatedCommerceInstance } from "../modules/association/types";
+import type {
+  ImsAuthParams,
+  ImsAuthProvider,
+} from "@adobe/aio-commerce-lib-auth";
+import type { AssociatedCommerceInstance } from "../management/association/types";
```
iivvaannxx:
```suggestion
import { AppNotAssociatedError } from "#errors/app-not-associated-error";
import { getAssociationData } from "#management/association/association-repository";

import type { CommerceHttpClientParams } from "@adobe/aio-commerce-lib-api";
import type {
  ImsAuthParams,
  ImsAuthProvider,
} from "@adobe/aio-commerce-lib-auth";
import type { AssociatedCommerceInstance } from "#management/association/types";
```

packages/aio-commerce-lib-app/source/access/commerce-instance.ts
```diff
@@ -51,28 +55,32 @@ export async function getCommerceInstance(): Promise<AssociatedCommerceInstance>
  *
  * The base URL and flavor come from the stored association data
  * ({@link getCommerceInstance}); only the auth credentials are supplied by the
- * caller, already resolved. Resolve them outside with `resolveAuthParams` from
- * `@adobe/aio-commerce-lib-auth` so this helper stays focused on building the
- * client for the associated instance.
+ * caller, already resolved. App Management requires IMS, so this accepts only
+ * IMS auth: resolve params with `resolveImsAuthParams`, or pass an
+ * `ImsAuthProvider` built with `getImsAuthProvider` / `forwardImsAuthProvider`
+ * from `@adobe/aio-commerce-lib-auth`.
```
iivvaannxx: Maybe a bit too verbose, a JSDoc doesn't need to be a calc of docs.

packages/aio-commerce-lib-app/source/actions/association/schema.ts
vinayrao2000: As per the team discussion, commerceEnv uses CommerceEnvSchema from lib-core, and CommerceFlavor is already an alias of CommerceEnv (same saas/paas values). Keeping the commerceEnv name per #510.

Fixed in 120de50c7c76ef4f5aefe788cc5dce6dc4cd52ba

packages/aio-commerce-lib-app/test/unit/access/commerce-instance.test.ts
```diff
@@ -109,6 +109,23 @@ describe("getCommerceClient", () => {
...
+    mockGetAssociationData.mockResolvedValue(data);
+
+    const fetchOptions = { timeout: 5000, headers: { "x-trace": "abc" } };
+    await getCommerceClient(auth, fetchOptions);
+
+    expect(MockAdobeCommerceHttpClient).toHaveBeenCalledWith({
+      auth,
+      config: { baseUrl: data.baseUrl, flavor: "paas" },
+      fetchOptions,
+    });
```
iivvaannxx: For this test you don't need to assert it's called with `auth` and `config`, only `fetchOptions`. Keeps the test less likely to fail :)

packages/aio-commerce-lib-app/test/unit/management/association/association-repository.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      store,
+      mockGetSystemConfigByKey: vi.fn(async (key: string) =>
+        store.has(key) ? store.get(key) : null,
+      ),
+      mockSetSystemConfigByKey: vi.fn(async (key: string, value: unknown) => {
+        if (value === null || value === undefined) {
+          store.delete(key);
+        } else {
+          store.set(key, value);
+        }
```
iivvaannxx: Mocking this provides little to no-value, as these helpers are directly invoked from within the tested API. Again, the underlying storage layer would be a more correct mocking target.

iivvaannxx (approved):

packages/aio-commerce-lib-config/source/modules/configuration/system/system-repository.ts
vinayrao2000: Made the 24h TTL explicit, SYSTEM_CONFIG_CACHE_TTL_SECONDS = 86_400, passed through to the cache layer, so it's documented in code rather than relying on the service default.

Fixed in de6e29f9886c5533324904b02c9b11d15aa7cc33

packages/aio-commerce-lib-app/docs/usage.md
vinayrao2000: Added the link to ky's Options docs in usage.md. 

Fixed in 8fc63746a4fb2ec42bf536e992a5555e9fdbe383

packages/aio-commerce-lib-app/source/access/commerce-instance.ts
vinayrao2000: Moved to the # subpath imports to match the package convention.

Fixed in c62e6231949f37b5fa0d5de25ec18ffdc8ba478f

packages/aio-commerce-lib-app/source/access/commerce-instance.ts
vinayrao2000: Condensed it to a single @param auth line.

Fixed in c62e6231949f37b5fa0d5de25ec18ffdc8ba478f

packages/aio-commerce-lib-app/test/unit/access/commerce-instance.test.ts
vinayrao2000: Narrowed it to assert only fetchOptions.

Fixed in 05da1c6e3362b4f7b6460b04f1bf99b0a6bf99a3

.changeset/error-base-options-export.md
iivvaannxx, line 1: This was done by @oshmyheliuk in #522, probably needs to be deleted and check that everything still works as you expected

packages/aio-commerce-lib-app/test/unit/management/association/association-repository.test.ts
vinayrao2000: Done, now mocks the storage layer (aio-lib-state / aio-lib-files) instead of the config helpers, so it tests the real round-trip + max-TTL caching.

Fixed in 22d73d9df8832df9d3657646ba15dcc228cb81c5

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -8,6 +8,8 @@ The `@adobe/aio-commerce-lib-app` library provides:
 - **Business Configuration**: Generate and manage the runtime actions that power the `commerce/configuration/1` extension point.
 - **Installation Management**: Generate and manage the runtime action that powers the app installation flow.
 - **Admin UI Configuration** (`commerce/backend-ui/2`): Generate and manage the runtime action and `workerProcess` declarations for Admin UI extensions on `commerce/backend-ui/2`. Currently supports grid column extensions, mass actions, order view buttons, and menu declarations.
+- **Admin UI SDK Configuration** (`commerce/backend-ui/1`, _deprecated_): Generate and manage the runtime action for the legacy Admin UI SDK extension point. Will be removed from the SDK — use `adminUi` and `commerce/backend-ui/2` for new apps.
```
obarcelonap, line 11: @vinayrao2000 **Admin UI SDK Configuration** for v1 extension point shouldn't be added again, could you please fix this

vinayrao2000: Fixed in 75053c640de9c013ea867ff011a7886fef5a12b0

PR #513, opened 2026-06-10
CEXT-6338: spec for Admin UI v2 web-src scaffold

iivvaannxx:
## Description

Adds the spec for CEXT-6338: automatic scaffolding of `web-src/` for `commerce/backend-ui/2` view extensions.

## Related Issue

[CEXT-6338](https://jira.corp.adobe.com/browse/CEXT-6338)

## Motivation and Context

The `generate all` command already emits a `view` operation into `ext.config.yaml` when the app config requires iframe-based UI (mass actions, menu, view buttons), but without a corresponding `web-src/` directory that operation is completely non-functional. Developers currently copy files manually from the Commerce samples repository.

## How Has This Been Tested?

Spec only — no code changes.

## Types of changes

## Checklist:


specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,347 @@
...
+✔ Scaffolding web-src for commerce/backend-ui/2...
+```
+
+The generated tree under `src/commerce-backend-ui-2/`:
+
+```
+web-src/
+  index.html
+  src/
+    app.jsx
```
obarcelonap, line 111: I thought the convention for react components was to use PascalCase

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,347 @@
...
+    components/
+      main-page.jsx
+      extension-page.jsx
+```
+
+The scaffold is split into two zones with different regeneration semantics:
+
+- **SDK-managed**: `index.html`, `src/app.jsx`, and `src/components/main-page.jsx` are always
```
obarcelonap: I was wondering why we need MainPage to be skd managed. From my perspective one component (App) should be enough to wrap all the necessary stuff. Ideally MainPage should be scaffolded once but user managed and we can get rid of ExtensionPage

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,347 @@
...
+
+`main-page.jsx` is SDK-managed: it resolves shared context and forwards the values to
+`<ExtensionPage />`, the component developers fill in:
+
+```jsx
+import { useSharedContext } from "@adobe/aio-commerce-lib-admin-ui/web";
+import { ExtensionPage } from "./extension-page";
+
+export function MainPage() {
+  const { extensionId, imsToken, imsOrgId } = useSharedContext();
```
obarcelonap, line 172: if this hook is available why do we need to send values as props of ExtensionPage? I understand this is backed by a provider so we can get this props anywhere down the App component hierarchy.

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,347 @@
...
+because `app.jsx` reads `metadata.id` from `#app.commerce.config` at runtime.
+
+After copying files, `generateWebSrc` installs any missing `peerDependencies` of `lib-admin-ui`
+into the app using the package manager detected from the app's lock file.
+
+### Scaffold contents
+
+| File                                | Zone         | Role                                                                                              |
+| ----------------------------------- | ------------ | ------------------------------------------------------------------------------------------------- |
+| `index.html`                        | SDK-managed  | Entry point loaded by the `view` operation URL; mounts the React app                              |
```
obarcelonap: if index.html is sdk managed, should add index.css as scaffolded once file and referenced from index.html. If not, how do we plan the dev can declare its own styles?

obarcelonap (commented):
I believe this approach effectively constrains the SPA to be implemented exclusively with React. While I’m comfortable with this direction, it would be good to explicitly document it as a drawback.

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
iivvaannxx: It is for the actual component, but not for the file. File names are kind of opinionated, as far as I know; there's no strict convention for that. I chose the kebab-case format because it aligns with the rest of the auto-generated code (e.g., `src/commerce-configuration-1` or `app-config.js`).

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
iivvaannxx: Makes sense 👍🏻

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
iivvaannxx: No need, I guess it's just for convenience (avoiding a call), I can remove those.

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
iivvaannxx: We can include an `index.css` entry, and users can compose their own via CSS `@import` if they want.

iivvaannxx:
> I believe this approach effectively constrains the SPA to be implemented exclusively with React. While I’m comfortable with this direction, it would be good to explicitly document it as a drawback.

@obarcelonap Not necessarily, codegen can always be extended to support other languages, and `@adobe/aio-commerce-lib-admin-ui/web` can be turned into `@adobe/aio-commerce-lib-admin-ui/web/react` with room for future frameworks, such as `@adobe/aio-commerce-lib-admin-ui/web/vue` (although I don't see that happening in the near future, TBH, as much as I would like to implement the flexibility).

As a side note, does `aio app build` even support any other library besides React?

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
obarcelonap: I'd remove because I feel that if we add this people will get it as pattern and will pass this values down the hierarchy instead of using the hook when needed.

obarcelonap:
> > I believe this approach effectively constrains the SPA to be implemented exclusively with React. While I’m comfortable with this direction, it would be good to explicitly document it as a drawback.
> 
> @obarcelonap Not necessarily, codegen can always be extended to support other languages, and `@adobe/aio-commerce-lib-admin-ui/web` can be turned into `@adobe/aio-commerce-lib-admin-ui/web/react` with room for future frameworks, such as `@adobe/aio-commerce-lib-admin-ui/web/vue` (although I don't see that happening in the near future, TBH, as much as I would like to implement the flexibility).
> 
> As a side note, does `aio app build` even support any other library besides React?

I meant you can not opt out of the code generation and even if removed will be regenerated which means you're stuck with React. I am not saying this is bad, for me is just enough. We can always add in the future a way to opt out of the scaffolding if there's a real need.
Just wanted to note that this is a drawback of the current scaffolding design.

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,342 @@
...
+  routes: [
+    { index: true, element: <MainPage /> },
+    { path: "bulk-cancel", element: <BulkCancelPage /> },
+    { path: "archive", element: <ArchivePage /> },
+  ],
+});
+```
+
+Each `path` here corresponds to the hash fragment declared in `app.commerce.config.ts` — for
+example `path: "#/bulk-cancel"` maps to the `"bulk-cancel"` route above. `BulkCancelPage` and `ArchivePage` are additional user-created components.
```
obarcelonap: could we validate that paths on `app.commerce.config.ts` exist in `routes`?

iivvaannxx: should be doable, and probably auto-generated too.

iivvaannxx: Done in [`2d39379`](https://github.com/adobe/aio-commerce-sdk/pull/513/commits/2d39379a78f913344e4996630d986cde0941e67c)

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,351 @@
...
+and delegates all shell wiring to `createExtensionApp` from `@adobe/aio-commerce-lib-admin-ui/web`:
+
+```jsx
+import { createExtensionApp } from "@adobe/aio-commerce-lib-admin-ui/web";
+import { MainPage } from "./components/main-page";
+import config from "#app.commerce.config";
+
+createExtensionApp({
+  extensionId: config.metadata.id,
+  routes: [{ index: true, element: <MainPage /> }],
```
obarcelonap, line 133: As we commented, `routes` definition has to be in a file that is user managed because we don't know the file structure that the dev is going to use.
A more flexible approach would be:
- if `routes` file is present, inject the exported array in `createExtensionApp` and install the required dependency
- if not present, use `rootComponent` prop pointing to `MainPage` so routing is not used by the framework but the dev could implement its own routing solution.

iivvaannxx: Addressed in [`8e57fc0`](https://github.com/adobe/aio-commerce-sdk/pull/513/commits/8e57fc061311491a3ad0bb54521c1d78b6743abe)

obarcelonap (approved):

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,373 @@
...
+    app.jsx
+    routes.jsx
+    pages/
+      main-page.jsx
+    components/
+```
+
+The scaffold is split into two zones with different regeneration semantics:
+
+- **SDK-managed**: `index.html` and `src/app.jsx` are always regenerated on every `generate all`
```
jcuerdo: We should then specify (in docs or in a file comment) that these files should not be modified.

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,373 @@
...
+
+**`createExtensionApp({ extensionId, routes })`** — replaces the entire `index.js` +
+`exc-runtime.js` + `App.js` + `ExtensionRegistration.js` quad from the raw Commerce sample. It
+handles the EC runtime loading internally via `@adobe/exc-app`, bootstraps the shell with a
+fallback to a mock runtime for local development, mounts a `HashRouter` with a Spectrum 2
+`Provider` and a default `ErrorBoundary`, registers `runtime.on('configuration')` and
+`runtime.on('history')` EC shell event handlers, and calls `register()` from `@adobe/uix-guest`
+with the given `extensionId`. The `routes` array follows the React Router v6 route object format
+and is rendered inside the router — the index route is typically the main page component.
+Alternatively, `rootComponent` can be passed instead of `routes` for apps that manage their own
```
jcuerdo: Is it possible to pass both? What happens if this happens?

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,373 @@
...
+```
+
+The scaffold is split into two zones with different regeneration semantics:
+
+- **SDK-managed**: `index.html` and `src/app.jsx` are always regenerated on every `generate all`
+  and pre-app-build run. They contain no user logic — developers should not modify them.
+- **User-managed**: `index.css`, `src/routes.jsx`, and `src/pages/main-page.jsx` are generated
+  once and never touched again. This is where developers declare their routes, write their UI, and
+  add their styles. CSS composition via `@import` is the recommended way to add additional
+  stylesheets. `src/components/` is left empty for developers to add shared UI primitives.
```
jcuerdo: Using .gitkeep?

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
iivvaannxx: Yes, but that's more of a doc concern/implementation detail that I don't think the spec needs to reflect. But I will take it into account 👍🏻

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
iivvaannxx: Not necessarily. `.gitkeep` is used to track empty folders in source control. We can create it and let users decide whether they keep it, use it, or remove it.

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
iivvaannxx: Should be exclusive, and should throw both a type error (if using TypeScript checking) and a runtime error. Or we can make one take precedence over the other, but that may be confusing.

iivvaannxx: Clarified in [`e2bf617`](https://github.com/adobe/aio-commerce-sdk/pull/513/commits/e2bf6170f01006c527cc4e089cd42d4d4eda17fc)

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,374 @@
...
+          type: "view",
+          path: "#/bulk-cancel",
+        },
+      ],
+    },
+  },
+});
+```
+
+Running `generate all` scaffolds the `web-src/` alongside the ext.config update:
```
asalloum5, line 97: Are we implicitly saying that we only check the type to see if it's view?

jcuerdo (approved):

asalloum5 (commented):
I have a doubt here, do we handle non-react apps? I didn't see anything related to this.

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
```diff
@@ -0,0 +1,374 @@
...
+
+```
+web-src/
+  index.html
+  index.css
+  src/
+    app.jsx
+    routes.jsx
+    pages/
+      main-page.jsx
```
asalloum5, line 113: Do we handle `tsx`?

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
iivvaannxx: Nope, we detect if web scaffolding is needed also if `menu` is present in the config.

specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md
iivvaannxx: This is not on our own, it depends on `aio` tooling. Not sure if `aio app build` handles TypeScript by default on `web-src` (it does not for runtime actions, without extra setup). It may handle it, as it uses Parcel for web and that bundler is thought to work OOTB almost always, but I would need to check.

I'll run tests, if compatible, I implement whatever is needed, if not, I'll create a ticket. But language support is not in the scope of this ticket I'd say, probably would be [CEXT-6081](https://jira.corp.adobe.com/browse/CEXT-6081)

PR #522, opened 2026-06-16
CEXT-6151: Add ACL admin UI schema and helper to check permissions

oshmyheliuk:
## Description

https://jira.corp.adobe.com/browse/CEXT-6151

1. Update adminUISdk schema to support ACL fields for menu items
2.  Add a public helper that extensions can call from the SPA and from runtime actions to check whether the current user is allowed to use a given ACL resource.

## Related Issue

https://github.com/magento-commerce/adobe-commerce-backend-uix/pull/352

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -835,6 +836,27 @@ adminUi: {
...
+    getAclResourceId,
+    getAdminUiPermissionClient,
+  } from "@adobe/aio-commerce-lib-admin-ui/api";
+
+  // Derive the resource id from metadata.id:
+  getAclResourceId("acme-promotions");
+  // → "Magento_CommerceBackendUix::adminuisdk_app_acme_promotions"
+
+  // Check access in a runtime action:
+  const client = getAdminUiPermissionClient({
```
obarcelonap, line 854: would it be possible to create a function in `aio-commerce-lib-app` that returns a permission client which auto resolves the app id from the one set in the commerce config so we don't need to pass this value on each function call.
Something like
```
const app = getAppPermissionClient({httpClient}); // resolves appId from app.commerce.config
```

obarcelonap (approved):

packages/aio-commerce-lib-app/docs/usage.md
iivvaannxx: > resolves appId From app.commerce.config

That's something we may need to be cautious with. Runtime code (that runs in a runtime action) that depends on that file can not easily read the configuration unless it's bundled with the action, which requires extra setup in ext.config.yaml via `include` or via some import trick (messy bundle/build stuff). 

It's the same problem we have (for now) with lib-config: we need to receive the schema in `initialize`. There are ways to make this work, but they are not trivial.

packages/aio-commerce-lib-admin-ui/source/api/lib/acl-resource-id.ts
```diff
@@ -0,0 +1,68 @@
...
+ * This is a cross-repo contract shared with the Commerce module. Both sides must produce
+ * the exact same output for the same input. Any change here must be coordinated with the
+ * Commerce module's ACL id generator.
+ *
+ * @param metadataId - The application's `metadata.id` value (e.g. `"approval-dashboard-app"`).
+ * @param menuId - The menu item's `id` value from `adminUi.menu.id` (e.g. `"approval_dashboard"`).
+ * @returns The full Commerce ACL resource id for the menu leaf node (e.g.
+ *   `"Magento_CommerceBackendUix::adminuisdk_app_approval_dashboard_app_menu_approval_dashboard"`).
+ */
+export function getMenuAclResourceId(
```
jcuerdo, line 52: Should be documented in usage.md?

packages/aio-commerce-lib-admin-ui/docs/usage.md
iivvaannxx, line 1: This documentation should be completely agnostic of `app.commerce.config`, it doesn't know about it.

packages/aio-commerce-lib-admin-ui/docs/usage.md
```diff
@@ -246,3 +247,100 @@ if (isCommerceMenu(id)) {
...
+Apps that set `adminUi.menu.aclProtected: true` in their `app.commerce.config.*` receive a dedicated per-app ACL resource node in the Commerce User Roles tree. Admins can grant or deny access to that app's menu on a per-role basis. Use `getAdminUiPermissionClient` from `@adobe/aio-commerce-lib-admin-ui/api` to check whether the current admin user holds that resource before serving menu content from a runtime action.
+
+#### Creating a client
+
+```typescript
+import { getAdminUiPermissionClient } from "@adobe/aio-commerce-lib-admin-ui/api";
+
+const permissionClient = getAdminUiPermissionClient({
+  httpClient, // AdobeCommerceHttpClient — reuse the one from createAdminUiApiClient
+  appId: "acme-promotions", // metadata.id from your app config; enables no-argument calls
```
iivvaannxx: What's the `appId` used for? What happens if I am not an App Management user? If this is intended to be used with App Management shouldn't it be in `lib-app`?

packages/aio-commerce-lib-admin-ui/source/api/lib/acl-resource-id.ts
```diff
@@ -0,0 +1,68 @@
...
+    .toLowerCase()
+    .replace(/[^a-z0-9_]/g, "_");
+}
+
+/**
+ * Derives the deterministic Commerce ACL resource id for an app from its metadata id.
+ *
+ * This is a cross-repo contract shared with the Commerce module. Both sides must produce
+ * the exact same output for the same input. Any change here must be coordinated with the
+ * Commerce module's ACL id generator.
```
iivvaannxx: This will end up in public docs. Isn't this kind of an internal comment?

packages/aio-commerce-lib-admin-ui/source/api/lib/permission-client.ts
```diff
@@ -0,0 +1,228 @@
...
+import { getAclResourceId } from "./acl-resource-id";
+
+import type { AdobeCommerceHttpClient } from "@adobe/aio-commerce-lib-api";
+
+const DEFAULT_CACHE_TTL_MS = 300_000;
+
+/** Options used to create an Admin UI SDK permission client. */
+export interface AdminUiPermissionClientOptions {
+  /** The application's `metadata.id` value. When provided, `check()` and `require()` can be called with no resource argument. */
+  appId?: string;
```
iivvaannxx, line 29: Now that I see how `appId` works, I would challenge the fact that it is tightly coupled to a concept outside this library (app IDs). It's a useful convenience, but I am not sure it's the right layer to implement it.

packages/aio-commerce-lib-admin-ui/source/api/lib/permission-client.ts
```diff
@@ -0,0 +1,228 @@
...
+   * Clears the cached result for `resource`. If called without an argument, clears
+   * all cached entries and in-flight tracking without aborting outstanding HTTP requests.
+   */
+  invalidate(resource?: string): void;
+  /**
+   * Resolves when the current user has the given resource granted.
+   * Throws `AdminUiPermissionDeniedError` if denied.
+   * Throws `AdminUiPermissionError` on 401, network, and parse errors.
+   * When called with no argument, defaults to the ACL resource id derived from `appId`.
+   * Throws `AdminUiPermissionError` immediately when neither `resource` nor a valid `appId` is available.
```
iivvaannxx: These lines should use proper JSDoc so that the autogenerated documentation picks those correctly. `@throws` with proper `@link`

packages/aio-commerce-lib-admin-ui/source/api/lib/permission-client.ts
```diff
@@ -0,0 +1,228 @@
...
+/** Wraps an arbitrary thrown value in an `AdminUiPermissionError`, passing through instances that are already one. */
+function toPermissionError(error: unknown) {
+  return error instanceof AdminUiPermissionError
+    ? error
+    : new AdminUiPermissionError("Permission check failed", {
+        cause: error,
+      });
+}
+
+/** Creates a client for checking Admin UI SDK ACL resources. */
```
iivvaannxx: `options` is not documented as `@param`.

packages/aio-commerce-lib-admin-ui/source/api/permissions/endpoints.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export async function checkPermission(
+  httpClient: AdobeCommerceHttpClient,
+  params: PermissionCheckParams,
+): Promise<PermissionCheckResponse> {
+  const raw = await httpClient
+    .post("adminuisdk/permission/check", {
+      json: { resource: params.resource },
+    })
+    .json<unknown>();
```
iivvaannxx: I believe `unknown` is the default so `.json()` should be equivalent.

packages/aio-commerce-lib-admin-ui/source/api/permissions/endpoints.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+/**
+ * Checks whether the current user has the given ACL resource granted via POST /V1/adminuisdk/permission/check.
+ * This is the raw HTTP call — prefer {@link getAdminUiPermissionClient} for caching and deduplication.
+ *
+ * @param httpClient - The {@link AdobeCommerceHttpClient} to use to make the request.
+ * @param params - The resource to check.
+ *
+ * @throws An `HTTPError` if the status code is not 2XX.
```
iivvaannxx: Maybe `@link` is better for this throws?

packages/aio-commerce-lib-admin-ui/source/errors.ts
```diff
@@ -0,0 +1,29 @@
...
+/** Base error for Admin UI SDK permission helper failures. */
+export class AdminUiPermissionError extends CommerceSdkErrorBase {}
+
+/** Error thrown when the current user is denied access to an Admin UI SDK ACL resource. */
+export class AdminUiPermissionDeniedError extends AdminUiPermissionError {
+  public readonly resource: string;
+
+  public constructor(
+    resource: string,
+    options?: ErrorOptions & { traceId?: string },
```
iivvaannxx: From where is this `ErrorOptions` sourced? In any case, I would create a custom `Options` for each error even if equivalent to the default, this way we decouple both and we avoid type contract changes in the future if we need to add some options.

specs/features/CEXT-6150-admin-ui-nested-acl.md
iivvaannxx, line 1: This wasn't merged separately? In any case `Implemented` should be marked.

packages/aio-commerce-lib-admin-ui/source/api/lib/acl-resource-id.ts
oshmyheliuk: Added examples here https://github.com/adobe/aio-commerce-sdk/pull/522/changes/e7265784b5e6d630372d9838aef7ec724787eae8

packages/aio-commerce-lib-admin-ui/docs/usage.md
oshmyheliuk: `appId` is used to generate the ACL resource ID for future permission checks in the permission client.
I think it should be a part of the admin-ui library. The comment in usage.md was confusing, so I updated it.

packages/aio-commerce-lib-admin-ui/source/api/lib/acl-resource-id.ts
oshmyheliuk: Yeah, it's more like an internal comment. We need to maintain consistent logic between Commerce and aio-commerce-sdk to generate the same ACL resources.
I updated comments https://github.com/adobe/aio-commerce-sdk/pull/522/commits

packages/aio-commerce-lib-admin-ui/source/api/lib/permission-client.ts
oshmyheliuk: Updated in https://github.com/adobe/aio-commerce-sdk/pull/522/changes/5f29d200b9029f5feaa93cec62d0c3677e9e3646

packages/aio-commerce-lib-admin-ui/source/api/lib/permission-client.ts
oshmyheliuk: Updated in https://github.com/adobe/aio-commerce-sdk/pull/522/changes/5f29d200b9029f5feaa93cec62d0c3677e9e3646

packages/aio-commerce-lib-admin-ui/source/api/permissions/endpoints.ts
oshmyheliuk: Updated in https://github.com/adobe/aio-commerce-sdk/pull/522/changes/e958f27efdd1539ee24bcf79cc448db27d0e9fd8

packages/aio-commerce-lib-admin-ui/source/api/permissions/endpoints.ts
oshmyheliuk: Added in https://github.com/adobe/aio-commerce-sdk/pull/522/changes/e958f27efdd1539ee24bcf79cc448db27d0e9fd8

packages/aio-commerce-lib-admin-ui/source/errors.ts
oshmyheliuk: I added an export for error options from the core package
https://github.com/adobe/aio-commerce-sdk/pull/522/changes/9ec12345de4fa46f39acb0cea3eec9bb49ff2df3. It sounds logical to re-use them.

specs/features/CEXT-6150-admin-ui-nested-acl.md
oshmyheliuk: Marked

packages/aio-commerce-lib-admin-ui/source/api/lib/permission-client.ts
oshmyheliuk: The `appID` is optional; the client can check permissions directly by providing the full ACL Resource ID. 
But this client can check only resources created by the Admin UI commerce module. All ACL resources are created based on the installed apps, so the client is tightly coupled to the Admin UI and the installed applications' logic.

It's not like you can use this client to check any ACL Resource in Commerce; for resources not created by the Admin UI module, the API endpoint would always return `false`.

packages/aio-commerce-lib-admin-ui/source/errors.ts
iivvaannxx: That's good, it's also done in #511 so if that one is merged first you might get a conflict. What I am curious is how did `typecheck` pass before if `ErrorOptions` wasn't in scope 🤣 

Going to leave the conversation open so you can acknowledge the conflict warning, but feel free to resolve 👍🏻

packages/aio-commerce-lib-admin-ui/source/api/lib/permission-client.ts
iivvaannxx: > It's not like you can use this client to check any ACL Resource in Commerce; for resources not created by the Admin UI module, the API endpoint would always return false.

Hmm, I see. And that's a technical limitation or a deliberate decision?

On a side note, will this work only for Admin UI V2? If so, then I guess it's fine because V2 its intended to work on App Management so my concern about "outside concept" disappears. 

I think that documenting this "limitation" would be a good move before merging this (if not done already, I might have missed it)

packages/aio-commerce-lib-admin-ui/docs/usage.md
iivvaannxx: Depending on the answer to https://github.com/adobe/aio-commerce-sdk/pull/522#discussion_r3431120621, this comment may be removed. I would suggest maybe not focusing the documentation on what needs to be configured in `app.commerce.config.ts` but on what the `app-config` runtime action should return (which, if I am not mistaken, is what Admin UI V2 calls). That runtime action might not be wired by `lib-app` at all, but manually.

But this one is not a blocker, if you think I'm just overthinking it, feel free to skip.

iivvaannxx (approved):
Approving so merging is not blocked. There are just a few comments I left open as they may or may not need resolution.

packages/aio-commerce-lib-admin-ui/source/errors.ts
oshmyheliuk: > typecheck pass before if ErrorOptions wasn't in scope 

Yeah, it's strange, I don't have an answer :)

packages/aio-commerce-lib-admin-ui/source/api/lib/permission-client.ts
oshmyheliuk: I added the section to usage.md that explains it https://github.com/adobe/aio-commerce-sdk/pull/522/changes/fe761226acb567e6a4b523498d37aa65354bcce3

oshmyheliuk:
I'm merging this PR as going to work on the phase 2 implementation. If there are still open questions, they can be addressed and resolved in the following PR

PR #534, opened 2026-06-22
CEXT-6394: implement snapshot-driven uninstall

obarcelonap:
## Description

Implements the snapshot-driven uninstall design from #533.

- Adds an optional `config` field to `InstallationStateBase`; the initial-state factory stamps it on every new state and it flows through in-progress, succeeded, failed, and retry states without extra plumbing.
- At uninstall time, the `POST /uninstallation` handler loads the persisted installation snapshot. If it is a completed state with a recorded config, that config drives the uninstall step tree. Otherwise it falls back to the current request config (legacy installs, in-progress install, or cleared record).
- The resolved config is threaded through the existing async invocation to the execution handler; no new transport is introduced.
- `docs/openapi.json` is bumped to 2.1.0 and the `config` field is added to all state schemas.

Stacked on #533.

## Related Issue

- [CEXT-6394](https://jira.corp.adobe.com/browse/CEXT-6394) (under epic [CEXT-6086](https://jira.corp.adobe.com/browse/CEXT-6086))

## Motivation and Context

App uninstallation previously derived what to remove from the current request config. When the config drifted after install, resources were silently orphaned. This implements the fix described in the spec: drive uninstall from the recorded install snapshot, with a graceful fallback for records that predate this change.

## How Has This Been Tested?

Unit tests added for:
- Snapshot config is used when a completed install snapshot with a recorded config exists (drift scenario).
- Falls back to request config when the snapshot has no recorded config (legacy install).
- Falls back to request config when the installation store holds an in-progress state.
- `createInitialState` stamps config on the state and `createRetryState` carries it through.

Full suite: 893 tests pass.

N/A

## Types of changes

## Checklist:


oshmyheliuk:
@obarcelonap, during uninstallation, there may be an error with custom installation scripts.
As the custom installation script can be in the configuration, but no longer physically exists. 
The code checks if the script file exists and throws an error if it does not
https://github.com/adobe/aio-commerce-sdk/blob/main/packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts#L186

```
  if (!scriptModule) {
    throw new Error(
      `Script ${script} not found in customScripts context. Make sure the script is defined in the configuration and the action was generated with custom scripts support.`,
    );
  }
```
I think we can update the code to ignore this error during uninstallation.

obarcelonap:
> @obarcelonap, during uninstallation, there may be an error with custom installation scripts. As the custom installation script can be in the configuration, but no longer physically exists. The code checks if the script file exists and throws an error if it does not https://github.com/adobe/aio-commerce-sdk/blob/main/packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts#L186
> 
> ```
>   if (!scriptModule) {
>     throw new Error(
>       `Script ${script} not found in customScripts context. Make sure the script is defined in the configuration and the action was generated with custom scripts support.`,
>     );
>   }
> ```
> 
> I think we can update the code to ignore this error during uninstallation.

done 74034e2

jcuerdo (approved):
👍

packages/aio-commerce-lib-app/docs/openapi.json
```diff
@@ -3,7 +3,7 @@
   "info": {
     "title": "App Management API",
     "description": "REST API for managing Adobe Commerce App Builder apps: resolving app configuration, managing configuration values, running the installation and uninstallation lifecycle, and managing the scope tree.",
-    "version": "2.0.0",
+    "version": "2.1.0",
```
iivvaannxx: Not needed!

packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts
```diff
@@ -184,6 +189,14 @@ function createCustomScriptStep(scriptConfig: CustomInstallationStep): AnyStep {
       logger.debug(`Uninstalling custom script: ${name}`);
 
       const scriptModule = getScriptModule(customScripts, script);
+      if (!scriptModule) {
+        logger.warn(
+          `Script ${script} not found in customScripts context, skipping uninstall. It may have been removed from the project after being configured.`,
+        );
+
+        return;
+      }
```
iivvaannxx, line 198: I've been looking at this for a couple of minutes but I can't understand why we would want to change it?

iivvaannxx (approved):

packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts
obarcelonap: see https://github.com/adobe/aio-commerce-sdk/pull/534#issuecomment-4770303353
basically it could be the deployed app does not contain the installation script anymore (removed) so we don't want to crash the uninstallation.

iivvaannxx: Okay, but we should consider how to surface a message like this to the app manager.

packages/aio-commerce-lib-app/docs/openapi.json
obarcelonap: reverted

packages/aio-commerce-lib-app/source/management/installation/custom-installation/custom-scripts.ts
obarcelonap: For the moment there's no way to write a msg from installation script visible in app-management ui. From my perspective this is and edge case, if this appears again we can find a way to do it.
I'll keep it as is.

PR #537, opened 2026-06-23
CEXT-6270: [SPEC] Enable Runtime Actions to Emit Configured Events by Provider and Event Name

jcuerdo:
## Types of changes

## Checklist:

### DIFF
--- a/specs/features/CEXT-6270-emit-configured-event.md
+++ b/specs/features/CEXT-6270-emit-configured-event.md
@@ -135,15 +135,30 @@ the same fallback used by `generateInstanceId`. The event code is the value retu
 `getIoEventCode(getNamespacedEvent(metadata, event.name), providerType)`, the same
 computation already performed during installation.
 
-### I/O Events publish endpoint
+### `publishEvent` in lib-events
 
-The I/O Events ingress (`https://eventsingress.adobe.io/`) accepts a CloudEvents 1.0 payload.
-The required fields are `specversion`, `type` (the event code), `source`
-(`urn:uuid:{providerId}`), `id` (a fresh UUID per call — the API requires `source + id` to
-be unique per distinct event), `datacontenttype: "application/json"`, and `data` (the
-payload). The request must carry two auth headers: `Authorization: Bearer {access_token}` and
-`x-api-key: {client_id}`. Both are available from the standard IMS auth params resolved by
-`resolveImsAuthParams`.
+A new export from `@adobe/aio-commerce-lib-events` that owns the I/O Events ingress HTTP
+call and CloudEvents envelope construction:
+
+```ts
+publishEvent<TPayload extends Record<string, unknown>>(params: {
+  client: AdobeIoEventsApiClient;
+  providerId: string;
+  eventCode: string;
+  payload: TPayload;
+}): Promise<void>
+```
+
+`publishEvent` operates at the resolved level: callers supply the provider UUID and the
+fully-qualified event code directly. It builds a CloudEvents 1.0 envelope and POSTs it to the
+ingress URL configured on `client`.
+
+The I/O Events ingress (`https://eventsingress.adobe.io/`) requires the following fields:
+`specversion`, `type` (the event code), `source` (`urn:uuid:{providerId}`), `id` (a fresh
+UUID per call — the API requires `source + id` to be unique per distinct event),
+`datacontenttype: "application/json"`, and `data` (the payload). The request must carry two
+auth headers: `Authorization: Bearer {access_token}` and `x-api-key: {client_id}`. Both come
+from the IMS auth params on `client`.
 
 The ingress base URL is a new optional field on `IoEventsHttpClientConfig` in `lib-api`:
 `ingressBaseUrl`, defaulting to `https://eventsingress.adobe.io/` when not set. This keeps
@@ -174,11 +189,12 @@ Internal flow of `publishEvent`:
 2. If null, throw `EventsDataNotInitializedError`.
 3. Look up `params.provider` in `data.providers`. If not found, throw `ProviderNotFoundError`.
 4. Look up `params.event` in the provider's `events` map. If not found, throw `EventNotFoundError`.
-5. POST a CloudEvents envelope to `client`'s `ingressBaseUrl` with `Authorization` and `x-api-key` headers.
+5. Call `publishEvent` from `lib-events` with the resolved `providerId` and `eventCode`.
 
 ### Changeset
 
 - `@adobe/aio-commerce-lib-api`: `minor` — adds `ingressBaseUrl` to `IoEventsHttpClientConfig`.
+- `@adobe/aio-commerce-lib-events`: `minor` — new `publishRawEvent` export.
 - `@adobe/aio-commerce-lib-app`: `minor` — new `publishEvent` export plus installation-side
   `system.events` write.
 - `@adobe/aio-commerce-sdk`: `minor` — re-exports the above.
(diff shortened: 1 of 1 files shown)


specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -0,0 +1,206 @@
...
+- **Created:** 2026-06-23
+- [ ] **Implemented**
+
+## Summary
+
+Add an `emitConfiguredEvent` helper to `@adobe/aio-commerce-lib-app` that lets a runtime
+action publish a custom I/O Event by referencing the provider and event as declared in
+`app.commerce.config.ts`, with no knowledge of I/O Events provider IDs, event codes, or
+CloudEvents formatting required. Provider resolution is backed by metadata written to system
+storage during installation, so emit calls require no management API round-trip at runtime.
```
iivvaannxx: I think `emitEvent` is better and would also work, or `emitAppEvent`.

specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -0,0 +1,206 @@
...
+
+### `emitConfiguredEvent` in lib-app
+
+A new export from the root entrypoint of `@adobe/aio-commerce-lib-app`:
+
+```ts
+emitConfiguredEvent(params: {
+  provider: string;
+  event: string;
+  payload: unknown;
```
iivvaannxx: `payload` needs to be an object so I would suggest `Record<string, unknown>`

specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -0,0 +1,206 @@
...
+coupling to the computation logic.
+
+## Unresolved questions
+
+- **Authentication.** Publishing to the I/O Events ingress requires an IMS access token.
+  The ticket example shows no `auth` parameter in the call, so the intent is for the SDK to
+  source the token automatically (e.g. from env vars or stored workspace credentials).
+  Confirm whether this is feasible, or whether the developer will need to pass `auth`
+  explicitly as in `getCommerceClient`. Also confirm which OAuth scope the ingress requires
+  and whether the SDK-managed credentials already include it.
```
iivvaannxx: This sounds to something that should be resolved in the spec. Because this will be done by developers in their runtime actions as-needed they are going to need to have a `provider` ready to be used for this call to work. I would suggest an approach similar to `getCommerceClient`, for which they are going to need to set the necessary `AIO_COMMERCE_AUTH` inputs in their actions.

specs/features/CEXT-6270-emit-configured-event.md
obarcelonap: I'd go for `publishEvent` so we're in sync with IO events docs: https://developer.adobe.com/events/docs/guides/api/eventsingress-api

specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Developer experience
+
+After this feature ships, emitting a custom event from a runtime action looks like this:
+
+```ts
+import { emitConfiguredEvent } from "@adobe/aio-commerce-lib-app";
+
+export async function main(params) {
+  await emitConfiguredEvent({
```
obarcelonap: We could make `emitConfiguredEvent` generic to enforce the payload type:

```
async function emitConfiguredEvent<TPayload>(config: {
  provider: string;
  event: string;
  payload: TPayload;
}): Promise<void>
```

specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -0,0 +1,206 @@
...
+}): Promise<void>
+```
+
+It constructs a CloudEvents 1.0 envelope (`specversion`, `type` = `eventCode`,
+`source` = `urn:uuid:{providerId}`, `id` = fresh UUID, `datacontenttype: "application/json"`,
+`data` = `payload`) and POSTs to the I/O Events ingress (`https://eventsingress.adobe.io/`).
+Authentication is sourced as described in Unresolved questions. Errors from the ingress
+propagate as-is to the caller.
+
+### `emitConfiguredEvent` in lib-app
```
obarcelonap: What do you plan to use as `event_id`?
See https://developer.adobe.com/events/docs/guides/api/eventsingress-api

specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -0,0 +1,206 @@
...
+## Unresolved questions
+
+- **Authentication.** Publishing to the I/O Events ingress requires an IMS access token.
+  The ticket example shows no `auth` parameter in the call, so the intent is for the SDK to
+  source the token automatically (e.g. from env vars or stored workspace credentials).
+  Confirm whether this is feasible, or whether the developer will need to pass `auth`
+  explicitly as in `getCommerceClient`. Also confirm which OAuth scope the ingress requires
+  and whether the SDK-managed credentials already include it.
+
+- **Installation failure handling.** If the `system.events` write fails, should the
```
obarcelonap: I believe it should fail, otherwise users could get unexpected runtime errors

jcuerdo:
@iivvaannxx @obarcelonap changes done.

specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -123,84 +129,68 @@ computation already performed during installation.
 
 ### I/O Events publish endpoint
```
obarcelonap: don't we need the changes in `@adobe/aio-commerce-lib-events`?  from my perspective it made sense how it was before:
* lib-events offers a low level api with `providerId`
* lib-app resolves the `providerId` based on commerce app config and delegates to lib-events

specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -123,84 +129,68 @@ computation already performed during installation.
...
 
-A new function `publishIoEvent` is added to `@adobe/aio-commerce-lib-events/io-events`:
+The I/O Events ingress (`https://eventsingress.adobe.io/`) accepts a CloudEvents 1.0 payload.
```
obarcelonap: In order to be able to test in stage, we need some kind of flag to use stage ingress. Please consider this in the design

iivvaannxx: If I recall correctly, for the I/O Events API, we allow a `baseUrl` config, and we default it to prod. It should be baked into the `AdobeIoEventsHttpClient`, but not sure if there's room for a different URL, since the Event Ingress and the I/O Events endpoints are not the same.

specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -147,21 +152,29 @@ publishEvent<TPayload extends Record<string, unknown>>(params: {
...
 ```
 
 The generic `TPayload` lets callers enforce the payload shape at the call site without
 requiring the SDK to know the schema. The constraint `Record<string, unknown>` rules out
 primitives and arrays, matching what the I/O Events ingress expects as `data`.
 
+`ingressBaseUrl` is an optional escape hatch for testing against non-production ingress
+endpoints directly in code. In production, prefer the `AIO_EVENTS_INGRESS_BASE_URL` action
+input so the URL is configured per environment rather than hardcoded at call sites.
+
```
iivvaannxx: If we align with the way we generally do things in the SDK, which is composability instead of helpers that do a lot of things internally, for me, this should be different. 

The `ingressBaseUrl` config should live in the `AdobeIoEventsHttpClient`, then users would need to do a dual step-here:

```ts
const client = getIoEventsClient(auth, { config: { ingressBaseUrl: "..." }})
publishEvent(client, event, payload) // or publishEvent({ client, event, payload }) if in the future we need to extend signature without breaking changes (probably not likely).
```

jcuerdo: I get your point, but isn't **getCommerceClient** doing the same? It also reads from storage internally and wires everything up.

iivvaannxx: When you say "doing the same" what do you mean exactly? Note that this example is a `getIoEventsClient`, since we need now an `AdobeIoEventsHttpClient` as opposed of `getCommerceClient` that creates an `AdobeCommerceHttpClient`.

jcuerdo: You're right, the analogy wasn't accurate. Let me apply the change.

specs/features/CEXT-6270-emit-configured-event.md
obarcelonap: @jcuerdo could you please answer this question. I think this is the pattern we use for the SDK, is there any strong reason to do not do it in this way?

jcuerdo: Done

obarcelonap (approved):

specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -0,0 +1,230 @@
...
+  await publishEvent({
+    client,
+    provider: "order-events",
+    event: "order.created",
+    payload: { orderId: "100000123", total: 149.99 },
+  });
+}
+```
+
+`provider` is the `key` of an event provider declared in `app.commerce.config.ts`. `event`
```
iivvaannxx: `key` is optional, what do we do if it's not declared? We throw I guess?

specs/features/CEXT-6270-emit-configured-event.md
```diff
@@ -0,0 +1,230 @@
...
+the ingress config co-located with the rest of the client config, consistent with how
+`baseUrl` controls the management API endpoint on the same client.
+
+### `publishEvent` in lib-app
+
+A new export from the root entrypoint of `@adobe/aio-commerce-lib-app`:
+
+```ts
+publishEvent<TPayload extends Record<string, unknown>>(params: {
+  client: AdobeIoEventsApiClient;
```
iivvaannxx: I would use `AdobeIoEventsHttpClient` for this, as it's less "restrictive". The helper from `lib-events` should actually take that to be consistent with the overall library so this function should proxy that call. We could pair it with a `getAdobeIoEventsClient` which mirrors the `getCommerceClient`.

iivvaannxx (approved):

PR #540, opened 2026-06-25
CEXT-6367: implement skills release process and promotion workflow

obarcelonap:
## Description

Implements the skills release process defined in the spec PR. Plugin packages gain `package.json` files so changesets can version them, the `publish-public.yml` workflow gains a promotion step that opens a PR in `adobe/skills` on release, and a shared sync script keeps `tile.json` and `.claude-plugin/plugin.json` in sync with the authoritative `package.json` version.

## Related Issue

https://jira.corp.adobe.com/browse/CEXT-6367

## Motivation and Context

Commerce plugins in `adobe/aio-commerce-sdk` have no automated path to `adobe/skills`. Without this, production consumers cannot install them through the stable channel and promotion requires manual file copying with no traceability.

## How Has This Been Tested?

Unit tests cover the full promotion script: plugin version detection, artifact preparation, the full git path (checkout → copy → commit → push → PR create/update), and idempotent reruns. A version-consistency regression test asserts that `package.json`, `tile.json`, and `.claude-plugin/plugin.json` all carry the same version for every plugin.

## Types of changes

## Checklist:


.github/DEVELOPMENT.md
```diff
@@ -516,3 +527,12 @@ If the deploy key ever needs to be rotated:
...
 4. Update the ruleset bypass actor with the new key ID via `gh api`.
+
+The Commerce plugin promotion step also requires an `ADOBE_SKILLS_TOKEN` Actions secret. Provision
+it as a fine-grained GitHub personal access token with the following settings:
+
+- **Resource owner:** `adobe`
+- **Repository access:** `adobe/skills` only
+- **Permissions:** Metadata (read), Contents (read/write), Pull requests (read/write)
+- **Expiration:** 366 days (the maximum); set a calendar reminder to regenerate it before it
+  expires, then update the `ADOBE_SKILLS_TOKEN` Actions secret with the new value.
```
iivvaannxx, line 538: How does this help a developer? Is more like a repo admin concern isn't it?

plugins/commerce/app-management/skills/commerce-app-storage/assets/setup-database.ts
iivvaannxx, line 1: Why this change in this PR?

scripts/source/ci/release/promote-skills.ts
```diff
@@ -0,0 +1,461 @@
...
+import { basename, dirname, join } from "node:path";
+
+import { runGitHubScript } from "./utils.ts";
+
+import type { AsyncFunctionArguments } from "./types.ts";
+
+const SOURCE_REPOSITORY = "adobe/aio-commerce-sdk";
+const TARGET_OWNER = "adobe";
+const TARGET_REPO = "skills";
+const TARGET_REPOSITORY_URL = "https://github.com/adobe/skills";
```
iivvaannxx: Nitpick: this could use the above two constants to construct the same url

.github/workflows/publish-public.yml
```diff
@@ -53,10 +56,43 @@ jobs:
           release-channel: public
           slack-webhook-url: ${{ secrets.SLACK_WEBHOOK_URL }}
 
+      - name: Detect changed Commerce plugin versions
+        id: changed-plugins
+        if: steps.changesets.outputs.hasChangesets == 'false'
```
iivvaannxx, line 66: I'm confused about this condition

scripts/source/ci/release/promote-skills.ts
```diff
@@ -0,0 +1,461 @@
...
+  const changedPackagePaths: string[] = [];
+
+  for (const packagePath of packagePaths) {
+    if (await packageVersionChanged(exec, sourceRepositoryPath, packagePath)) {
+      changedPackagePaths.push(packagePath);
+    }
+  }
+
+  return changedPackagePaths;
+}
```
iivvaannxx, line 159: Doesn't the Changesets Job output produce something that we can use instead of this code?

scripts/source/ci/release/promote-skills.ts
```diff
@@ -0,0 +1,461 @@
...
+    await cp(join(sourcePath, "tile.json"), join(targetPath, "tile.json"));
+    await cp(join(sourcePath, "README.md"), join(targetPath, "README.md"));
+    await cp(join(sourcePath, "skills"), join(targetPath, "skills"), {
+      recursive: true,
+    });
+    await cp(
+      join(sourcePath, ".claude-plugin"),
+      join(targetPath, ".claude-plugin"),
+      { recursive: true },
+    );
```
iivvaannxx: This can probably get a bit cleaner if you use a constant array of files to promote and iterate over it.

plugins/commerce/scripts/sync-plugin-version.mjs
iivvaannxx, line 1: I would suggest moving this to `scripts/`, and then in package.json you can reference it the same way we do with `pre-pack` and `post-pack`.

scripts/source/ci/release/promote-skills.ts
```diff
@@ -0,0 +1,461 @@
...
+  });
+}
+
+async function readJson<T>(path: string) {
+  return JSON.parse(await readFile(path, "utf-8")) as T;
+}
+
+async function writeJson(path: string, value: unknown) {
+  await writeFile(path, `${JSON.stringify(value, null, 2)}\n`);
+}
```
iivvaannxx: If you apply [this comment](https://github.com/adobe/aio-commerce-sdk/pull/540/changes#r3521072032) you will be able to reuse this code. Just FYI

iivvaannxx (commented):
I have a feeling many logic in the `promote-skills` script can probably be removed and repalced by one of the many packages that changesets offers to work with their files. See [their packages](https://github.com/changesets/changesets/tree/main/packages) (many are published to npm and you can pull them in as deps in the `scripts` workspace)

iivvaannxx:
Also, do we want to handle promotion/publish to the Tessl registry?

obarcelonap:
> Also, do we want to handle promotion/publish to the Tessl registry?

Only one set of skills in adobe/skills is published in Tessl registry. From my perspective, the tool is under evaluation, I'd keep it to asses if is useful for us for testing without wasting much effort.
In any case, you can also install directly from a gh repo using `tessl install github:owner/repo`

.github/DEVELOPMENT.md
obarcelonap: I wanted to document this requirement somehow, if you think there's a better placement happy to move it :)

plugins/commerce/app-management/skills/commerce-app-storage/assets/setup-database.ts
obarcelonap: When reviewing the skills, I thought it is dangerous to cleanup potential production data on `uninstall` that's why I just left it empty with some comments.
I'll create a new PR for this you're right is not in the context of this PR.

obarcelonap: here you go https://github.com/adobe/aio-commerce-sdk/pull/553

scripts/source/ci/release/promote-skills.ts
obarcelonap: The changesets job's `publishedPackages` output only contains packages that actually got published to npm, and these plugins are `"private": true`, so changesets excludes them from that list by design

obarcelonap:
> I have a feeling many logic in the `promote-skills` script can probably be removed and repalced by one of the many packages that changesets offers to work with their files. See [their packages](https://github.com/changesets/changesets/tree/main/packages) (many are published to npm and you can pull them in as deps in the `scripts` workspace)

Looked into this TLDR; none of the packages under changesets actually cover what this script needs.

> The core issue: plugins/commerce/app-management and plugins/commerce/app-migration are both "private": true, and changesets' publish path explicitly filters private packages out before publishing (publicPackages = packages.filter(pkg => !pkg.packageJson.private) in @changesets/cli). So there's no changesets output — job output or library — that reports version changes for these plugins; the detection has to happen independently, which is what getChangedPluginPackagePaths/packageVersionChanged does.
> 
> I did consider @changesets/git's getChangedPackagesSinceRef as a drop-in for the git-diff part, but it shells out to real git directly rather than going through the injected exec we use everywhere else in scripts/source/ci/release/*, which would mean rewriting the tests to use real git repos instead of the current mocked-exec fixtures. snapshot.ts already made the same call elsewhere in this directory — shelling out to changesets' documented CLI/action outputs rather than depending on its internal helper packages (which the @changesets/git README itself says aren't meant to be a stable public surface).

scripts/source/ci/release/promote-skills.ts
obarcelonap: done

plugins/commerce/scripts/sync-plugin-version.mjs
obarcelonap: done

scripts/source/ci/release/promote-skills.ts
obarcelonap: done

.github/workflows/publish-public.yml
obarcelonap: Added a comment, basically this condition is when the release pr is merged so no changeset files are present. This is the trigger to inspect if there are diffs in plugins.

scripts/source/ci/release/promote-skills.ts
obarcelonap: done

.github/DEVELOPMENT.md
iivvaannxx: Probably this is more internal information? What about a wiki?

.github/workflows/publish-public.yml
iivvaannxx: Got it, thanks 👍🏻

.github/DEVELOPMENT.md
obarcelonap: I'd prefer to have all the documentation in the repo

iivvaannxx: Then I would create a `MAINTAINERS.md` or something, not write it in `DEVELOPMENT.md`.

iivvaannxx (approved):
Approving. Regarding the Changesets utility packages, I did run Opus to check and it did actually find one potential replacement. Find below what it said.

> Review: new scripts/ code vs. Changesets built-ins
>
> I read the three new source files against the upstream Changesets v3 clone in changesets/. One clear reinvention, a couple of weak/partial overlaps, and one thing that is not reinventable. Details:
> 
> 1. Solid replacement — readChangelogEntries → getChangelogEntry
> 
> promote-skills.ts:295-324 hand-rolls a changelog parser: indexOf("## " + version), slice to the next ## , keep lines starting with - .
> 
> Changesets already ships this as getChangelogEntry(changelog, version) from @changesets/release-utils (changesets/packages/release-utils/src/utils.ts:33), and it's exported from the package root. It's strictly more robust than the local version — it skips fenced code blocks so it won't match ## headings inside example snippets, and it matches the closing heading by depth instead of any ## .
> 
> - Drop: readChangelogEntries (30 lines).
> - Replace with: import { getChangelogEntry } from "@changesets/release-utils" (add it as a devDependency — it's not in the lockfile yet, only @changesets/git is present transitively).
> - Caveat (behavior change): the local fn returns only -  bullet lines as string[]; getChangelogEntry returns { content, highestLevel } where content is the full markdown block (including ### Minor Changes sub-headings). Your buildPromotionPullRequestBody (promote-skills.ts:227) would need to embed content instead of joining a bullet array. The resulting PR body is arguably better (keeps the bump-level subheadings), but it's a visible format change — worth a glance at the promote-skills test snapshots.
>

.github/DEVELOPMENT.md
obarcelonap: Will do. Let me go ahead with this for the moment and I'll create a folow up PR extracting a `MAINTAINERS.md` file looking in all the markdown files we have right now.

obarcelonap:
> Approving. Regarding the Changesets utility packages, I did run Opus to check and it did actually find one potential replacement. Find below what it said.

@iivvaannxx looked into this, the mentioned function it is not stable yet and is part of a major changesets upgrade (3.0.0). Therefore, I just added a note in our function so when the new changesets version is released we revaluate if we can change the impl to rely on the changesets' counterpart.

jcuerdo (approved):

PR #542, opened 2026-06-26
CEXT-6338: Add Admin UI iframe browser helpers (lib-admin-ui /web)

iivvaannxx:
## Description

Adds the browser/React layer for Adobe Commerce Admin UI v2 (`commerce/backend-ui/2`) iframe extensions, exposed through a new `@adobe/aio-commerce-lib-admin-ui/web` entrypoint and re-exported from the meta package as `@adobe/aio-commerce-sdk/admin-ui/web`.

Key pieces:

- **`createExtensionApp({ metadata, routes, root? })`** — bootstraps the iframe app: loads the Experience Cloud shell runtime (with a mock-runtime fallback for local dev), registers and attaches the UIX guest connection, mounts a hash-based router wired into the Spectrum 2 `Provider`, and renders a Spectrum-themed error boundary. Imports `@react-spectrum/s2/page.css` so the `<html>` element is themed on first paint (avoids a flash of unstyled content).
- **React hooks** for host-provided context: `useIms` (IMS credentials, available in both the Commerce Admin and the Experience Cloud shell), `useSharedContext` (Commerce shared context), and the extension-point helpers `useMassActionContext`, `useOrderViewButtonContext`, and `useHostConnection`.
- `react`, `react-dom`, and `@react-spectrum/s2` are declared as **optional `peerDependencies`** of `lib-admin-ui` so only `/web` consumers carry them.
- The meta `@adobe/aio-commerce-sdk` package re-exports the entrypoint; its `tsconfig` (`jsx`, `types`) and optional peers/devDeps were added so it type-checks the React re-export.
- The spec `specs/features/CEXT-6338-admin-ui-v2-web-src-scaffold.md` was updated so it matches the implementation (no drift), kept free of branch-specific implementation details.

> Note: this branch implements the `lib-admin-ui` `/web` layer only. The `lib-app` `generateWebSrc` scaffolding described in the spec is out of scope here and lands separately.

## Related Issue

[CEXT-6338](https://jira.corp.adobe.com/browse/CEXT-6338)

## Motivation and Context

App Builder's `view` operation requires a web entry point to function. Rather than have every app inline the EC shell bootstrap, UIX registration, and shared-context wiring, this library encapsulates that boilerplate so apps get a thin, consistent, Spectrum-2-based starting point and absorb future churn through SDK upgrades instead of manual edits.

## How Has This Been Tested?

- `pnpm typecheck` — passes across all packages. (This PR also fixes a meta-package typecheck failure surfaced by the new web re-export: missing `jsx` and unresolved `@react-spectrum/s2`.)
- `pnpm test` — full suite passes.
- `pnpm build` + `publint` — the `/web` entrypoint correctly externalizes React/Spectrum subpath imports and preserves the side-effectful CSS import.

The web/React layer currently has no unit tests.

n/a

## Types of changes

## Checklist:

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) with guidance from the PR author


obarcelonap (commented):
don't we need to add examples in `usage.md` of `aio-commerce-lib-admin-ui`?

obarcelonap (approved):

packages/aio-commerce-lib-admin-ui/source/web/react/routing/hooks/use-spectrum-router.ts
```diff
@@ -0,0 +1,50 @@
...
+        href: ToOptions | string,
+        options?: Omit<NavigateOptions, keyof ToOptions>,
+      ) => {
+        if (typeof href === "string") {
+          return router.navigate({ to: getRouteTo(href), ...options });
+        }
+
+        return router.navigate({ ...href, ...options });
+      },
+      useHref: (href: ToOptions | string) => {
```
obarcelonap, line 38: The following was spotted by codex

> - TanStack route target: "/bulk-cancel"
> - Browser URL for hash routing: "#/bulk-cancel" or similar
> 
>   getRouteTo("bulk-cancel") only creates the TanStack route target: "/bulk-cancel". That is correct for router.navigate({ to: ... }).
> 
>   But useHref() is supposed to return a real browser href for links. With hash history, that must be built by TanStack:
> 
>   router.buildLocation({ to: "/bulk-cancel" }).href
> 
>   So the fix is: keep using getRouteTo() to normalize the route target, but do not return it directly from useHref().
> 
>   Minimal patch:
> 
>   useHref: (href: ToOptions | string) => {
>     const toOptions = typeof href === "string" ? { to: getRouteTo(href) } : href;
>     return router.buildLocation(toOptions).href;
>   },
> 
>   Why: if useHref("bulk-cancel") returns "/bulk-cancel", a Spectrum link can navigate the iframe document to /bulk-cancel. If it returns TanStack’s built href, the
>   URL respects createHashHistory().

packages/aio-commerce-lib-admin-ui/source/web/react/commerce/hooks/use-guest-connection.ts
```diff
@@ -0,0 +1,42 @@
...
+  extensionId: string,
+): Promise<GuestConnection> {
+  let promise = connectionCache.get(extensionId);
+  if (!promise) {
+    promise = attach({ id: extensionId });
+    promise.catch((err) => {
+      console.error("UIX guest attach failed:", err);
+    });
+
+    connectionCache.set(extensionId, promise);
```
obarcelonap: wouldn't be better to only cache the promise if completed? this way if failed attach would be retried in the next attempt of `getGuestConnectionPromise`

obarcelonap (approved):
did my best on reviewing this one :)

packages/aio-commerce-lib-admin-ui/source/web/react/commerce/hooks/use-guest-connection.ts
iivvaannxx: I will check this, but as far as I know, the [use](https://react.dev/reference/react/use) React API, requires a stable reference/object, otherwise, it can enter an infinite rendering loop. That's why we have this module-level cache. I would need to check if I can maintain that "stable" promise until it resolves (try not to recreate it, something, probably via `useRef`).

packages/aio-commerce-lib-admin-ui/source/web/react/auth/lib.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Resolves IMS credentials from the Commerce shared context.
+ * @param sharedContext - The Commerce shared context, established over the guest connection.
+ */
+export function resolveCommerceImsCredentials(
+  sharedContext: SharedContext["sharedContext"],
+): ImsContext | null {
+  const imsToken = sharedContext.get("imsToken") as string | null;
+  const imsOrgId = sharedContext.get("imsOrgId") as string | null;
+
+  if (!(imsToken && imsOrgId)) {
```
asalloum5, line 27: What if one day we decide to stop sending one the 2? I would return whatever was found not null if one is found and not the other.

packages/aio-commerce-lib-admin-ui/source/web/react/auth/lib.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Resolves IMS credentials from the Experience Cloud shell configuration.
+ * @param shellConfiguration - The Experience Cloud shell configuration, if available.
+ */
+export function resolveShellImsCredentials(
+  shellConfiguration: ShellConfiguration | null,
+): ImsContext | null {
+  const imsToken = shellConfiguration?.imsToken ?? null;
+  const imsOrgId = shellConfiguration?.imsOrg ?? null;
+
+  if (!(imsToken && imsOrgId)) {
```
asalloum5: Same here

packages/aio-commerce-lib-admin-ui/source/web/react/extension/error-boundary.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+function getErrorMessage(error: unknown): string {
+  if (error instanceof Error) {
+    return error.message;
+  }
+
+  return typeof error === "string" ? error : "An unexpected error occurred.";
+}
+
+function ErrorFallback({ error, resetErrorBoundary }: Readonly<FallbackProps>) {
```
asalloum5, line 37: Do we really want to introduce and maintain this? I think this should be handled by the app developer

packages/aio-commerce-lib-admin-ui/source/web/react/extension/standalone-app.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  useEffect(() => {
+    syncRootColorScheme(undefined);
+  }, []);
+
+  return (
+    <Provider colorScheme={undefined} router={spectrumRouter}>
+      <ExtensionErrorBoundary>
+        <ImsContextProvider credentials={null}>
+          <Outlet />
```
asalloum5: Can we rename this to something more readable? I don't understand what Outlet means in this context

packages/aio-commerce-lib-admin-ui/source/web/react/auth/lib.ts
iivvaannxx: This is not on us to decide (whether `imsToken` or `imsOrg` are sent in here), since this branch is handled when we are in the experience cloud. As a matter of fact this is probably stale since now we clearly branch depending on where we are running (Commerce, Experience, Standalone), if this is only executed in the experience branch, the `null` fallbacks don't make sense. I'll check.

packages/aio-commerce-lib-admin-ui/source/web/react/auth/lib.ts
iivvaannxx: This is required to align with the experience branch, since they both provide these via context through the same `useIms`. Doing what you say would mean that we need a `useCommerceIms` and a `useExperienceIms` because the credentials will be typed differently ([see also this comment](https://github.com/adobe/aio-commerce-sdk/pull/542/changes/BASE..c9e3dc08ef70064e5388535deb71232292e70c59#r3512062427) as per why we can't do that change in the experience branch). 

I get your point, as it would require less maintenance, but it's a change that ends up introducing new maintenance surface (new helpers), so I don't see an immediate value in it. Right now, this is our contract. If that changes in the future, we should update the library accordingly, but I don't think it's worth doing this today. I'll be skipping. Unresolve if you have a major concern.

packages/aio-commerce-lib-admin-ui/source/web/react/extension/standalone-app.tsx
iivvaannxx: `Outlet` is not a component of us. I agree it's definitely not the best name though. I'll rename it via an import alias 👍🏻

packages/aio-commerce-lib-admin-ui/source/web/react/extension/error-boundary.tsx
iivvaannxx: The error boundary is definitely something that I would keep, as it helps show errors like the ones below when users misuse the library helpers. As for the content, I could agree that it might not be necessary to render an illustration, but I don't think it hurts maintenance either; it just renders a nicer UI. 

If in the future something comes to us saying they don't want it, we can introduce a way to customize it, but for now I don't think it's necessary, at least not in this PR.

Image

Image

jcuerdo (approved):

packages/aio-commerce-lib-admin-ui/source/web/react/commerce/hooks/use-guest-connection.ts
iivvaannxx: You're right, and it was worse than it looks: a failed attach stayed in the cache, so one flaky connection meant a reload was the only way back.

On caching only once it's complete, that's the one thing we can't do here. `use()` needs the pending promise to keep a stable reference while it's suspended, so if we don't cache until it resolves, every render hands use a brand-new promise and Suspense loops forever. What works is eagerly caching the entry and dropping it when it's rejected, so the next attempt starts fresh. That's what I went with.

I moved the logic into a `createRetryablePromiseCache` helper (memoize while pending/resolved, evict on rejection with an identity guard) and reused it for the other two establishment calls that had the same problem (`register` and `getCommerceHost`).

See changes in [`fbd29da`](https://github.com/adobe/aio-commerce-sdk/pull/542/commits/fbd29da6cc6dc60e09a48e06c21cd42b1242f7f9)

packages/aio-commerce-lib-admin-ui/source/web/react/auth/lib.ts
iivvaannxx: I checked and indeed `imsOrg` and `imsToken` come straight from exc-app's `RuntimeConfiguration`, and both are typed as plain string (no undefined), so once a shell configuration exists, the credentials are always there. The per-field `??` null was dead against that contract, so I removed it.

I did keep the top-level `if (!shellConfiguration) return null;`, because that null is real and has nothing to do with login.

The Commerce branch is a different case: there, the values come from `sharedContext.get(...)` as `string | null`, an untyped bag, so its truthy guard is doing actual work. That's why the two resolvers don't look the same.

See changes in [`e6afd32`](https://github.com/adobe/aio-commerce-sdk/pull/542/commits/e6afd32c30a02288274588182820afb9085904c1)

packages/aio-commerce-lib-admin-ui/source/web/react/extension/standalone-app.tsx
iivvaannxx: Renamed to `ActiveRoute` in [`950e8c1`](https://github.com/adobe/aio-commerce-sdk/pull/542/commits/950e8c10149d75344c869661ce4b302a4660296a)

packages/aio-commerce-lib-admin-ui/source/web/react/routing/hooks/use-spectrum-router.ts
iivvaannxx: Fixed in [`42f02d7`](https://github.com/adobe/aio-commerce-sdk/pull/542/commits/42f02d71101295e11ae9631d0533f06035412835)

iivvaannxx:
@asalloum5 @obarcelonap Addressed/answered all your comments, I'm merging this one. If any of you have any other concerns after the fixes/resolutions, let's take them in a separate PR I'll be open to discussion in Slack.

PR #548, opened 2026-06-30
CEXT-6338: implement Admin UI web-src codegen

iivvaannxx:
## Description

Adds Admin UI web-src codegen to lib-app, including:

- Generates the Admin UI web source scaffold for view-based backend UI extensions.
- Adds build-time variables for scaffold dependency versions.
- Wires package.json, Parcel, template, docs, and typecheck support for generated web-src apps.
- Updates generated action templates to import app config through .

## Related Issue

CEXT-6338

## Motivation and Context

This is stacked on the scripting-utils dependency helper PR and adds the lib-app-specific codegen changes needed for Admin UI v2 web-src scaffolding.

## How Has This Been Tested?

Not run by me after the final stacked-branch cleanup. This branch includes updated integration/unit coverage for the generated web-src behavior.

N/A

## Types of changes

## Checklist:


iivvaannxx (commented):
Note: I'll implement more tests tomorrow

packages/aio-commerce-lib-app/source/commands/generate/actions/templates/admin-ui/web-src/src/components/welcome.jsx
```diff
@@ -0,0 +1,14 @@
+import { useIms } from "@adobe/aio-commerce-lib-admin-ui/web";
+
+/** A simple welcome component that displays the IMS token and org ID. */
+export function Welcome() {
+  const { imsToken, imsOrgId } = useIms();
+  return (
+    <>
+      <h1>Welcome to your Adobe Commerce App</h1>
+
+      <p>IMS Token: {`${imsToken.slice(0, 10)}...`}</p>
```
obarcelonap: Would just show is present (undefined check) instead of first 10 chars.

packages/aio-commerce-lib-admin-ui/source/mass-actions/worker/schema.ts
```diff
@@ -33,8 +33,11 @@ export const MassActionGridTypeSchema = v.picklist([
 export const MassActionRequestSchema = v.object({
   requestId: nonEmptyStringValueSchema("requestId"),
   gridType: MassActionGridTypeSchema,
-  ids: v.pipe(
+  selectedIds: v.pipe(
```
obarcelonap, line 36: superseded by #549

packages/aio-commerce-lib-app/source/commands/hooks/pre-app-build.ts
```diff
@@ -88,7 +89,18 @@ export async function run(extension: Extension, templatesDir = TEMPLATES_DIR) {
...
   if (extension === "backend-ui/2") {
     if (hasAdminUi(appManifest)) {
-      await updateExtConfig(appManifest, BACKEND_UI_V2_EXTENSION_POINT_ID);
+      const extConfig = await updateExtConfig(
+        appManifest,
+        BACKEND_UI_V2_EXTENSION_POINT_ID,
+      );
+
+      if (extConfig.operations?.view) {
+        await generateWebSrc(
```
obarcelonap, line 100: don't we need to call `prepareWebSourceImportAlias` here too? there's jsx code using `#web` alias

obarcelonap (commented):
Agent (codex) flagged the following issue, please take a look

> [CRITICAL] packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts:191-196, packages/aio-commerce-lib-app/source/commands/generate/actions/templates/app-management/app-config.js.template:6, packages/aio-commerce-lib-app/source/commands/generate/actions/templates/app-management/installation.js.template:5, packages/aio-commerce-lib-app/source/commands/generate/actions/templates/business-configuration/config.js.template:6
>
> Static mode maps the `#app.commerce.config` import alias directly to the generated `app.commerce.manifest.json` file (`prepareStaticAppConfigImportAlias` → `getManifestPath()`), and the generated action templates do `import x from "#app.commerce.config"` with no `with { type: "json" }` attribute. Node's ESM loader requires that attribute for `.json`-extension imports — confirmed locally, it throws `ERR_IMPORT_ATTRIBUTE_MISSING`. Every statically generated runtime action (app-config, installation, business-configuration) will fail to load.
>
> Note the dynamic path already avoids this: `writeJavaScriptAppConfigModule` (lib.ts:122-144) writes a `.js` wrapper that re-exports the config as ESM specifically so nothing imports raw JSON. The static path should do the same instead of pointing the alias straight at the `.json` manifest — e.g. have `prepareStaticAppConfigImportAlias` write a small `.js` module that imports the manifest with `type: "json"` and re-exports it, mirroring the dynamic branch, rather than adding the import attribute to every template.
>
> Also worth adding an integration test that actually executes/imports a generated static action rather than just asserting the alias exists in `package.json` — that's what let this slip through.

packages/aio-commerce-lib-admin-ui/source/mass-actions/worker/schema.ts
iivvaannxx: Reverted in [`b50c2bf`](https://github.com/adobe/aio-commerce-sdk/pull/548/commits/b50c2bf322ab18e514d21d048dc0d66be55e30f4)

packages/aio-commerce-lib-app/source/commands/hooks/pre-app-build.ts
iivvaannxx: Good catch! Will add

iivvaannxx:
@obarcelonap I think the import attribute is not always mandatory. I was able to make it work with some config I now don't remember. Bundlers are able to handle this.

But it's true that it's better to have it to avoid clumsy workarounds. I will add a small patch after the code generation to add the import attribute when the import alias points to a JSON instead of a module.

packages/aio-commerce-lib-app/source/commands/generate/actions/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+export const CUSTOM_SCRIPTS_MAP_PLACEHOLDER = "// {{CUSTOM_SCRIPTS_MAP}}";
+export const CUSTOM_SCRIPTS_LOADER_PLACEHOLDER = "// {{CUSTOM_SCRIPTS_LOADER}}";
+
+/** File extensions that require bundling or TypeScript web-src templates. */
+export const TYPESCRIPT_CONFIG_EXTENSIONS = new Set([".ts", ".mts", ".cts"]);
+
+export const JSX_FILE_EXTENSION = ".jsx";
+export const TSX_FILE_EXTENSION = ".tsx";
+
+export const WEB_SOURCE_HTML_ENTRYPOINT = "index.html";
```
jcuerdo: Where  WEB_SOURCE_HTML_ENTRYPOINT and WEB_SOURCE_ENTRYPOINT are going to be used?

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
```diff
@@ -318,6 +278,271 @@ export async function updateExtConfig(
...
+    let content = await readFile(sourcePath, "utf-8");
+
+    if (extension === "tsx") {
+      content = content.replaceAll(
+        `${JSX_FILE_EXTENSION}"`,
+        `${TSX_FILE_EXTENSION}"`,
+      );
+    }
+
+    await writeFile(targetPath, content, { encoding: "utf-8", flag: "wx" });
```
jcuerdo, line 536: A bit picky, "wx" can leave a half-written scaffold if any src/ file exists but index.html doesn't — the copy crashes mid-way, index.html gets written, and the guard permanently skips on the next run. Worth 
  checking for existing files upfront before starting the copy.

Image

jcuerdo (approved):
LGTM but please check my comments

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
iivvaannxx: I actually experienced that, and I thought it was good behavior. The gate that controls whether we generate this is whether the entry point is present (99% of the time, this will be `index.html`). 

So if a scaffold exists but without the entrypoint, it's good if it crashes, because we prevent overriding things we probably shouldn't override. What do you think?

packages/aio-commerce-lib-app/source/commands/generate/actions/constants.ts
iivvaannxx: I used those in the past to do a special string `replace` that I ended up not needing, this may have become stale after that. I'll check and remove if not used. 👍🏻

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
jcuerdo: Honestly, I don't have a strong opinion, for sure, it's a very very strange case. It's just a picky comment, if you preffer to crash in that case, it's ok for me.

iivvaannxx: I'll keep it 👍🏻. We can always evaluate in the future if there's better behavior.

packages/aio-commerce-lib-app/source/commands/generate/actions/constants.ts
iivvaannxx: Indeed they were stale. Removed them in [`f49238b`](https://github.com/adobe/aio-commerce-sdk/pull/548/commits/f49238b008ef1a6ab645bf2192ce8f2fcbbc9cb1)

packages/aio-commerce-lib-app/source/commands/generate/actions/templates/admin-ui/web-src/src/components/welcome.jsx
iivvaannxx: I was going to apply this, but the wiring in `lib-admin-ui` ensures it's populated, so an error is thrown if you use `useIms` outside of the expected environments (e.g., in standalone mode) which always populate them. Then the `undefined` check doesn't make much sense. 

If you have any other suggestions, I'll be happy to consider any (maybe just leave the org?), otherwise I'll skip.

packages/aio-commerce-lib-admin-ui/source/mass-actions/worker/schema.ts
```diff
@@ -33,11 +33,8 @@ export const MassActionGridTypeSchema = v.picklist([
 export const MassActionRequestSchema = v.object({
   requestId: nonEmptyStringValueSchema("requestId"),
   gridType: MassActionGridTypeSchema,
-  selectedIds: v.pipe(
+  ids: v.pipe(
```
obarcelonap: @iivvaannxx it seems you're going to discard the changes done in #549 
probably because you reverted the commit while you should drop it

packages/aio-commerce-lib-app/source/commands/generate/actions/templates/admin-ui/web-src/src/components/welcome.jsx
obarcelonap: in that case I'd just keep the org value in the html, maybe in a human friendly wording (eg. your current org is XX)

iivvaannxx:
@obarcelonap About the import attribute, changes are in [`1d262fe`](https://github.com/adobe/aio-commerce-sdk/pull/548/commits/1d262fe38877db81bd85a08af31739701c862551)

packages/aio-commerce-lib-admin-ui/source/mass-actions/worker/schema.ts
iivvaannxx: Oops, yeah I did the revert and upon syncing with main it caused this. Fixed in [`9ce2965`](https://github.com/adobe/aio-commerce-sdk/pull/548/commits/9ce29659109cc49375b950d9748c9c21fbefe0d8)

obarcelonap (approved):

packages/aio-commerce-lib-app/source/commands/generate/actions/templates/admin-ui/web-src/src/components/welcome.jsx
iivvaannxx: Done in [`db858c0`](https://github.com/adobe/aio-commerce-sdk/pull/548/commits/db858c0029c47284fe00342ec87011cc6fbd6eac)

packages/aio-commerce-lib-app/source/commands/hooks/pre-app-build.ts
iivvaannxx: Fixed in [`cfbf77b`](https://github.com/adobe/aio-commerce-sdk/pull/548/commits/cfbf77b0f1f07176c5a301fa33e744bda365ed9e)

PR #551, opened 2026-07-02
CEXT-6433: enable admin ui sdk before extension registration

asalloum5:
## Types of changes

## Checklist:

### DIFF
--- a/.changeset/enable-admin-ui-sdk.md
+++ b/.changeset/enable-admin-ui-sdk.md
@@ -1,6 +1,0 @@
----
-"@adobe/aio-commerce-lib-admin-ui": minor
-"@adobe/aio-commerce-lib-app": minor
----
-
-Add `enableAdminUiSdk` to the Admin UI API client, which enables the Admin UI SDK in Commerce via `PUT /V1/adminuisdk/config`. The app installation flow now enables the SDK before registering the extension, so Admin UI extensions are served correctly after install.
--- a/.changeset/enable-admin-ui-sdk.md
+++ b/.changeset/enable-admin-ui-sdk.md
@@ -1,0 +1,6 @@
+---
+"@adobe/aio-commerce-lib-admin-ui": minor
+"@adobe/aio-commerce-lib-app": minor
+---
+
+Add `enableAdminUiSdk` to the Admin UI API client, which enables the Admin UI SDK in Commerce via `PUT /V1/adminuisdk/config`. The app installation flow now enables the SDK before registering the extension, so Admin UI extensions are served correctly after install.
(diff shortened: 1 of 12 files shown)


obarcelonap (approved):

.changeset/enable-admin-ui-sdk.md
iivvaannxx, line 1: These need to be two separate changesets. Otherwise we will get a changelog entry in `lib.app` saying:

> Add `enableAdminUiSdk` to the Admin UI API client

Which is not a change related to that lib.

packages/aio-commerce-lib-admin-ui/docs/usage.md
```diff
@@ -36,6 +36,10 @@ const client = createAdminUiApiClient({
   },
 });
 
+// Enable the Admin UI SDK. This must run before registering an extension,
+// otherwise Commerce will not serve the registered extension.
+await client.enableAdminUiSdk();
```
iivvaannxx, line 41: Shouldn't this be better modeled as a `toggle` with a boolean parameter?. The SDK should not implement specific use cases, but give tools to the developers to do any kind of operation.

obarcelonap: I like `enableAdminUiSdk` since it is more idiomatic.
We can create `disableAdminUiSdk` but there's no use case for that so we can skip for the moment.

iivvaannxx: mmm, well, I don't have a strong opinion, but it doesn't align with `lib-events` since we don't do that in the `PUT` for the eventing configuration.

asalloum5: the use cases are different I would say, I like to keep it this way since it's clear what it does and I agree we can add a `disableAdminUiSdk` if needed in the future

iivvaannxx: > the use cases are different I would say

Are different now because the /config endpoint only has that boolean, so you can treat the whole operation as enable/disable, but if in the future there is an extra setting, this will not "scale"/adapt well, and we'll need a separate `updateConfig`-like operation, which is what, in the end, we're doing here. 

When that happens, either we will have duplicate operations (`updateConfig` will also be able to enable/disable), or we will require a deprecation/major

Not saying it's a blocker or unacceptable, but it's a trade-off we're making here that I wanted to raise.

.changeset/enable-admin-ui-sdk.md
iivvaannxx, line 1: Delete the old one!

iivvaannxx (approved):

jcuerdo (approved):

packages/aio-commerce-lib-admin-ui/docs/usage.md
asalloum5: I don't think we'll have other config that will be open for Admin UI SDK, at least I don't see any for now. I would keep it this way and do the changes if needed in the future.

PR #559, opened 2026-07-06
chore: upgrade ultracite to 7.9.2 and fix newly-enabled lint rules

obarcelonap:
## Description

Upgrades `ultracite` (the Biome preset this repo's `biome.jsonc` extends) from the temporary 7.8.3 pin back to 7.9.2, and fixes every violation the newly-enabled rules surface repo-wide. Safe formatting/sorting violations were auto-fixed; the remainder (noShadow, noAwaitInLoops, noUnnecessaryConditions, useErrorCause, useDestructuring, and a few others) were fixed by hand, preserving existing behavior. A handful of previously-sequential loops were parallelized where iterations were genuinely independent; loops with ordering, abort-on-first-failure, or external API rate-limit sensitivity were kept sequential with a `biome-ignore` explaining why.

## Related Issue

N/A

## Motivation and Context

`ultracite` 7.9.0 enabled a large batch of new Biome lint/assist rules, which broke CI on #554 (a routine lock-file-maintenance PR). That PR pinned `ultracite` to 7.8.3 so it could merge safely, deferring the rule adoption to this dedicated follow-up.

## How Has This Been Tested?

Ran `pnpm check:ci`, `pnpm typecheck`, `pnpm test`, and `pnpm build` across the whole repo — all green. Also ran a targeted correctness review of every behavior-changing fix (loop parallelization, removed conditionals) against call sites and storage/type guarantees before merging.

N/A

## Types of changes

## Checklist:


obarcelonap:
Note: \`pnpm-workspace.yaml\` has a \`minimumReleaseAgeExclude\` entry for \`ultracite@7.9.2\`, added automatically by \`pnpm install\` because that version was published within pnpm's 24h supply-chain minimum-release-age window at the time. It's currently required — removing it makes \`pnpm install\` (and CI) fail with \`ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION\`.

This should be removed before merging, once \`ultracite@7.9.2\`'s publish timestamp clears the 24h window (published 2026-07-06T05:20:10Z, so safe to remove after ~2026-07-07T05:20:10Z). At that point `pnpm install` will no longer need the exclusion and it can just be deleted.

package.json
```diff
@@ -50,7 +50,7 @@
     "typedoc": "^0.28.17",
     "typedoc-plugin-markdown": "^4.10.0",
     "typescript": "catalog:",
-    "ultracite": "7.8.4",
+    "ultracite": "^7.9.2",
```
iivvaannxx: I think ultracite stays better pinned. Otherwise, you and I may pull different versions, and we can get this kind of messy difference without understanding where it comes 🫠

obarcelonap: I think this can't happen since the lockfile is committed and contains the exact versions.
If we pin, the difference is that renovates' updating the lock file can't bump the dep so it will be bumped as part of non-major renovate's config. 
Personally I think that's the same but if you think that's better I can pin, no strong opinion.

iivvaannxx: True, totally forgot about the lockfile lol 🤣 . Ignore the comment

iivvaannxx (approved):

PR #561, opened 2026-07-08
CEXT:6270: Enable Runtime Actions to Emit Configured Events by Provider and Event Name

jcuerdo:
## How Has This Been Tested?

To test it, there are a couple of actions:
- **Emit an external event:** https://development-271530-956ochrerattlesnake-cext6270new.adobeioruntime.net/api/v1/web/my-package/emit-event
- **Emit a commerce event:** https://development-271530-956ochrerattlesnake-cext6270new.adobeioruntime.net/api/v1/web/my-package/emit-event-commerce

There is another action that receives handles the event and log the result:
- https://development-271530-956ochrerattlesnake-cext6270new.adobeioruntime.net/api/v1/web/my-package/handle-event

And the logs are visible in https://developer-stage.adobe.com/console/projects/271530/4566206088344893560/workspaces/4566206088344914831/logs

image

## Types of changes

## Checklist:


.changeset/crisp-candles-lie.md
iivvaannxx, line 1: I find myself writing this review comment quite many times already, please try to memorize 😄 

These need to be separate changesets, you are mixing a lot of unrelated messages in a single changelog entry. This will produce a message in `aio-commerce-lib-app` that says that `lib-events` gains a `publishRawEvent` and that `lib-api` gains an `ingressBaseUrl`, which is completely unrelated.

Each changeset should be scoped to a single package, UNLESS the change is virtually equivalent from the user's perspective (that usually only happens with the metapackage). Also, implementation details like the one below should be omitted.

> Provider IDs and event codes are resolved from metadata written to system storage at installation time — no management API round-trip at runtime.

packages/aio-commerce-lib-admin-ui/docs/usage.md
iivvaannxx, line 1: This change seems accidental

packages/aio-commerce-lib-api/docs/api-reference/index/type-aliases/CommerceHttpClientConfig.md
iivvaannxx, line 1: API references should not be regenerated in normal PRs, they are regenerated during release, automatically.

packages/aio-commerce-lib-api/docs/usage.md
iivvaannxx, line 1: Probably accidental change also?

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -1047,6 +1025,92 @@ A plain `aio app deploy` on its own does not add the action: the `pre-app-build`
 
 For an app that was already associated under the older SDK, re-associate it after redeploying so the store call runs and backfills the instance data — a redeploy alone does not populate data for an existing association.
 
+### Emitting Configured Events from Runtime Actions
+
+Runtime actions can publish a custom I/O Event by referencing the provider and event exactly as declared in the `eventing` section of `app.commerce.config.ts` — no need to know the I/O Events provider UUID, the fully-qualified event code, or the CloudEvents envelope format.
+
+At installation time the SDK writes each configured provider's I/O Events ID and its event codes to system storage. `publishEvent` reads that metadata to resolve the provider key and event name at runtime, so there is no management API round-trip on the hot path.
```
iivvaannxx: These kind of notes/implementation details are not very useful in public documentation. I'd omit.
```suggestion
Runtime actions can publish a custom I/O Event by referencing the provider and event exactly as declared in the `eventing` section of `app.commerce.config.ts`. At installation time, the SDK writes each configured provider's I/O Events ID and event codes to system storage. `publishEvent` resolves those automatically by the given key and publishes the event.
```

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -1047,6 +1025,92 @@ A plain `aio app deploy` on its own does not add the action: the `pre-app-build`
...
+} catch (error) {
+  if (error instanceof PublishEventError) {
+    // Handle any publish-event failure (or narrow with the specific subclasses).
+  }
+  throw error;
+}
+```
+
+> [!NOTE]
+> To target a non-production ingress endpoint, pass `config.ingressBaseUrl` when creating the client. The lower-level `publishRawEvent` (from `@adobe/aio-commerce-lib-events`) is available when you already have a resolved provider ID and event code; see the [`@adobe/aio-commerce-lib-events` usage guide](../../aio-commerce-lib-events/docs/usage.md).
```
iivvaannxx: Omit this, in the end the only ones we need this its us, for Stage. Users reading this will always use prod.

packages/aio-commerce-lib-app/source/management/installation/events/external.ts
```diff
@@ -92,8 +104,28 @@ async function createExternalEvents(
...
   }
 
+  const existing = (await getSystemConfigByKey<StoredEventsData>(
+    EVENTS_STORAGE_KEY,
+  )) ?? {
+    providers: {},
+  };
+  await setSystemConfigByKey(EVENTS_STORAGE_KEY, {
+    providers: { ...existing.providers, ...storedProviders },
+  });
```
iivvaannxx, line 137: Nitpick: this overrides the existing providers, even if they are the same or empty. The underlying API round-trip could be avoided in some situations.

packages/aio-commerce-lib-app/source/management/installation/events/utils.ts
```diff
@@ -41,6 +41,20 @@ const PROVIDER_TYPE_TO_LABEL = {
...
+ * slugified `label`. This is the single source of truth for the key: it is used
+ * both when persisting provider metadata at install time and when composing the
+ * I/O Events `instance_id`, so runtime lookups always match what was stored.
+ *
+ * @param provider - The event provider to derive the key from.
+ */
+export function getProviderStorageKey(provider: EventProvider) {
+  return provider.key ?? provider.label.toLowerCase().replace(/\s+/g, "-");
+}
+
```
iivvaannxx: I think it's better to make the `key` a requirement for this feature. Being able to also match by label introduces a second way of doing the same I don't think we actually need and forces users to know internal details of the library (how slugifcation works) to be able to match.

It's probably better to just say "if you want to publish an event, add a key".

packages/aio-commerce-lib-events/test/integration/io-events-api.test.ts
```diff
@@ -161,6 +161,82 @@ describe("Adobe IO Events API - Integration Tests", () => {
...
+      });
+
+      expect(capture.body).not.toBeNull();
+      expect(capture.body?.specversion).toBe("1.0");
+      expect(capture.body?.source).toBe(`urn:uuid:${TEST_PROVIDER_ID}`);
+      expect(capture.body?.type).toBe(TEST_EVENT_CODE);
+      expect(capture.body?.datacontenttype).toBe("application/json");
+      expect(capture.body?.data).toEqual(payload);
+      expect(capture.body?.id).toBeTypeOf("string");
+      expect(capture.body?.time).toBeTypeOf("string");
```
iivvaannxx: Tip: use `expect.assert(capture.body)` instead of `not.toBeNull()` and you'll no longer need those `?` in calls like `body?.`.

packages/aio-commerce-lib-events/source/io-events/api/event-ingress/endpoints.ts
```diff
@@ -0,0 +1,71 @@
...
+  params: PublishRawEventParams<TPayload>,
+): Promise<void> {
+  const { providerId, eventCode, payload } = params;
+
+  const cloudEvent = {
+    specversion: "1.0",
+    id: globalThis.crypto.randomUUID(),
+    source: `urn:uuid:${providerId}`,
+    type: eventCode,
+    time: new Date().toISOString(),
```
iivvaannxx, line 58: This field `time` is not documented [here](https://developer.adobe.com/events/docs/guides/api/eventsingress-api), from where did it came from? Same as `datacontenttype`

packages/aio-commerce-lib-events/source/io-events/api/event-ingress/endpoints.ts
```diff
@@ -0,0 +1,71 @@
...
+    prefixUrl: httpClient.config.ingressBaseUrl,
+  });
+
+  await ingressClient.post("", {
+    body: JSON.stringify(cloudEvent),
+    // The management client defaults to Accept: application/hal+json; the ingress
+    // requires application/cloudevents+json and application/json respectively.
+    headers: {
+      "content-type": "application/cloudevents+json",
+      Accept: "application/json",
```
iivvaannxx, line 73: The `phidata` header should also be sent when the event in `lib-app` has the `hipaa_audit_required`, as far as I understand from the Event Ingress API docs. Therefore, you need here a way to set that header.

packages/aio-commerce-lib-events/docs/usage.md
```diff
@@ -261,6 +261,32 @@ await ioEventsClient.deleteRegistration({
 });
 ```
 
+#### Publishing Events to the Ingress
+
+`publishRawEvent` sends an event to the Adobe I/O Events ingress. It builds a CloudEvents 1.0 envelope from the resolved `providerId` and `eventCode` you supply and POSTs it to the ingress endpoint configured on the client. Authentication is applied automatically from the client's IMS auth.
```
iivvaannxx: Why did this change from `publishEvent` to `publishRawEvent`? 🤔

packages/aio-commerce-lib-events/docs/usage.md
```diff
@@ -261,6 +261,32 @@ await ioEventsClient.deleteRegistration({
...
+```typescript
+const ioEventsClient = createAdobeIoEventsApiClient({
+  auth: {/* IMS auth params */},
+  config: { ingressBaseUrl: "https://your-ingress-endpoint" },
+});
+```
+
+> [!TIP]
+> If you are emitting an event declared in your `app.commerce.config.ts`, prefer `publishEvent` from [`@adobe/aio-commerce-lib-app`](../../aio-commerce-lib-app/docs/usage.md#emitting-configured-events-from-runtime-actions). It resolves the provider ID and event code for you from the app configuration, so you only pass the provider key and event name. Use `publishRawEvent` when you already have the resolved provider ID and event code.
+
```
iivvaannxx: Some notes:

1. I'd skip documenting the configuration of different ingress endpoints; users will only need production.
2. `lib-events` doesn't know anything about `lib-app`, only the other way around, remove the `!TIP`.
3. Probably implementation details such as how the payload is formed can be skipped, but I don't have a strong opinion.

packages/aio-commerce-lib-app/source/management/installation/events/stored-events-data.ts
iivvaannxx, line 1: I have a feeling this file is kind of unnecessary? Why not put the types in `events/types.ts` and the constant as a `getEventsStorageKey` helper somewhere else?

iivvaannxx (commented):
@jcuerdo The PR has quite some noise with modifications to unrelated files (probably formatter? not sure what caused them) and also changes to the API reference which we avoid in PRs since it's auto-generated in release, could you remove those?

packages/aio-commerce-lib-admin-ui/docs/usage.md
jcuerdo: Done eef34043

packages/aio-commerce-lib-api/docs/usage.md
jcuerdo: Done eef34043

.changeset/crisp-candles-lie.md
jcuerdo: Done 7e216621

packages/aio-commerce-lib-api/docs/api-reference/index/type-aliases/CommerceHttpClientConfig.md
jcuerdo: Done e0a9595a

packages/aio-commerce-lib-app/docs/usage.md
jcuerdo: Done 7ea462a1

packages/aio-commerce-lib-events/source/io-events/api/event-ingress/endpoints.ts
jcuerdo: Done 7e216621

packages/aio-commerce-lib-events/docs/usage.md
jcuerdo: This was not in the spec, but the name divergence is intentional; having **publishEvent** at two abstraction levels creates import ambiguity.  So I decided to use a different name. Don't have a strong opinion here; if you have strong arguments to use the same name for both, I can change it. What do you think?

packages/aio-commerce-lib-app/source/management/installation/events/stored-events-data.ts
jcuerdo: Done 7e216621

packages/aio-commerce-lib-events/docs/usage.md
jcuerdo: Done 7ea462a1

packages/aio-commerce-lib-events/docs/usage.md
iivvaannxx: For me it's confusing because it sounds as if you were publishing a "different kind" of event, when in the end one is just a wrapper of the other, so they do the same. Import ambiguity should only happen in the file where the `lib-app` helper is defined and the `lib-events` one is used, since it's the only case when we have the two in the same scope (other cases make not much sense, users should only need one). 

For that specific case I think an import alias for the imported symbol from `lib-events` works better.

packages/aio-commerce-lib-app/source/access/events.ts
```diff
@@ -0,0 +1,101 @@
...
+ * ```
+ */
+export async function publishEvent<
+  TPayload extends Record<string, unknown> = Record<string, unknown>,
+>(params: {
+  client: AdobeIoEventsApiClient;
+  provider: string;
+  event: string;
+  payload: TPayload;
+  hipaaAuditRequired?: boolean;
```
iivvaannxx: This data, although known by the developer, is configured through the event in the `app.commerce.config`, if you have the event data stored, take it from the instead from re-asking it here. Better DX.

.changeset/afraid-cobras-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": minor
+---
+
+New `publishEvent` export for publishing a configured I/O Event from a runtime action by provider key and event name as declared in `app.commerce.config.ts`. Provider and event resolution happens automatically from metadata stored at installation time. Accepts an optional `hipaaAuditRequired` flag to mark events containing PHI data for HIPAA compliance.
```
iivvaannxx:
```suggestion
New `publishEvent` export for publishing a configured I/O Event from a runtime action by provider key and event name as declared in `app.commerce.config.ts`. Provider and event resolution happens automatically from metadata stored at installation time.
```

See https://github.com/adobe/aio-commerce-sdk/pull/561/changes#r3550625089

.changeset/dry-balloons-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-events": minor
+---
+
+New `publishRawEvent` export on `AdobeIoEventsApiClient` for publishing a CloudEvents 1.0 event directly to the Adobe I/O Events ingress using a resolved provider ID and event code. Accepts an optional `hipaaAuditRequired` flag to mark events containing PHI data for HIPAA compliance.
```
iivvaannxx:
```suggestion
New `publishRawEvent` export on `AdobeIoEventsApiClient` for publishing a CloudEvents 1.0 event directly to the Adobe I/O Events ingress using a given provider ID and event code.
```

Implementation detail.

packages/aio-commerce-lib-events/source/io-events/api/event-ingress/endpoints.ts
jcuerdo: Both time and datacontenttype are optional attributes defined in the CloudEvents 1.0 spec.

-  CloudEvents 1.0 spec optional attributes: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#optional-attributes

packages/aio-commerce-lib-app/source/management/installation/events/types.ts
```diff
@@ -117,3 +117,22 @@ export type OffboardEventsParams = {
...
+/**
+ * Shape of the `system.events` entry written to system storage at installation time.
+ * Keyed by `provider.key`.
+ */
+export type StoredEventsData = {
+  providers: Record<string, StoredProviderEntry>;
+};
+
+/** Storage key used for the events installation data. */
+export const EVENTS_STORAGE_KEY = "events";
```
iivvaannxx: Nitpick, "source" in a `types.ts` file.

packages/aio-commerce-lib-events/source/io-events/api/event-ingress/endpoints.ts
```diff
@@ -0,0 +1,78 @@
...
+  /** The fully-qualified event code. */
+  eventCode: string;
+  /** The event payload. Must be a JSON object. */
+  payload: TPayload;
+  /**
+   * When `true`, marks the event as containing Protected Health Information (PHI)
+   * by sending `x-event-phidata: true` to the ingress. Required for HIPAA compliance.
+   * Defaults to `false` when omitted.
+   */
+  hipaaAuditRequired?: boolean;
```
iivvaannxx: In this library I would honor the actual name expected by the API, not the one we configure in events. This should be `isPhiData`.

packages/aio-commerce-lib-app/source/management/installation/events/external.ts
jcuerdo: Noted — skipping for now to keep this PR focused. Will track as a future optimization.

packages/aio-commerce-lib-app/source/management/installation/events/utils.ts
jcuerdo: Done — providers without an explicit `key` are skipped during storage (guarded with `if (provider.key)`). The label fallback is kept only for `instance_id` generation since that doesn't affect runtime lookup.

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -287,7 +288,6 @@ eventing: {
   - **operator**: The comparison operator. Valid values: `"greaterThan"`, `"lessThan"`, `"equal"`, `"regex"`, `"in"`, `"onChange"`
   - **value**: The value to compare against
 - **destination**: Optional destination for the event. Must be a valid destination name.
-- **hipaaAuditRequired**: Optional boolean value to indicate if the event requires HIPAA audit.
```
iivvaannxx, line 290: Accidental delete?

packages/aio-commerce-lib-app/source/management/installation/events/utils.ts
```diff
@@ -41,12 +41,15 @@ const PROVIDER_TYPE_TO_LABEL = {
...
 /**
  * Generates a unique instance ID for I/O Events for this app deployment.
- * Uses `{metadata.id (first 100 chars)}-{providerKeyOrSlug}-{workspaceId}` (lowercased).
+ * Uses `{metadata.id (first 100 chars)}-{providerKey}-{workspaceId}` (lowercased).
  *
  * @param metadata - The metadata of the application
- * @param provider - The event provider (optional `key`, else label is slugified)
+ * @param provider - The event provider; must have an explicit `key`
  * @param workspaceId - Adobe I/O Developer Console workspace ID for this deployment
```
iivvaannxx, line 53: This doesn't seem in sync. The method allows optional keys (which is intended). Unless this is a formatter issue, this function should not change in this PR, as we're not doing anything related to instance ids at all.

plugins/commerce/app-migration/skills/commerce-app-migrate/agents/admin-ui-sdk.md
iivvaannxx, line 1: Accidental change

packages/aio-commerce-lib-app/source/config/schema/eventing.ts
```diff
@@ -245,6 +245,7 @@ const CommerceEventSchema = v.object({
 /** Schema for external event configuration */
 const ExternalEventSchema = v.object({
   ...BaseEventSchema.entries,
+  hipaa_audit_required: v.optional(booleanValueSchema("hipaa_audit_required")),
```
iivvaannxx, line 248: I would check this with @danipt. Might make sense for publishing with the `isPhiData`, but not sure if we want to do this for external events.

iivvaannxx (approved):

iivvaannxx:
@jcuerdo There are still some unneeded changes in some libraries that you can remove. The PR labels should only report `lib-events`, `lib-app`, `lib-api` and `spec`

packages/aio-commerce-lib-webhooks/docs/usage.md
```diff
@@ -40,9 +40,7 @@ const client = createCommerceWebhooksApiClient({
     baseUrl: "https://my-commerce-instance.com",
     flavor: "paas", // or "saas"
   },
-  auth: {
```
jcuerdo, line 43: Again 🤦‍♂️

packages/aio-commerce-lib-app/source/config/schema/eventing.ts
obarcelonap: From my perspective I'd not add anything related to hipaa to lib-app since it overcomplicates the implementation.
I'd just keep the lib-events flag.

packages/aio-commerce-lib-app/docs/usage.md
```diff
@@ -1047,6 +1047,87 @@ A plain `aio app deploy` on its own does not add the action: the `pre-app-build`
...
+Runtime actions can publish a custom I/O Event by referencing the provider and event exactly as declared in the `eventing` section of `app.commerce.config.ts`. At installation time, the SDK writes each configured provider's I/O Events ID and event codes to system storage. `publishEvent` resolves those automatically by the given key and publishes the event.
+
+`publishEvent(params)` takes:
+
+- `client` — an [`AdobeIoEventsApiClient`](../../aio-commerce-lib-events/docs/usage.md) created with the IMS auth to use for the ingress call.
+- `provider` — the `key` of an event provider declared in `app.commerce.config.ts`.
+- `event` — the `name` of an event declared under that provider.
+- `payload` — the event payload; any JSON object. The SDK wraps it in a CloudEvents 1.0 envelope before sending.
+
+Given this configuration:
```
obarcelonap, line 1038: I don't think this is a good example since you're suggesting to use a commerce provider to publish events while at the same time we receive them through the same provider. This is techincally feasible but from my perspective it shouldn't be the main use case.
The main goal is to be able to create your own custom provider and publish events so an external system or another runtime action can process those.
Just creating a regular provider (no commerce, without runtimeActions in events) and a runtime action that publishes with this provider should be enough.

jcuerdo: Done [01083ba](https://github.com/adobe/aio-commerce-sdk/pull/561/commits/01083ba6e840a869d748da00977270618fed0d08)

obarcelonap (approved):

PR #563, opened 2026-07-09
CEXT-6421: add isWebhookSuccessful helper to lib-webhooks

obarcelonap:
## Description

Add `isWebhookSuccessful()` to `@adobe/aio-commerce-lib-webhooks/responses` to determine whether a webhook action's result is successful.

## Related Issue

CEXT-6421

## Motivation and Context

Adobe Commerce webhooks always respond with HTTP 200, even when the handler wants to signal an error — the actual outcome is only visible in the response body's `op` field (`op: "exception"` = failure). Tooling that instruments webhook actions generically, such as tracing libraries (e.g. `aio-lib-telemetry`), can't rely on the status code alone to classify the result. This helper closes that gap.

## How Has This Been Tested?

Added unit tests covering: non-200 status codes, missing/no-op bodies, exception operations, success/add/replace/remove operations, arrays of operations, and non-object inputs. Ran `pnpm --filter @adobe/aio-commerce-lib-webhooks test` (100% coverage), `typecheck`, and `lint`.

## Types of changes

## Checklist:

- [x] I have signed the Adobe Open Source CLA.


.changeset/eighty-tires-read.md
iivvaannxx, line 1: The use case described is way too specific

packages/aio-commerce-lib-webhooks/docs/usage.md
```diff
@@ -467,6 +467,19 @@ export async function customizeCheckout(params) {
 }
 ```
 
+### Checking Action Results
+
+Because webhook actions always respond with HTTP 200, tooling that instruments actions generically (for example, tracing libraries like `aio-lib-telemetry`) can't rely on the status code to know whether a webhook handler actually succeeded. Use `isWebhookSuccessful()` to inspect the returned response body instead:
```
iivvaannxx: Same here

packages/aio-commerce-lib-webhooks/source/responses/helpers.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @example
+ * ```typescript
+ * import { isWebhookSuccessful } from "@adobe/aio-commerce-lib-webhooks/responses";
+ *
+ * const result = await runWebhookAction(params);
+ * span.setStatus(isWebhookSuccessful(result) ? { code: SpanStatusCode.OK } : { code: SpanStatusCode.ERROR });
+ * ```
+ */
+export function isWebhookSuccessful(result: unknown): boolean {
```
iivvaannxx, line 32: This can be better typed as a type predicate.

```suggestion
export function isWebhookSuccessful(result: unknown): result is SuccessOperation | AddOperation<TValue> | ReplaceOperation<TValue> | RemoveOperation {
```

Furthermore, I can understand the `as` casting logic for such a helper, but if you split this in smaller helpers `isSuccessResponse`, `isReplaceResponse` (or `isSuccessOperation`, `isReplaceOperation` maybe), then you can compose the logic and reach the same result in a more type-clean way.

obarcelonap: Your fault 06da2d2a

packages/aio-commerce-lib-core/source/responses/helpers.ts
```diff
@@ -66,6 +66,97 @@ export type ActionResponse<
   | SuccessResponse<TSuccessBody, THeaders>
   | ErrorResponse<TErrorBody, THeaders>;
 
+function isObjectRecord(value: unknown): value is Record<string, unknown> {
+  return Boolean(value) && typeof value === "object" && !Array.isArray(value);
```
iivvaannxx:
```suggestion
  return typeof value === "object" && object !== null && !Array.isArray(value);
```

packages/aio-commerce-lib-webhooks/source/responses/operations/types.ts
```diff
@@ -81,3 +81,45 @@ export type WebhookOperationResponse<TValue = unknown> =
   | AddOperation<TValue>
   | ReplaceOperation<TValue>
   | RemoveOperation;
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return Boolean(value) && typeof value === "object";
+}
```
iivvaannxx, line 87: LLM's love this kind of function lol. In any case, this should be like the one in `lib-core` and also check for not an array, since a `typeof` array also returns `object`.
```suggestion
function isRecord(value: unknown): value is Record<string, unknown> {
  return typeof value === "object" && typeof value !== "null" && !Array.isArray(value);
}
```

Duplication is fine for now, but if there were to exist any other version of this maybe we should move it to `common-utils`

packages/aio-commerce-lib-webhooks/source/responses/helpers.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (!isWebhookSuccessResponse(result)) {
+    return false;
+  }
+
+  if (!result.body) {
+    return true;
+  }
```
iivvaannxx, line 39: Is this correct to assume?

packages/aio-commerce-lib-webhooks/source/responses/types.ts
```diff
@@ -0,0 +1,55 @@
...
+  SuccessResponse,
+  "body" | "statusCode"
+> & {
+  body?: WebhookOperationResponse | WebhookOperationResponse[];
+  statusCode: typeof HTTP_OK;
+};
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return Boolean(value) && typeof value === "object";
+}
```
iivvaannxx, line 36: And here we have it again lol 🤣

iivvaannxx (approved):

packages/aio-commerce-lib-webhooks/source/responses/helpers.ts
obarcelonap: Core will do nothing if there's no operation but won crash the commerce instance so I'd say  yes, could be just a synchronous check relying on the statusCode of the response.

packages/aio-commerce-lib-webhooks/source/responses/types.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+export type WebhookSuccessResponse = Omit<
+  SuccessResponse,
+  "body" | "statusCode"
+> & {
+  body?: WebhookOperationResponse | WebhookOperationResponse[];
+  statusCode: typeof HTTP_OK;
+};
+
+function isRecord(value: unknown): value is Record<string, unknown> {
```
jcuerdo, line 34: Why not `!Array.isArray(value)` check here? In both cases, we are using a WebhookSuccessResponse.

In that case, could we reuse this isRecord() function instead of having 2 implementations?

jcuerdo (approved):
LGTM, just one comment

packages/aio-commerce-lib-webhooks/source/responses/types.ts
obarcelonap: We could but these are in different packages and I don't want to export this util function. I'll keep it as is.

obarcelonap: @jcuerdo sorry I missed the first sentence. Created a follow up pr to fix the check
https://github.com/adobe/aio-commerce-sdk/pull/565

PR #566, opened 2026-07-10
CEXT-6421: Publish previews in npm

obarcelonap:
## Description

Preview (alpha, per-PR `/snapshot`) and internal (beta) package snapshots now publish to the public npm registry instead of Artifactory, reusing the same npm token and `publishConfig` that stable releases already use. Also drops the now-unused release-channel branching in the Slack announcement, defaults the npm registry URL in the shared setup action, and fixes a leftover pre-release notice that incorrectly said beta snapshots weren't publicly available.

## Related Issue

N/A

## Motivation and Context

Artifactory-hosted snapshots caused two recurring problems:

- Consumers testing a preview from another public GitHub repo couldn't install alpha/beta packages at all, since Artifactory isn't reachable without internal credentials.
- Anyone who did have access needed to keep tweaking their local `.npmrc` to point at Artifactory, which was a poor developer experience for something as routine as trying out a preview build.

Publishing snapshots to the public npm registry removes both frictions, at the cost of alpha/beta versions being publicly visible (mitigated via the existing deprecate workflow if one needs to be pulled back).

## How Has This Been Tested?

- `pnpm test` for the affected `scripts` package (40/40 passing).
- `pnpm typecheck` across the whole monorepo.
- `pnpm lint` (biome) on all changed files.
- Manual review of the updated workflow YAML for the preview, internal, and public publish flows.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/.github/actions/setup-release/action.yml
+++ b/.github/actions/setup-release/action.yml
@@ -4,7 +4,8 @@ description: Setup pnpm and Node.js with registry auth, and install dependencies
 inputs:
   registry-url:
     description: npm registry URL to publish @adobe scope packages to
-    required: true
+    required: false
+    default: "https://registry.npmjs.org"
   token:
     description: Auth token for the npm registry
     required: true
@@ -16,7 +17,6 @@ runs:
     - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
 
     # No registry-url/scope/token — install always resolves packages from public npm.
-    # Avoids 404 on @adobe/aio-lib-* packages not present in Artifactory.
     - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
       with:
         node-version: lts/*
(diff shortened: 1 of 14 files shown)


.github/actions/setup-release/action.yml
```diff
@@ -4,7 +4,8 @@ description: Setup pnpm and Node.js with registry auth, and install dependencies
 inputs:
   registry-url:
     description: npm registry URL to publish @adobe scope packages to
-    required: true
+    required: false
+    default: "https://registry.npmjs.org"
```
iivvaannxx: Does this make sense to keep?

iivvaannxx (approved):

.github/actions/setup-release/action.yml
obarcelonap: will keep for the moment so it is clear where we publish

iivvaannxx: In other situations, I would agree, not on this one. If you keep this, you're allowing someone to input it and fail the workflow (unlikely, but unnecessary too). I would put a comment or something if that's your intention.

obarcelonap: fixed 4014da5

PR #571, opened 2026-07-13
CEXT-6454: Admin UI Entrypoints Specification

iivvaannxx:
## Description

Adds the design spec for CEXT-6454. This is a **spec-only PR** — per `specs/conventions.md`, the spec PR contains only the spec file, and implementation follows in a separate PR once the design is agreed. **Not for merge as-is**; it exists to align the team at the design sync.

The spec reconciles the pre-meeting ticket with the design discussion and frames the problem plus two candidate solutions:

- **Problem:** the Admin UI menu carries no path of its own, so it is implicitly bound to the app's index route — the same page Experience Shell renders by default. A page forced to serve both contexts breaks when it uses a Commerce-only API (e.g. `useCommerce`) outside Commerce.
- **Approach A — config `menu.path`:** the menu opens an explicit path instead of the app root, so the menu page and the Shell-default page can differ. Requires SDK **and** Commerce module (`adobe-commerce-backend-uix`) changes.
- **Approach B — a single declared entry point:** SDK-only; introduces `entry` on `createExtensionApp` and moves the throwing host hooks to a `data | error` result pattern so a shared entry can degrade gracefully.

It also documents the shared cleanup (remove the `index: true` route variant in favor of a uniform `{ path, element }` shape), how the knobs (`menu.path`, `entry`, result-pattern hooks) largely compose rather than being mutually exclusive, and the open decisions for the sync.

## Related Issue

[CEXT-6454](https://jira.corp.adobe.com/browse/CEXT-6454)

## Motivation and Context

The ticket was created before the design discussion and described a single settled approach. This spec captures what the discussion actually surfaced — a second candidate approach, the error-handling tradeoff, and the alternatives considered — so the team can align on which combination to build before implementation begins.

## How Has This Been Tested?

N/A — documentation only. No code changes.

## Types of changes

## Checklist:

- [x] This is a spec document; the API reference and package docs are unaffected.


specs/features/CEXT-6454-decouple-admin-ui-menu-route.md
```diff
@@ -0,0 +1,351 @@
...
+
+- Changing how Experience Shell selects its default page beyond it continuing to
+  render the `"/"` route.
+
+- Adding broad context-detection helpers for apps beyond the frame detection the
+  SDK already exposes.
+
+## The two approaches
+
+### Approach A — explicit `menu.path` (routing separation)
```
obarcelonap, line 83: Can't we use this approach while maintaining backward compatibility with the extension?
Something like:

* If menu is defined but root is not, set in the SDK menu as the root (backwards compatibility, no need to change extension).
* If root is defined but menu is not, leave it as is (not using menu extension point).
* If both menu and root are defined, keep both (user wants to have differnt entrypoints).
* If neither is defined, create a default root (using another extension point, we provide a default to the PHP module).

This way, the PHP module doesn't need to be changed and we preserve the current behavior. At the same time, we support different entry points in the future if needed.

obarcelonap (approved):
I prefer option A and I think it can be rethought a bit to be backwards compatible (see comment).
In any case I can live with option B too but from my perspective the intent of the entrypoint concept might not be clear to the user (when to use it, what's the difference with menu, etc)

specs/features/CEXT-6454-decouple-admin-ui-menu-route.md
```diff
@@ -0,0 +1,351 @@
...
+  a touch-every-app migration, and the surface grows the longer it waits.
+
+- **Approach A:** a required `menu.path` means existing `adminUi.menu` configs
+  must add one; two-repo coordination risks a mis-sequenced rollout opening menus
+  at an unregistered path; and it does not prevent a throwing hook on the `"/"`
+  route.
+
+- **Approach B:** moving host hooks from throwing to a result pattern touches
+  every hook consumer and is the larger API rework; it also keeps the menu and
+  Shell default as one page, so context separation is only in-page.
```
asalloum5, line 296: `it also keeps the menu and Shell default as one page, so context separation is only in-page`
Why is this considered as a drawback?

iivvaannxx: I guess that although, from a technical point of view, your assertion about a single entry point is correct, it's still "more work" to do that kind of context separation manually rather than with a built-in OOTB separation (not what we're targeting anyways). 

I say having that as a drawback is fair, but I don't feel strongly about it, so if it bothers you, I can remove it.

specs/features/CEXT-6454-decouple-admin-ui-menu-route.md
iivvaannxx: I am having a bit of trouble understanding what you mean. 

> If menu is defined but root is not, set in the SDK menu as the root (backwards compatibility, no need to change extension).

From what I understand here is if I have `menu` in `app.commerce.config` and someone does not declare a `/` path or `entry` I should set the menu as the root. But how am I supposed to know which element to render in the menu if the user doesn't give it to me explicitly? The root/entry needs to be manually specified always, doesn't it?

iivvaannxx: Discussed and clarified offline. Having `menu.path` requires PHP changes, and there's no workaround on it.

iivvaannxx:
Final option is B, will not edit the spec to remove A for history purposes. Instead will add a header saying the chosen option.

PR #574, opened 2026-07-15
Fix Commerce integration (OAuth 1.0a) request signing for URLs with array-style query parameters

aligent-lturner:
## Description

`getIntegrationAuthProvider` handed the fully-encoded request URL to  `oauth-1.0a`'s `authorize({ url })`. Internally that library's `deParam`  decodes parameter **values** but not **keys**, so a percent-encoded key such as  `searchCriteria%5BpageSize%5D` is re-encoded a second time (`%255B…`) when the  signature base string is built. Adobe Commerce reconstructs the base string  with  single-encoded keys, the two disagree, and the request is rejected with  `401 "The signature is invalid."`. This affects every OAuth 1.0a (PaaS) call  that  carries array-style query parameters — i.e. anything using `searchCriteria`  (`products`, `categories`, `products/attributes`, …).

 This change stops delegating URL parsing to `oauth-1.0a`. In `getHeaders` we  now:

- parse the URL and read its query via `URLSearchParams`, which decodes  **both**   keys and values;
- pass those decoded pairs as `request.data`, which `oauth-1.0a` merges into  the  signature parameter string (RFC 5849 §3.4.1.3 / §3.4.1.1); and
- sign against the **query-less** base URI (§3.4.1.2).

The parameters therefore remain part of the signature — they are simply sourced  correctly (decoded once, then single-encoded), so the base string matches  Commerce's. Repeated keys collapse to an array, matching `oauth-1.0a`'s  existing
multi-value handling. Public API and behaviour for query-less requests are  unchanged.
  
## Related Issue

Root cause is the upstream, unmaintained `oauth-1.0a` (last release 2019):
- ddo/oauth-1.0a#49 ("Base string incorrect for array type GET query string") and
- ddo/oauth-1.0a#24. Fixed here in the provider rather than waiting on the dependency.

## Motivation and Context

Any consumer using the Commerce integration (OAuth 1.0a) auth on a PaaS instance cannot call the standard search endpoints — the most common REST calls in an  App Builder app — because the signature never validates. There is no config or credential workaround; it fails purely on the request shape. This restores `searchCriteria`-based reads/writes for OAuth1 integrations.

## How Has This Been Tested?
- **New regression test** in `test/integration-auth.test.ts`: with a frozen  nonce and timestamp, the same logical request expressed with percent-encoded  (`%5B…%5D`) vs. literal (`[…]`) bracket keys must produce an identical `Authorization` header. Verified it **fails on the pre-fix code** (identical nonce/timestamp but different `oauth_signature`) and **passes after** the fix.
- Full package suite: `vitest run` — **35/35 passing**.
- `biome check` (formatter + linter + assist): clean, no fixes.
- `tsc --noEmit`: clean.
- **End-to-end** against a live Adobe Commerce PaaS instance: reproduced  `401 "The signature is invalid."` on `GET
  /V1/products/attributes?searchCriteria[...]` with the current release, and confirmed the corrected base string (decoded keys) is accepted (`200`).

Environment: Node.js 22, `vitest` 4, `@biomejs/biome` 2.

N/A

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality  to change)

## Checklist:

### DIFF
--- a/.changeset/fix-integration-oauth-query-signature.md
+++ b/.changeset/fix-integration-oauth-query-signature.md
@@ -2,4 +2,4 @@
 "@adobe/aio-commerce-lib-auth": patch
 ---
 
-Fix Commerce integration (OAuth 1.0a) request signing for URLs with array-style query parameters (e.g. `searchCriteria[...]`). Query parameters are now decoded via `URLSearchParams` and signed as normalized parameters against the query-less base URI (RFC 5849 §3.4.1.2), instead of delegating URL parsing to `oauth-1.0a` — whose `deParam` does not decode parameter keys, leaving percent-encoded keys double-encoded in the signature base string. This caused Adobe Commerce to reject affected requests with "The signature is invalid."
+Fix Commerce integration (OAuth 1.0a) request signing for URLs with array-style query parameters (e.g. `searchCriteria[...]`). This caused Adobe Commerce to reject affected requests with "The signature is invalid."
--- a/.changeset/fix-integration-oauth-query-signature.md
+++ b/.changeset/fix-integration-oauth-query-signature.md
@@ -1,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-auth": patch
+---
+
+Fix Commerce integration (OAuth 1.0a) request signing for URLs with array-style query parameters (e.g. `searchCriteria[...]`). Query parameters are now decoded via `URLSearchParams` and signed as normalized parameters against the query-less base URI (RFC 5849 §3.4.1.2), instead of delegating URL parsing to `oauth-1.0a` — whose `deParam` does not decode parameter keys, leaving percent-encoded keys double-encoded in the signature base string. This caused Adobe Commerce to reject affected requests with "The signature is invalid."
--- a/packages/aio-commerce-lib-auth/source/lib/integration-auth/provider.ts
+++ b/packages/aio-commerce-lib-auth/source/lib/integration-auth/provider.ts
@@ -110,9 +110,35 @@ export function getIntegrationAuthProvider(
 
   return {
     getHeaders: (method: HttpMethodInput, url: AdobeCommerceUrl) => {
-      const urlString = parse(UrlSchema, url);
+      const parsed = new URL(parse(UrlSchema, url));
+
+      // OAuth 1.0a signs the request parameters (RFC 5849 §3.4.1.3) merged into
+      // the base string (§3.4.1.1); the base-string URI itself excludes the
+      // query (§3.4.1.2). oauth-1.0a normally derives those parameters from the
+      // URL via `deParam`, which decodes parameter values but NOT keys — so
+      // percent-encoded keys (e.g. `searchCriteria%5B...%5D`) get double-encoded
+      // in the base string and Adobe Commerce rejects the request with "The
+      // signature is invalid.". Instead we decode the query with URLSearchParams
+      // (keys and values) and pass it as `request.data`, which oauth-1.0a folds
+      // into the very same parameter string (getParameterString merges `data`) —
+      // so the params stay signed, just sourced correctly — and we sign against
+      // the query-less base URI. Repeated keys collapse to an array, matching
+      // oauth-1.0a's own multi-value handling.
+      const data: Record<string, string | string[]> = {};
+      for (const [key, value] of parsed.searchParams) {
+        const existing = data[key];
+        if (existing === undefined) {
+          data[key] = value;
+        } else {
+          data[key] = Array.isArray(existing)
+            ? [...existing, value]
+            : [existing, value];
+        }
+      }
+
+      const baseUri = `${parsed.origin}${parsed.pathname}`;
       return oauth.toHeader(
-        oauth.authorize({ method, url: urlString }, oauthToken),
+        oauth.authorize({ data, method, url: baseUri }, oauthToken),
       );
     },
   };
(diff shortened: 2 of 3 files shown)


packages/aio-commerce-lib-auth/source/lib/integration-auth/provider.ts
```diff
@@ -110,9 +110,35 @@ export function getIntegrationAuthProvider(
...
+      // query (§3.4.1.2). oauth-1.0a normally derives those parameters from the
+      // URL via `deParam`, which decodes parameter values but NOT keys — so
+      // percent-encoded keys (e.g. `searchCriteria%5B...%5D`) get double-encoded
+      // in the base string and Adobe Commerce rejects the request with "The
+      // signature is invalid.". Instead we decode the query with URLSearchParams
+      // (keys and values) and pass it as `request.data`, which oauth-1.0a folds
+      // into the very same parameter string (getParameterString merges `data`) —
+      // so the params stay signed, just sourced correctly — and we sign against
+      // the query-less base URI. Repeated keys collapse to an array, matching
+      // oauth-1.0a's own multi-value handling.
```
iivvaannxx: This seems to me to be one of those AI book-long comments 🙈. A lot of internal implementation details are in here that may not be needed. What about something like this?

```suggestion
      // oauth-1.0a's authorize method re-encodes keys even if already encoded
      // Split query parameters from base URL to avoid double encoding
```

packages/aio-commerce-lib-auth/source/lib/integration-auth/provider.ts
```diff
@@ -110,9 +110,35 @@ export function getIntegrationAuthProvider(
...
+      for (const [key, value] of parsed.searchParams) {
+        const existing = data[key];
+        if (existing === undefined) {
+          data[key] = value;
+        } else {
+          data[key] = Array.isArray(existing)
+            ? [...existing, value]
+            : [existing, value];
+        }
+      }
```
iivvaannxx, line 123:
```suggestion
      const data: Record<string, string | string[]> = {};
      const uniqueKeys = Array.from(new Set(parsed.searchParams.keys()))
      
      for (const key of uniqueKeys) {
        const values = parsed.searchParams.getAll(key);
        data[key] = values.length > 1 ? values : values[0];
      }
```

.changeset/fix-integration-oauth-query-signature.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-auth": patch
+---
+
+Fix Commerce integration (OAuth 1.0a) request signing for URLs with array-style query parameters (e.g. `searchCriteria[...]`). Query parameters are now decoded via `URLSearchParams` and signed as normalized parameters against the query-less base URI (RFC 5849 §3.4.1.2), instead of delegating URL parsing to `oauth-1.0a` — whose `deParam` does not decode parameter keys, leaving percent-encoded keys double-encoded in the signature base string. This caused Adobe Commerce to reject affected requests with "The signature is invalid."
```
iivvaannxx: We usually try to keep changesets without implementation details that do not help the description much.

```suggestion
Fix Commerce integration (OAuth 1.0a) request signing for URLs with array-style query parameters (e.g. `searchCriteria[...]`). This caused Adobe Commerce to reject affected requests with "The signature is invalid."
```

packages/aio-commerce-lib-auth/source/lib/integration-auth/provider.ts
aligent-lturner: Yep - I wanted to make sure it was clear what was going on, but it is a bit too wordy.

packages/aio-commerce-lib-auth/source/lib/integration-auth/provider.ts
aligent-lturner: I'm not sure these are actually equivalent? `keys()` gives the keys as strings, so destructuring would just give the first 2 characters?

Did you mean this?:
```
for (const key of parsed.searchParams.keys()) {
  const values = parsed.searchParams.getAll(key);
  data[key] = values.length > 1 ? values : values[0];
}
```

It's terser like that, but it would also potentially doing some redundant work if there were multiple occurrences of a key (e.g. `?a=1&a=2&a=3` would call `getAll(a)` three times. The output is the same, and it's probably not going to be a big impact at all, so happy to go with that if you greatly prefer the less-code version.

.changeset/fix-integration-oauth-query-signature.md
aligent-lturner: No worries - happy to leave those out. 👍

packages/aio-commerce-lib-auth/source/lib/integration-auth/provider.ts
iivvaannxx: Yes, sorry, I edited it to get only the `key`. I edited again to avoid processing the same key multiple times. Let me know what you think; for me, it reads a bit more easily than the original version.

iivvaannxx:
@aligent-lturner The formatter is complaining about an indented white line. Can you run `pnpm run format` inside the `lib-auth` package and commit the fix?

iivvaannxx (approved):

aligent-lturner:
> @aligent-lturner The formatter is complaining about an indented white line. Can you run `pnpm run format` inside the `lib-auth` package and commit the fix?

This is done now

obarcelonap (approved):

PR #580, opened 2026-07-20
CEXT-6454: decouple Admin UI menu from routes

iivvaannxx:
## Description

Implements option B from the CEXT-6454 Admin UI entry-point specification, using `menu` as the Adobe Commerce Admin menu page declaration.

- Adds the optional `menu` option to `createExtensionApp` and decouples it from additional path-based routes.
- Removes index-route declarations and uses a uniform `{ path, element }` route shape. The root path is reserved only when `menu` is declared.
- Changes the host-dependent browser hooks to return complementary `{ data, error }` results so pages shared with Experience Cloud Shell can handle unavailable Commerce context without crashing.
- Makes the retryable promise cache understand fulfilled error results.
- Updates the Admin UI generator, public usage documentation, tests, and the specification's implementation status.

## Related Issue

[CEXT-6454](https://jira.corp.adobe.com/browse/CEXT-6454)

## Motivation and Context

The Admin menu page was previously represented by the app's index route, which also made it the default Experience Cloud Shell page. Commerce-only hooks throw outside Commerce, so a shared page could crash when loaded by another host.

Declaring the menu page independently makes that intent explicit, while result-returning host hooks let shared pages degrade gracefully or forward the error to an error boundary when desired.

## How Has This Been Tested?

- Admin UI unit and browser tests: 35 files, 218 tests passed.
- Lib App unit and integration tests: 58 files, 949 tests passed.
- Workspace TypeScript typecheck passed.
- Biome checks passed for the affected packages.
- Admin UI ESM and CJS builds completed successfully.

N/A

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/packages/aio-commerce-lib-admin-ui/docs/usage.md
+++ b/packages/aio-commerce-lib-admin-ui/docs/usage.md
@@ -580,7 +580,7 @@ function Actions() {
 
 #### Reading the mass-action selection
 
-`useMassActionContext` returns the row IDs the mass action was triggered with. It reads from the host-provided Commerce context and returns an error when that context doesn't include a mass-action selection:
+`useMassActionContext` returns the row IDs the mass action was triggered with. It reads from the host-provided Commerce context and returns an error when the selection is missing, empty, or contains a non-string row ID:
 
 ```jsx
 import { useMassActionContext } from "@adobe/aio-commerce-lib-admin-ui/web";
@@ -589,7 +589,7 @@ function MassActionPage() {
   const { data, error } = useMassActionContext();
   if (error) return null;
 
-  // data.selectedIds is string[] (an empty selection is valid).
+  // data.selectedIds is a non-empty string[].
 }
 ```
 
--- a/packages/aio-commerce-lib-admin-ui/docs/usage.md
+++ b/packages/aio-commerce-lib-admin-ui/docs/usage.md
@@ -490,7 +490,7 @@ The `./web` entrypoint provides the browser side of a `commerce/backend-ui/2` ap
 
 #### Mounting an app
 
-`createExtensionApp` mounts the app into the element with id `root` (or into the `root` element you pass). `routes` must start with the index route; additional routes are path-based:
+`createExtensionApp` mounts the app into the element with id `root` (or into the `root` element you pass). When your app has a Commerce Admin menu, pass its page through the optional `menu` property. Experience Cloud Shell also renders this page by default. Use `routes` for additional pages:
 
 ```jsx
 // src/app.jsx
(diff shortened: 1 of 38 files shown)


packages/aio-commerce-lib-admin-ui/docs/usage.md
```diff
@@ -502,38 +502,63 @@ import { SettingsPage } from "./pages/settings-page.jsx";
 
 createExtensionApp({
   metadata: { extensionId: "my-extension-id" },
-  routes: [
-    { index: true, element: <MainPage /> },
-    { path: "/settings", element: <SettingsPage /> },
-  ],
+  menu: <MainPage />,
+  routes: [{ path: "/settings", element: <SettingsPage /> }],
```
asalloum5, line 506:
```suggestion
  mainPageElement: <MainPage />,
  routes: [{ path: "/settings", element: <SettingsPage /> }],
```
Just a suggestion so it's clear what to refer.

packages/aio-commerce-lib-admin-ui/docs/usage.md
```diff
@@ -545,34 +570,41 @@ function CommerceInfo() {
...
 ```jsx
 import { useMassActionContext } from "@adobe/aio-commerce-lib-admin-ui/web";
 
 function MassActionPage() {
-  const { selectedIds } = useMassActionContext();
-  // selectedIds is string[] (an empty selection is valid).
+  const { data, error } = useMassActionContext();
+  if (error) return null;
+
+  // data.selectedIds is string[] (an empty selection is valid).
```
asalloum5: This is weird to have empty selection as valid and we have an error, not sure I fully understand this

packages/aio-commerce-lib-admin-ui/source/web/react/commerce/hooks/use-host-connection.ts
```diff
@@ -27,37 +28,41 @@ type HostFrameField = {
...
-      close: () => requireField().close(),
-      closeWithError: () => requireField().onError(),
+      actions: {
+        close: () => field.close(),
+        closeWithError: () => field.onError(),
+      },
+      error: null,
     };
-  }, [host]);
+  }, [data, error]);
```
oshmyheliuk: Now, useMemo depends on (data, error) from useSharedContext(), but it returns a fresh {data, error} object on every render, so it looks like useMemo won't work now, or you need to add useMemo in useSharedContext as well.
The same in `use-extension-context`

packages/aio-commerce-lib-admin-ui/source/web/react/commerce/hooks/use-host-connection.ts
```diff
@@ -27,37 +28,41 @@ type HostFrameField = {
...
+    }
+
+    const { field } = data.host as { field?: HostFrameField };
+    if (!field) {
+      return {
+        actions: null,
+        error: new Error(
+          "Host frame actions are unavailable. They require an established guest connection with a host that exposes frame actions.",
+        ),
+      };
```
oshmyheliuk: Not an issue, but returning error and ok results can be wrapped in some helper function, as the same pattern is already used in multiple places.

packages/aio-commerce-lib-admin-ui/docs/usage.md
```diff
@@ -502,38 +502,63 @@ import { SettingsPage } from "./pages/settings-page.jsx";
 
 createExtensionApp({
   metadata: { extensionId: "my-extension-id" },
-  routes: [
-    { index: true, element: <MainPage /> },
```
jcuerdo, line 506: Now we have to review  https://github.com/OneAdobe/commerce-app-purchase-approval and https://github.com/adobe/adobe-commerce 
  -samples/blob/feature/admin-ui-sdk-v2-samples/admin-ui-sdk/v2/ examples that uses **index:true**

packages/aio-commerce-lib-admin-ui/docs/usage.md
```diff
@@ -502,38 +502,63 @@ import { SettingsPage } from "./pages/settings-page.jsx";
...
+
 The whole app is wrapped in React's [`<StrictMode>`](https://react.dev/reference/react/StrictMode). When running locally with `aio app dev` or `aio app run` (which serve React's development build), StrictMode runs its extra development-only checks: components render twice, and effects run an extra setup + cleanup cycle on mount. Duplicated renders, effect runs, or requests fired from effects during development are therefore expected — not a bug — and surface unsafe side effects early. These checks do not run in production builds, where StrictMode has no effect.
 
 #### Reading IMS credentials
 
-`useIms` returns the IMS credentials provided by the host (`{ imsToken, imsOrgId }`). It works inside both the Commerce Admin and the Experience Cloud shell, and throws when the app runs standalone (no host provides credentials):
+The browser hooks return a `Result` with complementary `data` and `error` fields. When the requested host data is available, `data` contains it and `error` is `null`. Otherwise, `data` is `null` and `error` explains why the `data` can't be provided.
```
obarcelonap: `Result` type is not exported but it is mentioned in the docs
https://github.com/adobe/aio-commerce-sdk/blob/a5054c08fc2c23e65666ea9238ef3d43b6b53192/packages/aio-commerce-lib-admin-ui/source/web/index.ts#L38
I'd say either we export or we generalize this wording: return an object with `data` and `error` props...

obarcelonap (approved):

packages/aio-commerce-lib-admin-ui/docs/usage.md
jcuerdo: Also, migration skill should be taken into account.

jcuerdo (commented):
Missing changeset

iivvaannxx:
> Missing changeset

@jcuerdo Changeset is not needed since this is unreleased work. Nothing currently publicly released is affected that deserves a changelog entry.

packages/aio-commerce-lib-admin-ui/docs/usage.md
iivvaannxx: > Now we have to review samples

We do, yes. This is expected after this change, since the confusion that drove it started there.


> Also, migration skill should be taken into account.

Will do 👍🏻

packages/aio-commerce-lib-admin-ui/docs/usage.md
iivvaannxx: Initially, it was exported. I prefer not to, since it's an internal type used to model our results not something developers should import. I'll rephrase

packages/aio-commerce-lib-admin-ui/docs/usage.md
iivvaannxx: Is valid in the sense that we don't throw an error, but the type is assignable so I didn't make it fail. I can do it if you want, don't have a strong opinion. Should never happen anyways since Commerce doesn't support executing a mass action without a selection.

packages/aio-commerce-lib-admin-ui/docs/usage.md
iivvaannxx: I understand the suggestion here, but there's a tradeoff (one we contemplated in the spec). Ideally the property name should indicate which extension point/config it maps to, since a main driver of this change was that `index: true` was confusing and not easily relatable to the `menu`.

The reasoning below is agnostic to the `Element` suffix; adding or omitting it doesn't change anything.

- `mainPage` (or `entry`, as the spec suggested) reads well as an explicit "your app starts here", and since it isn't domain-specific, it fits both Experience and Commerce. The "downside" is that it doesn't hint that it gets rendered with the `menu`.

- `menu` maps directly onto the `menu` config in `app.commerce.config`, so the connection is obvious. The flip side is that it's domain-specific and reads as Commerce-only, even though Experience uses it too (it's just an alias for the `/` root path).

We landed on `menu` because Experience isn't our main target here (this library is aimed at Commerce, specifically the Admin UI), so the Commerce-side clarity wins. Let me know what you think.

cc @obarcelonap as he was the one suggesting this property to be explicitly correlatable to the `menu` in `app.commerce.config`.

obarcelonap: I like `menu`, sorry to say it again but our extension point is within Commerce so we should try to design keeping that in mind even if we support other extension mechanisms (like the famous experience entrypoint).

packages/aio-commerce-lib-admin-ui/docs/usage.md
iivvaannxx: Rephrased in [`ffcd598`](https://github.com/adobe/aio-commerce-sdk/pull/580/commits/ffcd5987639ccb57ecf3c1cd277553a59e36d6a5)

packages/aio-commerce-lib-admin-ui/source/web/react/commerce/hooks/use-host-connection.ts
iivvaannxx: Nice catch, fixed in [`eb83f41`](https://github.com/adobe/aio-commerce-sdk/pull/580/commits/eb83f41fb1f9de4d3b623966905b33a3b78ea8d4). Added regression tests to ensure values are memoized and functionaility doesn't break.

packages/aio-commerce-lib-admin-ui/source/web/react/commerce/hooks/use-host-connection.ts
iivvaannxx: Refactored in [`e418a79`](https://github.com/adobe/aio-commerce-sdk/pull/580/commits/e418a79fb54b2294318888baa083f7eece953f19)

packages/aio-commerce-lib-admin-ui/docs/usage.md
iivvaannxx: Updated AI Skills in [`39cbd07`](https://github.com/adobe/aio-commerce-sdk/pull/580/commits/39cbd07ea7addba4bf2ebbf048ea81dd2e9cd267)

packages/aio-commerce-lib-admin-ui/docs/usage.md
iivvaannxx: I ended up checking for emptiness and item types since we assert it's a `string[]`. Now, an error is returned if the payload doesn't match expectations. It's unlikely to happen but we're covered there. See [`e10d293`](https://github.com/adobe/aio-commerce-sdk/pull/580/commits/e10d29357df4c7f87950a47cce41cd6e81b5254a)

packages/aio-commerce-lib-admin-ui/docs/usage.md
asalloum5: I'm fine keeping it `menu`

jcuerdo (approved):

PR #592, opened 2026-07-23
CEXT-6498: publishEvent: uninstall does not clear stored event/provider data

jcuerdo:
## Description
publishEvent (added in aio-commerce-sdk PR #561, package @adobe/aio-commerce-lib-app) resolves an event's I/O Events provider ID and event code by reading data written to system config storage under the key EVENTS_STORAGE_KEY (="events") at install time. This storage is never cleared or updated during uninstall, so a stale, already-deleted providerId can remain in storage and cause publishEvent to fail after uninstall.

## Test

Tested with: https://development-271530-956ochrerattlesnake-cext6498.adobeioruntime.net/api/v1/web/my-package/emit-event

After install:
image
After uninstall:
image

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/source/management/installation/events/external.ts
```diff
@@ -162,5 +163,11 @@ async function removeExternalEvents(
     );
   }
 
+  await removeStoredEventProviders(
```
jcuerdo, line 166: We could simplify is just doing   await setSystemConfigByKey(EVENTS_STORAGE_KEY, {});

In external.ts‎ and commerce.ts, but we will duplicate logic if both external and commerce have providers.

By the other hand, I check if there was a common install/uninstall for "events" in general, but there is not AFAIK.

packages/aio-commerce-lib-app/source/management/installation/events/utils.ts
```diff
@@ -44,6 +48,42 @@ const METADATA_ID_MAX_LENGTH_FOR_INSTANCE_ID = 100;
...
+    return;
+  }
+
+  const providers = Object.fromEntries(
+    Object.entries(existing.providers).filter(
+      ([key]) => !providerKeysToRemove.has(key),
+    ),
+  );
+
+  await setSystemConfigByKey(EVENTS_STORAGE_KEY, { providers });
```
iivvaannxx, line 89: We have a `deleteSystemConfig` or something similar. Maybe it makes sense to check whether `providers` will become empty after this operation, and remove it directly instead of leaving a dangling artifact in `lib-files`.

iivvaannxx (commented):
This would work, but I wonder if it's just not easier to delete the whole events system config on uninstall instead of per-provider.

packages/aio-commerce-lib-app/source/management/installation/events/utils.ts
jcuerdo: We don't have it, but as you can see here https://github.com/adobe/aio-commerce-sdk/blob/eb1eae5afdd4e67e19c795ce1dc88eb22669cf3d/packages/aio-commerce-lib-config/source/modules/configuration/system-config.ts#L54

Doing  `await setSystemConfigByKey(EVENTS_STORAGE_KEY, null)` will remove the key. 

I'll work on it

jcuerdo: Done ¡¡

packages/aio-commerce-lib-app/source/management/installation/events/external.ts
obarcelonap: Sorry but I dont' get this comment. If I recall correctly, this happens on uninstall so we want to wipe all providers.
Why do we need to set the provider keys as arg if we need to cleanup everything?

obarcelonap: Clarified by chat, LGTM

obarcelonap (approved):

packages/aio-commerce-lib-app/source/management/installation/events/external.ts
```diff
@@ -162,5 +163,11 @@ async function removeExternalEvents(
...
+    config.eventing.external
+      .map(({ provider }) => provider.key)
+      .filter((key): key is string => Boolean(key)),
```
iivvaannxx: Nitpick: This asserts `key` is `string` but it uses `Boolean` for a truthyness conversion (I assume to check `!== undefined`). Wouldn't it be clearer to have `typeof key === string`?

packages/aio-commerce-lib-app/test/unit/management/installation/events/commerce.test.ts
```diff
@@ -402,3 +404,170 @@ describe("commerceEventsStep orchestration", () => {
...
+            provider: {
+              description: "Provides commerce events",
+              key,
+              label: "Order Events Provider",
+            },
+          },
+        ],
+      },
+    };
+  }
```
iivvaannxx: Can we not mix fixture logic within tests? We have `fixtures` for that. Probably a similar fixture already exists also. This also goes for `importCommerceStepWithMocks`.

packages/aio-commerce-lib-app/test/unit/management/installation/events/utils.test.ts
```diff
@@ -519,3 +519,133 @@ describe("existing data normalization", () => {
...
+        "order-events-provider": { events: {}, id: "stale-uuid" },
+      },
+    });
+
+    await removeStoredEventProviders(["order-events-provider"]);
+
+    expect(configMocks.setSystemConfigByKey).toHaveBeenCalledWith(
+      "events",
+      null,
+    );
```
iivvaannxx, line 649: For this kind of tests that interact with business config, I usually think it's best to mock the storage layer into a map or something (I think we have the test infra for that?) And check actual values instead of testing whether a method has been called with specific parameters (implementation details, flaky tests)

iivvaannxx (approved):
Approving with nitpicks. Would be great if fixed, but not blockers

PR #593, opened 2026-07-23
CEXT-6338: suggest Spectrum S2 migration in Admin UI skills

iivvaannxx:
## Summary
- `commerce-app-admin-ui` now detects when an existing Admin UI `web-src` still depends on classic React Spectrum (v3) instead of `@react-spectrum/s2`, and suggests upgrading — pointing to the `commerce-app-migrate` skill to actually run it, rather than performing the upgrade itself.
- `commerce-app-migrate` gains a `spectrum-s2-upgrade` reference the migration executor follows: detect classic Spectrum, explain S2 is compatible and the recommended path forward, and offer to help run the migration (Agent Skill install or codemod) — still gated on the user's go-ahead.
- Added one eval to each skill's `evals.json` covering the new detect-and-suggest behavior.
- Two changesets (minor) for `@adobe/aio-commerce-plugin-app-management` and `@adobe/aio-commerce-plugin-app-migration`.

## Test plan
- [x] `evals.json` files validated as well-formed JSON

### DIFF
--- a/.changeset/admin-ui-spectrum-s2-suggestion.md
+++ b/.changeset/admin-ui-spectrum-s2-suggestion.md
@@ -1,0 +1,5 @@
+---
+"@adobe/aio-commerce-plugin-app-management": minor
+---
+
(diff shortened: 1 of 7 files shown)


obarcelonap (approved):
very nice

jcuerdo (approved):

PR #621, opened 2026-08-11
CEXT-6507: implement webhook additions/removals for upgrade planning

obarcelonap:
## Description

Implements the webhooks domain's `plan`/`apply` resource capability, on top of the shared upgrade orchestration types added in CEXT-6527: `planWebhookSubscriptions` diffs the baseline snapshot against the target configuration into add/remove operations (no external reads or writes), and `applyWebhookSubscriptions` executes them idempotently against live Commerce state, aborting on the first failure so a partial apply never reports success.

Removals are always ordered before additions, so a webhook rename (same `webhook_method`/`webhook_type`, different `batch_name`/`hook_name`) never briefly double-registers a hook point. OAuth credentials (`developer_console_oauth`) are kept out of plans and persisted snapshots entirely, and are resolved fresh only at the moment a webhook is actually subscribed.

The existing `install`/`uninstall`/`validate` handlers and lifecycle are untouched.

## Related Issue

CEXT-6507

## Motivation and Context

App Management needs to compare the webhook definitions in an installed application against the target version during an upgrade, and automatically create newly added webhooks or remove obsolete ones, without unnecessarily touching webhooks that didn't change.

## How Has This Been Tested?

Added unit tests for `planWebhookSubscriptions` and `applyWebhookSubscriptions` covering: no baseline, no target, mixed add/remove, cleanup-vs-configuration removal categorization, rename ordering, environment filtering, idempotent re-application, abort-on-failure, and that credentials never appear in a plan or a persisted snapshot. Added integration tests exercising both functions against a mocked Commerce API. Full package test suite (1018 tests), typecheck, and lint all pass.

N/A

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/source/management/domains/webhooks/helpers.ts
```diff
@@ -381,25 +655,32 @@ export function resolveDeveloperConsoleOAuthCredentials(
...
+function webhookIdentitiesMatch(
+  a: WebhookIdentity,
+  b: WebhookIdentity,
+): boolean {
+  return (
+    normalizeWebhookMethod(a.webhook_method) ===
+      normalizeWebhookMethod(b.webhook_method) &&
+    a.webhook_type === b.webhook_type &&
+    a.batch_name === b.batch_name &&
+    a.hook_name === b.hook_name
```
jcuerdo: What if there is a change in other fields like runtimeAction or rules for example? Shouldn't we validate the complete webhook to update it if some other field is being modified?

obarcelonap: Webhook modifications are in the scope of the ticket https://jira.corp.adobe.com/browse/CEXT-6508
Current ticket only targets adding/removing webhooks

jcuerdo: Then is ok for me 👍

jcuerdo (approved):

packages/aio-commerce-lib-app/source/management/domains/webhooks/helpers.ts
```diff
@@ -63,6 +72,45 @@ export type WebhookUnsubscriptionResult = {
...
+
+/** Narrows any webhook-like value down to its identity fields. */
+function toIdentity(webhook: WebhookIdentity): WebhookIdentity {
+  return {
+    batch_name: webhook.batch_name,
+    hook_name: webhook.hook_name,
+    webhook_method: webhook.webhook_method,
+    webhook_type: webhook.webhook_type,
+  };
+}
```
iivvaannxx: What's the purpose of this method? 🤣  Maybe you mistyped something?

packages/aio-commerce-lib-app/source/management/domains/webhooks/helpers.ts
```diff
@@ -270,6 +308,242 @@ export async function deleteWebhookSubscriptions(
   return { unsubscribedWebhooks };
 }
 
+/** A fully-resolved webhook payload, minus credentials, plus whether they're required. */
+type ResolvedWebhookPayload = WebhookIdentity &
+  Omit<WebhookSubscribeParams, "developer_console_oauth"> & {
+    requiresAdobeAuth: boolean;
+  };
```
iivvaannxx: Isn't this type very similar to the one defined at the top?

packages/aio-commerce-lib-app/source/management/domains/webhooks/helpers.ts
```diff
@@ -270,6 +308,242 @@ export async function deleteWebhookSubscriptions(
...
+
+  return Promise.resolve({
+    kind: "planned",
+    plan: {
+      operations: [...removeOperations, ...addOperations],
+      path,
+      possibleCleanupResources,
+      retainedWebhooks,
+    },
+  });
```
iivvaannxx: Shouldn't removed webhooks also be added to unresolved cleanup resources? What happens if I remove a webhook, the upgrade fails after removing it, and then I decide I don't want to remove it?

packages/aio-commerce-lib-app/source/management/domains/webhooks/helpers.ts
```diff
@@ -270,6 +308,242 @@ export async function deleteWebhookSubscriptions(
...
+  const resolvedCleanupResources: CleanupResource<WebhookIdentity>[] = [];
+  const subscribedWebhooks: WebhookSubscribeParams[] = [
+    ...plan.retainedWebhooks,
+  ];
+
+  for (const operation of plan.operations) {
+    if (operation.kind === "add") {
+      // `after` is always fully resolved for `add` (see planWebhookSubscriptions).
+      const { requiresAdobeAuth, ...resolvedWebhook } =
+        operation.after as ResolvedWebhookPayload;
```
iivvaannxx: What is typed like that you need the type assertion?

packages/aio-commerce-lib-app/test/integration/management/webhooks.test.ts
iivvaannxx, line 1: Happy paths are covered, should we add some edge cases tests?

packages/aio-commerce-lib-app/source/management/domains/webhooks/helpers.ts
obarcelonap: since the identity is persisted we need to narrow the info in the object, otherwise it may contain more than needed.

packages/aio-commerce-lib-app/source/management/domains/webhooks/helpers.ts
obarcelonap: changed

packages/aio-commerce-lib-app/test/integration/management/webhooks.test.ts
obarcelonap: added additional coverage

packages/aio-commerce-lib-app/source/management/domains/webhooks/helpers.ts
obarcelonap: done

packages/aio-commerce-lib-app/source/management/domains/webhooks/helpers.ts
obarcelonap: `WebhookOperationValue` is shared between `add`'s `after` and `remove`'s `before`, and has to keep every non-identity field optional because a cleanup-driven remove only carries a bare identity (see `staleFromCleanup`). `kind === "add"` doesn't narrow that away — it's still the same `TValue`. In practice `add`'s `after` is always fully resolved (built in `resolveWebhookPayload`), so the assertion just bridges that invariant, which the shared type can't express.

iivvaannxx: For now let's leave it like this, but IMO this is a type contract issue. It means `before` and `after` don't always share the same TValue so the generic in the plan should be <TBefore, TAfter = TBefore> so that both types can be represented accurately. I'll apply this in #623 and you can address it in a future PR.

iivvaannxx (approved):
`helpers.ts` is quite big already. I would do similar to what #624 does and separate into a `plan.ts` and `apply.ts`

PR #624, opened 2026-08-11
CEXT-6505: Manage Event Additions and Removals During Application Upgrade

oshmyheliuk:
## Description

https://jira.corp.adobe.com/browse/CEXT-6505

Adds upgrade `plan`/`apply` support to the eventing domain so that app **upgrades** (not just fresh installs) reconcile Adobe I/O Events and Commerce state to match the target config.

Each eventing leaf (Commerce and external) now diffs the installed baseline against the target configuration and converges the difference:

- **Provider added (with a new provider)** → creates the I/O Events provider, its event metadata, registrations, and — for Commerce — the Commerce provider and event subscriptions.
- **Provider removed (last event of a provider dropped)** → tears the whole provider down (unsubscribe from Commerce, remove the Commerce provider, remove the I/O Events registration, provider, and event metadata), reusing the existing `uninstall` path.
- **Event added to an existing provider** → registers the new metadata (and Commerce subscription); creates a new registration when the event targets a new runtime action, or updates the existing registration in place when it shares a runtime action.
- **Event removed while the provider persists** → deletes the event metadata (and Commerce subscription); removes the registration when it was the sole event on its runtime action, or updates the registration when other events still route to that action.

Also includes a small internal refactor extracting shared helpers (`eventCodeOf`, `partitionByKey`) into `utils.ts` to remove duplication between `plan.ts` and `apply.ts`.

## Related Issue

CEXT-6505 — Manage Event Additions and Removals During Application Upgrade

## Motivation and Context

Until now the eventing domain could install and uninstall event sources, but an app **upgrade** had no way to reconcile eventing state when the config changed between versions. Merchants who added or removed events/providers in a new app version would be left with stale registrations, orphaned metadata, and lingering Commerce subscriptions. This change makes upgrades converge I/O Events and Commerce state to exactly match the target config, for both Commerce and external event sources.

## How Has This Been Tested?

- Unit tests (Vitest) covering every add/remove/update scenario at both the `plan` and `apply` layers, for **both** Commerce and external event kinds:
  - Provider add / provider removal / whole-provider teardown.
  - Event add on a new runtime action (registration created) and on an existing action (registration updated).
  - Event removal with a shared runtime action (metadata/subscription removed, registration updated) and as the sole event on its action (registration removed).
  - External-specific invariants: no Commerce subscription operations, and non-`com.adobe.commerce.` event codes.
- `apply` assertions verify the concrete side effects (`updateRegistration`, `deleteRegistration`, `deleteEventMetadataForProvider`, `deleteEventSubscription`) via call-count/argument checks.
- Full package suite green: **1021 tests passing** (`pnpm --filter @adobe/aio-commerce-lib-app test`).
- `pnpm --filter @adobe/aio-commerce-lib-app typecheck` — clean.
- `pnpm --filter @adobe/aio-commerce-lib-app lint` (Biome + Redocly) — clean.

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
```diff
@@ -0,0 +1,581 @@
...
+      projectId: appData.projectId,
+      registrationId: registration.registration_id,
+      runtimeAction,
+      workspaceId: appData.workspaceId,
+    });
+    logger.info(
+      `Updated registration "${registration.name}" (action "${runtimeAction}") on provider "${providerData.label}".`,
+    );
+  } catch (error) {
+    const message = await unwrapHttpError(error);
```
obarcelonap: don't we need to throw and mark apply as failed? why we decide to continue in this case?

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
```diff
@@ -0,0 +1,581 @@
...
+      consumerOrgId: appData.consumerOrgId,
+      projectId: appData.projectId,
+      registrationId: registration.registration_id,
+      workspaceId: appData.workspaceId,
+    });
+    logger.info(
+      `Deleted registration "${registration.name}" (action "${runtimeAction}") from provider "${providerData.label}".`,
+    );
+  } catch (error) {
+    const message = await unwrapHttpError(error);
```
obarcelonap: I understand we can swallow this because action might be not present if code is already deployed, is that the case that fails? otherwise I am not sure if we swallow the error.

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
```diff
@@ -0,0 +1,556 @@
...
+      return `${kind}:registration:${value.providerKey}:${value.runtimeAction}`;
+    case "subscription":
+      return `${kind}:subscription:${value.name}`;
+    default:
+      return kind;
+  }
+}
+
+/** Accumulates the operations, cleanup resources and removed providers for one leaf. */
+class LeafPlanBuilder {
```
obarcelonap, line 111: I don't see this builder is considering `unresolvedCleanupResources`, we should deal with that somehow

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
oshmyheliuk: I was continuing with the "best effort" approach, but yes, it makes sense to fail in this and other similar cases. I will update the PR.

oshmyheliuk: Updated the code. Now the apply will fail if the registration wasn't updated. And also in some other cases.

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
oshmyheliuk: It makes sense to fail. The code was updated

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
oshmyheliuk: Added logic for dealing with `cleanupResources`

packages/aio-commerce-lib-app/source/management/domains/events/branch.ts
```diff
@@ -13,13 +13,49 @@
...
+  meta: {
+    ...externalEventsStep.meta,
+    upgrade: {
+      description:
+        "Reconciles external event providers, metadata and registrations",
+      label: "Update External Events",
+    },
+  },
+  plan: planExternalEvents,
+};
```
iivvaannxx: Is there a reason on creating new steps wrapping the old ones over just adding the new fields to the existing ones?

packages/aio-commerce-lib-app/source/management/domains/events/types.ts
```diff
@@ -141,3 +142,99 @@ export type StoredProviderEntry = {
...
+      providerLabel: string;
+      eventCode: string;
+    }
+  | {
+      resourceType: "registration";
+      providerKey: string;
+      providerLabel: string;
+      runtimeAction: string;
+    }
+  | { resourceType: "subscription"; name: string };
```
iivvaannxx: After removing cleanup resources you'll no longer need this most likely.

packages/aio-commerce-lib-app/source/management/domains/events/types.ts
```diff
@@ -141,3 +142,99 @@ export type StoredProviderEntry = {
...
+ */
+export type EventingDomainPlan = DomainPlan<
+  EventingOperationValue,
+  EventingCleanupIdentity
+> & {
+  /** The target app metadata, used to namespace event codes/names when onboarding. */
+  metadata: ApplicationMetadata;
+
+  /** The baseline app metadata, used to resolve deployed resources during teardown. Null on first upgrade. */
+  baselineMetadata: ApplicationMetadata | null;
```
iivvaannxx, line 209: Isn't this field provided to you by the orchestrator through the baseline config? If it isn't then I believe it's a gap, tell me if it's not and I make sure it does. Same goes for `metadata` property.

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
```diff
@@ -194,6 +205,47 @@ export function getIoEventCode(name: string, providerType: EventProviderType) {
...
+ * @param metadata - The application metadata used to namespace the event name.
+ * @param providerType - The type of the event provider.
+ */
+export function eventCodeOf(
+  event: AppEvent,
+  metadata: ApplicationMetadata,
+  providerType: EventProviderType,
+) {
+  return getIoEventCode(getNamespacedEvent(metadata, event.name), providerType);
+}
```
iivvaannxx, line 222: I think `resolveIoEventCode` will provide you the same functionality.

iivvaannxx (commented):
I am at mid-review of plan/apply logic, but made some comments that are most likely actionable already. Will submit another review once I checked the main logic.

packages/aio-commerce-lib-app/source/management/domains/events/branch.ts
oshmyheliuk: It was added to avoid creating of import cycle `commerce.ts` <-> `apply.ts`. The wrappers have been dropped https://github.com/adobe/aio-commerce-sdk/pull/624/changes/14e33a9d2a745714da11881d17dafefa1b78ab44

packages/aio-commerce-lib-app/source/management/domains/events/types.ts
oshmyheliuk: Yes, I removed those

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
oshmyheliuk: Yes, they do the same, but `eventCodeOf` is just adapted to the domain types for convenience.
It's called multiple times like this: `eventCodeOf(event, metadata, type)`. Using `resolveIoEventCode` would require some conversion on each call.
I would keep it, but can remove it if you prefer to.

packages/aio-commerce-lib-app/source/management/domains/events/types.ts
oshmyheliuk: I don't see that baselinMetadata or metadata is passed to the apply step

```
      const result = await step.apply(domainPlan, {
        ...context.lifecycleContext,
        ...accumulatedContext,
        attemptId: context.attemptId,
      });
```

So that's why it is currently passing through the plan.

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
```diff
@@ -0,0 +1,518 @@
...
+type PlanLeafInput = {
+  path: string[];
+  type: EventProviderType;
+  isCommerce: boolean;
+  targetSources: EventSource[];
+  targetMetadata: ApplicationMetadata | null;
+  baselineSources: EventSource[];
+  baselineSnapshot: EventingProviderSnapshot[] | null;
+  baselineMetadata: ApplicationMetadata | null;
+  env: ReturnType<typeof getInstallCommerceEnv>;
```
iivvaannxx: I recall we had a public type for `CommerceEnv`?

packages/aio-commerce-lib-app/source/management/domains/events/types.ts
```diff
@@ -141,3 +142,73 @@ export type StoredProviderEntry = {
...
+  events: AppEvent[];
+};
+
+/**
+ * The snapshot data an eventing leaf persists after applying its plan: the set of providers (with
+ * their deployed events) it currently owns. Serves as `baseline.data` for the next plan.
+ */
+export type EventingSnapshotData = {
+  providers: EventingProviderSnapshot[];
+};
```
iivvaannxx, line 164: I take these are "generic" (non-commerce/external) specific types for the snapshots. But can't we use `CommerceEventsStepData` and `ExternalEventsStepData` and create a union type of both to avoid duplicating types representing the same?

packages/aio-commerce-lib-app/source/management/domains/events/types.ts
```diff
@@ -141,3 +142,73 @@ export type StoredProviderEntry = {
...
+      providerKey: string;
+      type: EventProviderType;
+      runtimeAction: string;
+      eventCodes: string[];
+    }
+  | {
+      resourceType: "subscription";
+      providerKey: string;
+      name: string;
+    };
```
iivvaannxx, line 197: If the idea is to eventually render plans to the UI, which contain a set of operations to `apply` wouldn't we need all the data for each resource to be "displayed"? Like, for example, event subscription fields (in the case of subscription).

packages/aio-commerce-lib-app/source/management/domains/events/types.ts
iivvaannxx: I will fix this in the follow-up PR of tests and docs for orchestration. If we can fix it before merging it would be great; if not, we can refactor in a future PR.

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
```diff
@@ -0,0 +1,518 @@
...
+import type { EventsStepContext } from "./context";
+import type {
+  EventingDomainPlan,
+  EventingOperationValue,
+  EventingProviderSnapshot,
+  EventingSnapshotData,
+} from "./types";
+
+/** A config event source: a provider and its declared events. */
+type EventSource = { provider: EventProvider; events: AppEvent[] };
```
iivvaannxx, line 51: Should we use a union type of `CommerceEventSource` and `ExternalEventSource`?

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
```diff
@@ -0,0 +1,518 @@
...
+  /** Diffs the sub-resources of a provider present in both baseline and target. */
+  public diffPersistingProvider(
+    target: EventingProviderSnapshot,
+    baseline: EventingProviderSnapshot,
+    targetMetadata: ApplicationMetadata,
+    baselineMetadata: ApplicationMetadata,
+    isCommerce: boolean,
+  ): void {
+    // Provider display (label/description) has no in-place update API; a cosmetic change is
+    // intentionally left as-is rather than blocking the whole upgrade. Only sub-resources diff.
```
iivvaannxx: For the case of providers they do have a label and description columns as data no? Can we really consider that cosmetic?

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
iivvaannxx: I don't have a strong opinion. I guess `resolveIoEventCode` is slightly more testable because it doesn't force you to mock a full `metadata` object? If you think replacing this with `resolveIoEventCode` is not worth it, what about the opposite? Making `resolveIoEventCode` use `eventCodeOf`?

My main concern here is only about having duplicate code.

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
```diff
@@ -194,6 +205,47 @@ export function getIoEventCode(name: string, providerType: EventProviderType) {
...
+ * Partitions two keyed collections into the items unique to each side. `added` holds `target`
+ * items whose key is absent from `baseline`; `removed` holds `baseline` items whose key is absent
+ * from `target`. The two sides may key differently (e.g. namespaced under different metadata).
+ *
+ * @param target - The target-side items.
+ * @param baseline - The baseline-side items.
+ * @param targetKey - Derives the comparison key for a target item.
+ * @param baselineKey - Derives the comparison key for a baseline item.
+ */
+export function partitionByKey<T>(
```
iivvaannxx: Nice utility, but the name sounds a bit "unintuitive" to me. What about something like just `diff`? Or `diffKeys`? Or something similar, these are just suggestions. Also, did you know `Set` has a [`difference`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference) method? Maybe you can use it to compute this

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
```diff
@@ -0,0 +1,518 @@
...
+        type,
+      };
+
+      if (stringify(before) !== stringify(after)) {
+        this.update(
+          before,
+          after,
+          `Update registration: ${provider.label} → ${runtimeAction}`,
+        );
+      }
```
iivvaannxx, line 339: Isn't this PR only about additions/removals?

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
```diff
@@ -0,0 +1,518 @@
...
+  baselineMetadata: ApplicationMetadata | null;
+  env: ReturnType<typeof getInstallCommerceEnv>;
+};
+
+/** Reduces config event sources to env-scoped provider snapshots (skipping empty providers). */
+function sourcesToSnapshots(
+  sources: EventSource[],
+  type: EventProviderType,
+  env: ReturnType<typeof getInstallCommerceEnv>,
+): EventingProviderSnapshot[] {
```
iivvaannxx, line 71: Nothing super important, but I'd be careful with the wording to avoid confusion. I personally see "snapshots" as what we store after a lifecycle operation. If we also use this nomenclature here, maybe we can confuse the two?

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
```diff
@@ -0,0 +1,518 @@
...
+    baselineSources,
+    baselineSnapshot,
+    baselineMetadata,
+    env,
+  } = params;
+
+  const targetProviders = sourcesToSnapshots(targetSources, type, env);
+  const baselineProviders =
+    baselineSnapshot?.filter((snapshot) => snapshot.type === type) ??
+    sourcesToSnapshots(baselineSources, type, env);
```
iivvaannxx, line 413: Is this falling back to the baseline config if baselineSnapshot is not present? It's unlikely you'll have the former if the latter is not present. Not sure if it's correct either, since the baseline snapshot should always report a value. Right now, it does not because the `baselineProvider` is only looking in `lib-files` (eventually it should look in our backend I guess?). 

For now, I would make it throw if the snapshot does not exist.

I might also misunderstood the code, correct me if I'm wrong.

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
```diff
@@ -0,0 +1,518 @@
...
+    );
+  }
+
+  return {
+    kind: "planned",
+    plan: {
+      baselineMetadata,
+      baselineProviders,
+      // A plan always has at least one side, so at least one metadata is present.
+      metadata: (targetMetadata ?? baselineMetadata) as ApplicationMetadata,
```
iivvaannxx: This line is confusing to me. What is it used for if you don't know if it's the baseline or the target?

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  // 2. Converge every target provider. `install` is create-or-get, so it handles added providers,
+  //    added metadata, and registrations for newly declared runtime actions.
+  if (plan.targetProviders.length > 0) {
+    await options.install(
+      buildLeafConfig(plan.targetProviders, plan.metadata, options),
+      eventsContext,
+    );
+  }
```
iivvaannxx, line 114: This is smart. I would argue that, if we can, maybe instead of reusing the leaf logic, we extract the "install" and "uninstall" logic into a helper, like webhooks do, and reuse that instead?

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
```diff
@@ -0,0 +1,558 @@
...
+  events: AppEvent[],
+  type: EventProviderType,
+  metadata: ApplicationMetadata,
+): Set<string> {
+  return new Set(events.map((event) => eventCodeOf(event, metadata, type)));
+}
+
+/** Whether two string sets contain exactly the same members. */
+function areSameSets(a: Set<string>, b: Set<string>): boolean {
+  return a.size === b.size && [...a].every((value) => b.has(value));
```
iivvaannxx: Nitpick improvement with set operations that don't use arrays (slightly more space efficient):


```suggestion
  return a.size === b.size && a.isSubsetOf(b);
```

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
```diff
@@ -0,0 +1,558 @@
...
+      // biome-ignore lint/performance/noAwaitInLoops: registrations are updated sequentially to avoid an Adobe I/O Events rate-limit burst
+      await putRegistration(
+        providerData,
+        type,
+        runtimeAction,
+        events,
+        targetMetadata,
+        existingData,
+        context,
+      );
```
iivvaannxx, line 313: Similar to comment above, should we handle updates in this PR?

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
```diff
@@ -0,0 +1,558 @@
...
+    }
+  }
+
+  for (const runtimeAction of baselineActions.keys()) {
+    if (!targetActions.has(runtimeAction)) {
+      // biome-ignore lint/performance/noAwaitInLoops: registrations are deleted sequentially to avoid an Adobe I/O Events rate-limit burst
```
iivvaannxx, line 319: Is this rate-limit burst actually documented? How does `install` do it? Does it parallelize?

packages/aio-commerce-lib-app/test/unit/management/domains/events/plan.test.ts
iivvaannxx, line 1: Is there some code that we can re-use or move to `fixtures`?

iivvaannxx (commented):
Overall `apply` logic LGTM. Most of my concerns are around planning and the fact that we're mixing additions/removals with updates for registrations.

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
oshmyheliuk: `resolveIoEventCode` is kind of a public API already, so we shouldn't remove it. I wouldn't say that this is duplicated code, it's just two different wrappers for the same function.

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
oshmyheliuk: I wouldn't do this as Union, as inside they have different types of `events`: `CommerceEvent` and `ExternalEvent`. The shared logic needs to be a `AppEvent[]`

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
oshmyheliuk: It's not like a cosmetic change, but it doesn't have any functional impact. Do we want to recreate provider/metadata/event registrations if someone changed the provider description? 
I wouldn't do it. We can update this later if there's a request.
I will update the description https://github.com/adobe/aio-commerce-sdk/pull/624/changes/b51cacd5d9e6ce32752857249e535910c3a691f5

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
oshmyheliuk: Updated https://github.com/adobe/aio-commerce-sdk/pull/624/changes/b51cacd5d9e6ce32752857249e535910c3a691f5

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
iivvaannxx: Sounds reasonable; we should flag it on the ticket for Dani, at least. A comment or similar.

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
iivvaannxx: Oh no, didn't mean to remove it, just use it inside. I get your point, although I do not completely agree with it. Won't push it anymore, feel free to apply my suggestion or not. Resolving the comment.

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
oshmyheliuk: Renamed to diffByKeys https://github.com/adobe/aio-commerce-sdk/pull/624/changes/bb82916c1acbf79802f62cf803249f8f4b0e1319

Set.difference doesn't work in this case

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
oshmyheliuk: It's still about addition/removal for cases where events are added/removed to an event registration that already has other events.

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
oshmyheliuk: Changed metadata -> targetMetadata https://github.com/adobe/aio-commerce-sdk/pull/624/changes/d19833728b904468b753ab3d8313e5dda7f2ea5b

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
oshmyheliuk: I don't think it's correct to compare this to webhooks, since the webhook installation logic is much simpler than eventing.
Probably we can extract some "helpers", but I would prefer to do it in some follow-up PRs instead of reworking everything here.

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
oshmyheliuk: Similar to my answer, it's not an update as such, it's more about adding/removing events, which leads to a registration update. Without this, the task will be partially incomplete/broken logic.

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
oshmyheliuk: It's not documented, it's kind of a defensive assumption, as I remember, we faced a similar error during installation. Yes, the install is parallelized.

packages/aio-commerce-lib-app/test/unit/management/domains/events/plan.test.ts
oshmyheliuk: Moved some parts to fixtures https://github.com/adobe/aio-commerce-sdk/pull/624/changes/536b6307b132c788ca2db61eb5cb521ed8f97410

packages/aio-commerce-lib-app/source/management/domains/events/types.ts
oshmyheliuk: It can be easily enriched with the required data against real requirements, depends on what will be displayed. The details can be recovered from snapshots.

packages/aio-commerce-lib-app/source/management/domains/events/types.ts
oshmyheliuk: The `CommerceEventsStepData` and `ExternalEventsStepData` have different nesting. The goal isn't  to combine commerce + external into one list, it's to have a single provider-snapshot shape that both leaves reuse

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
oshmyheliuk: For me, naming is correct, as it's used in in-memory "snapshots". Can't be renamed if you think about better naming.

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
oshmyheliuk: If the app was installed but never upgraded, there is no snapshot, as it's only created by an upgrade. The fallback is to reconstruct the same shape from the baseline config, as the baseline config contains everything that is needed.
I can't throw here as it's probably going to fail the first upgrade.
Even if we add Snapshot saving on install, we still need a fallback for apps installed with prior versions of the SDK.

> eventually it should look in our backend I guess? 

Probably, if we decide to store it in the backend

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
iivvaannxx: Fair enough, might be well suited for [CEXT-6556](https://jira.corp.adobe.com/browse/CEXT-6556)

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
iivvaannxx: Makes sense, resolving 👍🏻

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
iivvaannxx: Got it, we can optimize in the future

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
iivvaannxx: Great 👍🏻. Thought it would work, but didn't check.

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
iivvaannxx: Nevermind then 👍🏻

packages/aio-commerce-lib-app/source/management/domains/events/plan.ts
iivvaannxx: > If the app was installed but never upgraded, there is no snapshot, as it's only created by an upgrade

Installation snapshots have always been there, and is what the orchestrator reads first. It's only optional in the type because we can't guarantee user didn't delete it, that's why I was saying it should probably throw. But maybe the throw doesn't belong here, so nevermind.

iivvaannxx (approved):

PR #625, opened 2026-08-12
CEXT-6506: Execute Event Configurations Changes During Application Upgrade

oshmyheliuk:
## Description

Reconciles **in-place configuration changes** to existing Commerce event subscriptions during an app upgrade. Previously the events diff matched subscriptions only by event name and emitted `add`/`remove` — an event in both baseline and target was never re-examined, so editing its `fields`/`rules`/`priority`/`hipaa_audit_required` and redeploying silently drifted the live subscription from config.

- **`@adobe/aio-commerce-lib-events`**: new `updateEventSubscription` client method wrapping the Commerce `PUT eventing/eventSubscribe/:name` (merge) endpoint.
- **`@adobe/aio-commerce-lib-app`**: the upgrade now detects subscription config changes (order-insensitive, normalized), emits a first-class `update` plan op, and applies them **two-tier** — additive/same-key changes in place via the merge endpoint; changes that remove or re-key a field/rule via unsubscribe→resubscribe (the registration survives, re-linking by event code). A failed update **fails the upgrade step** rather than drifting silently.

## Related Issue

[CEXT-6506](https://jira.corp.adobe.com/browse/CEXT-6506). Stacked on CEXT-6505 (event add/remove during upgrade).

## Types of changes

## Checklist:


.changeset/commerce-update-event-subscription.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-events": minor
+---
+
+Add `updateEventSubscription` to the Commerce Events API client, wrapping the `PUT eventing/eventSubscribe/:name` endpoint to update an existing subscription's fields and rules in place. The endpoint merges by key and cannot remove entries.
```
iivvaannxx: Implementation detail

```suggestion
Add `updateEventSubscription` to the Commerce Events API client, wrapping the `PUT eventing/eventSubscribe/:name` endpoint to update an existing subscription's fields and rules in place.
```

.changeset/eventing-subscription-config-updates.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": minor
+---
+
+Reconcile Commerce event subscription configuration changes during an app upgrade. When an event that exists in both the installed baseline and the target config changes its fields, rules, priority or HIPAA flag, the upgrade now updates the live subscription instead of leaving it stale: additive and same-key changes are applied in place, while changes that remove or re-key a field or rule are applied by re-subscribing the event.
```
iivvaannxx: Similar to the other changeset, I think it could be freed up of some implementation details.

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
```diff
@@ -938,6 +948,87 @@ async function removeDroppedSubscriptions(
   }
 }
 
+/**
+ * Reconciles configuration changes on Commerce subscriptions present on both the baseline and
+ * target. Additive/same-key changes are applied in place via the Commerce merge-update endpoint;
+ * orphaning changes (field/rule removal, rename, rule operator/field change) are applied by
+ * unsubscribe + resubscribe. Unlike the best-effort removals above, a failure here fails the
```
iivvaannxx: Reading this comment, I now noticed the `best-effort removals above` part. FYI, I recall Dani saying to me that uninstallation should not be best-effort, and if it fails, we should fail the operation. Of course, nothing to act on this PR, and you can check with him when he's back, but just so you're aware.

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
```diff
@@ -938,6 +948,87 @@ async function removeDroppedSubscriptions(
...
+    const name = getNamespacedEvent(targetMetadata, targetEvent.name);
+    const baselineEvent = baselineByName.get(name);
+    if (!baselineEvent) {
+      // Added event — created by the idempotent install pass.
+      continue;
+    }
+
+    const changeMode = getSubscriptionChangeKind(
+      baselineEvent as CommerceEvent,
+      targetEvent as CommerceEvent,
```
iivvaannxx: Probably you can do the `as CommerceEvent` when constructing the `Map` and you won't need doing it at every other line.

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
```diff
@@ -246,6 +251,80 @@ export function partitionByKey<T>(
...
+  ) {
+    return "none";
+  }
+
+  const baselineKeys = subscriptionMergeKeys(baseline);
+  const targetKeys = subscriptionMergeKeys(target);
+
+  const droppedKey =
+    [...baselineKeys.fields].some((key) => !targetKeys.fields.has(key)) ||
+    [...baselineKeys.rules].some((key) => !targetKeys.rules.has(key));
```
iivvaannxx: `isSubsetOf` might come handy here:


```suggestion
  const droppedKey =
    !baselineKeys.fields.isSubsetOf(targetKeys.fields) ||
    !baselineKeys.rules.isSubsetOf(targetKeys.rules);
```

packages/aio-commerce-lib-app/test/unit/management/domains/events/apply.test.ts
iivvaannxx, line 1: As usual I recommend moving all generic "fixture" code to `fixtures` so it can be reused.

packages/aio-commerce-lib-app/test/unit/management/domains/events/utils.test.ts
```diff
@@ -649,3 +651,135 @@ describe("removeStoredEventProviders", () => {
...
+      ],
+    });
+    expect(getSubscriptionChangeKind(baseline, target)).toBe("none");
+  });
+
+  test("treats an omitted optional flag as its default (no change)", () => {
+    const baseline = commerceEvent();
+    const target = commerceEvent({ priority: false });
+    expect(getSubscriptionChangeKind(baseline, target)).toBe("none");
+  });
```
iivvaannxx, line 705: While this makes sense, is it really the behavior we want? As a user might be confusing for me to change or add some value and don't see an operation for it?

iivvaannxx (commented):
LGTM for subscriptions, but what about updates to labels/descriptions that drive data of registrations/metadata and providers?

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
oshmyheliuk: Currently the best-effort is done only for `removeDroppedMetadata` for orphaned I/O metadata (which probably was already removed through the commerce unsubscribe), for providers/subscription there is proper `throws` that fail the `uninstall`.

.changeset/eventing-subscription-config-updates.md
oshmyheliuk: Removed implementation details

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
oshmyheliuk: Done https://github.com/adobe/aio-commerce-sdk/pull/625/changes/174de93536499334dd23907d78525fff7a419b23

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
oshmyheliuk: Applied with some changes as the biome check was failing

packages/aio-commerce-lib-app/test/unit/management/domains/events/apply.test.ts
oshmyheliuk: Moved some part to fixtures

packages/aio-commerce-lib-app/test/unit/management/domains/events/utils.test.ts
oshmyheliuk: The `priority: false` is a default value for event subscription, so if someone adds it to the event configuration between App Versions, it won't change anything, so such an operation can be skipped during the upgrade. I don't see a point in showing  it in the upgrade process if nothing will be changed.
But yes, it can be changed if we want to display it.

oshmyheliuk:
> LGTM for subscriptions, but what about updates to labels/descriptions that drive data of registrations/metadata and providers?

For now, I would leave as it is as there is no proper API to update it. We can discuss it with @danperezAdobe and decide if we want to recreate everything if the label/description changed. If yes, I will do it in the separate PR

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
```diff
@@ -540,6 +551,84 @@ async function removeDroppedSubscriptions(
...
+          destination: event.destination,
+          fields: event.fields,
+          force: event.force,
+          hipaa_audit_required: event.hipaa_audit_required,
+          name,
+          parent: event.name,
+          priority: event.priority,
+          provider_id: providerId,
+          rules: event.rules,
+        });
```
iivvaannxx: Can't some bits of `event` be spreaded?

iivvaannxx (approved):

mariam21arauj (approved):

iivvaannxx:
@oshmyheliuk The [`eventing`](https://github.com/adobe/aio-commerce-sdk/blob/ce81962018a7d3ead168952dc4911f8f283f6418/packages/aio-commerce-lib-app/source/management/domains/events/branch.ts) parent branch does not define `meta.upgrade`. This will crash at runtime when executing the tree, you should add it.

packages/aio-commerce-lib-app/source/management/domains/events/apply.ts
oshmyheliuk: Refactored by moving subscription into separate variable

oshmyheliuk:
> @oshmyheliuk The [eventing](https://github.com/adobe/aio-commerce-sdk/blob/ce81962018a7d3ead168952dc4911f8f283f6418/packages/aio-commerce-lib-app/source/management/domains/events/branch.ts) parent branch does not define meta.upgrade. This will crash at runtime when executing the tree, you should add it.

Added

PR #626, opened 2026-08-12
CEXT-6509 Manage Admin UI Component Additions and Removals During Application Upgrade

vinayrao2000:
## Description

Adds upgrade `plan`/`apply` support to the **Admin UI** domain so that application **upgrades** (not just fresh installs/uninstalls) reconcile the registered Admin UI extension to match the target configuration.

`planAdminUi` diffs the installed baseline against the target config **component-by-component** — the top-level `menu`, per-entity `gridColumns`, per-`id` `massActions`, and order `viewButtons` — emitting one `add`/`remove` operation per component that appeared or disappeared. Because Commerce registers the extension as a single unit (no per-component API), the operations collapse to one whole-extension `AdminUiExtensionAction` on the plan:

- **Component added** (block newly declared) → `register` (enable SDK → register).
- **Component added/removed while the block persists** → `refresh` (re-register to re-sync from the App Registry).
- **All Admin UI removed** → `unregister` (validated, 404-tolerant `DELETE`).

Also includes: a strict `unregisterExtensionForUpgrade` for the upgrade removal path (throws on any non-404 failure, treats 404 as already-removed), a `refreshExtension` wrapper (interim re-POST until CEXT-6559's `/refresh` ships), and a graceful `blocked` result when the runtime namespace is unavailable rather than throwing and crashing the planning pass.

## Related Issue

[CEXT-6509](https://jira.corp.adobe.com/browse/CEXT-6509) — Manage Admin UI Component Additions and Removals During Application Upgrade.

## Motivation and Context

The Admin UI domain could install and uninstall an app's extension, but an **upgrade** had no way to reconcile Admin UI state when the config changed between versions. A developer who added a menu entry, dropped an order-grid column, or added a mass action in a new app version would leave the merchant with an admin that no longer matches the deployed code. This change makes upgrades converge the registered extension to exactly match the target config, as a sibling of the merged webhooks upgrade (#621) and the in-flight eventing upgrade (#624).

## How Has This Been Tested?

- **Unit tests (Vitest)** covering every path at the `plan` and `apply` layers:
  - `plan`: `register` / `refresh` / `unregister` classification, per-component add/remove detection, modification-skip (no op for a component changed on both sides), empty-block no-op, duplicate-id disambiguation, and a `blocked` result when `__OW_NAMESPACE` is unavailable.
  - `apply`: enable-then-register on `register`, re-register on `refresh`, strict `unregister` (throws on non-404, tolerates 404), no-op, and cleanup-resource reuse from the plan.
- Full package suite green (`pnpm --filter @adobe/aio-commerce-lib-app test`), `typecheck` clean, `lint` (Biome + Redocly) clean.
- **Manual E2E** against a live Stage Commerce instance (self-deployed app), covering the full matrix:
  - modify a column *inside* gridColumns → correctly skipped (no change);
  - drop `order.gridColumns` → columns removed via `refresh`;
  - restore `order.gridColumns` → columns re-added via `refresh`;
  - drop the whole `adminUi` block → extension `unregister`ed (menu + columns gone);
  - re-add the block → `register`ed fresh;
  - add `product.gridColumns` (multi-entity) and an order `massAction` (array component) → both appeared;

N/A

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
```diff
@@ -83,22 +132,87 @@ export async function unregisterExtension(
...
+ */
+export async function unregisterExtensionForUpgrade(
+  context: AdminUiExecutionContext,
+): Promise<void> {
+  const { logger } = context;
+  const extensionName = process.env.__OW_NAMESPACE;
+
+  if (!extensionName) {
+    throw new Error("__OW_NAMESPACE environment variable is not set");
+  }
```
oshmyheliuk: The part that extracts the extension name from `process.env.__OW_NAMESPACE` can be moved to a separate function and reused. That logic is already used several times in this file.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
```diff
@@ -62,6 +69,48 @@ export async function registerExtension(context: AdminUiExecutionContext) {
...
+ *
+ * Interim implementation: re-POSTs the registration, which is idempotent and
+ * re-syncs the registrations without duplicating the extension. Swap to the
+ * dedicated `POST .../refresh` endpoint (CEXT-6559) once it ships to instances.
+ *
+ * @param context - The execution context providing the Admin UI client and logger.
+ */
+export function refreshExtension(context: AdminUiExecutionContext) {
+  return registerExtension(context);
+}
```
oshmyheliuk, line 126: The call to the refresh endpoint can already be implemented in this task. In any case, we need a fallback if the endpoint doesn't exist, as it's not guaranteed that all commerce instances will be using the latest version of the Admin UI SDK module.

oshmyheliuk: > endpoint (CEXT-6559) once it ships to instances

We don't control this in PaaS, only in ACCS. So no need to wait.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
```diff
@@ -83,22 +132,87 @@ export async function unregisterExtension(
...
+    );
+  }
+}
+
+/**
+ * Resolves the identity of this app's Admin UI extension from context, or `null`
+ * when the deployment namespace (`__OW_NAMESPACE`) is not available. Used by
+ * `plan`, which must not throw — a missing namespace is surfaced as a blocking
+ * issue instead of crashing the whole planning pass.
+ */
```
oshmyheliuk: I don't think that JSDoc for the function should explain who uses it. The JSDoc should explain what the function is doing. The same for `resolveExtensionIdentity`.

If you specify that this function is used by `plan` maybe this logic should be moved directly to the plan.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
vinayrao2000: Done, updated:

- Added refreshExtension to aio-commerce-lib-admin-ui and wired it in, the plan/apply now calls the real POST .../refresh endpoint directly.
- Added a fallback: on a 404 it falls back to re-registering. And per your note, this fallback is permanent, not a "swap once merged" thing, since PaaS instances won't all pick up the new module version the way ACCS does.
- One wrinkle: the refresh endpoint returns no extensionId, so I carry the baseline's id forward (baselineExtensionId) into the snapshot when refresh succeeds.

Fixed in c074d18a520680338446c2772447cbdd835a213e

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
vinayrao2000: Done, extracted getExtensionName() and requireExtensionName(), and switched all four process.env.__OW_NAMESPACE reads in this file over to them.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
vinayrao2000: Done, Fixed.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
```diff
@@ -83,22 +181,81 @@ export async function unregisterExtension(
...
+    | ValidationExecutionContext<AdminUiStepContext>
+    | ApplyContext<AdminUiStepContext>,
+): AdminUiIdentity {
+  const identity = tryResolveExtensionIdentity(context);
+  if (!identity) {
+    throw new Error("__OW_NAMESPACE environment variable is not set");
+  }
+
+  return identity;
+}
```
oshmyheliuk: I don't see why two almost identical functions are needed. You can leave only one, which is used in the plan, and pass identity as part of the plan, so no need to call one of the functions in `apply` and it can be dropped.

vinayrao2000: Done, apply now reads plan.identity (resolved once during planning) instead of re-resolving it.

Fixed in 29972b3f8f482f21b3e024965bfe3c18ebc0a5a6

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
```diff
@@ -62,6 +82,84 @@ export async function registerExtension(context: AdminUiExecutionContext) {
...
+
+  try {
+    await adminUiClient.refreshExtension({
+      extensionName,
+      workspaceName: appData.workspaceName,
+    });
+    logger.info(
+      `Admin UI extension "${extensionName}" refreshed successfully.`,
+    );
+    return { extensionId: null };
```
iivvaannxx: What is the value on always returning a `null` value? If the endpoint does not return this back just do not return.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
```diff
@@ -11,10 +11,34 @@
...
 import { throwHttpError } from "#management/common/utils/http-error";
 
-import type { AdminUiExecutionContext } from "./utils";
+import type { ApplyContext } from "#management/common/workflow/resource";
+import type { ValidationExecutionContext } from "#management/common/workflow/step";
+import type { AdminUiIdentity } from "./types";
+import type { AdminUiExecutionContext, AdminUiStepContext } from "./utils";
+
+/** HTTP status Commerce returns when the extension to delete no longer exists. */
+const HTTP_NOT_FOUND = 404;
```
iivvaannxx: We have codes already exported from `lib-core` or `lib-api`, I don't recall now.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/types.ts
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Snapshot data persisted after a successful Admin UI registration or refresh.
+ * `extensionId` is `null` when the dedicated refresh endpoint handles the sync:
+ * it re-syncs registrations without returning one, and there is no read endpoint
+ * to fetch it back, so the baseline's id is carried forward instead (see
+ * {@link AdminUiDomainPlan.baselineExtensionId}).
+ */
+export type AdminUiSnapshotData = {
+  extensionId: string | null;
+};
```
iivvaannxx: Use `InferStepOutput` type on the `adminUi` step definition, similar to events.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/types.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/** The Commerce entity an Admin UI component attaches to (absent for the menu). */
+export type AdminUiComponentEntity = "customer" | "order" | "product";
+
+/** The addressable kinds of Admin UI component an upgrade operation can target. */
+export type AdminUiComponentKind =
+  | "grid-columns"
+  | "mass-action"
+  | "menu"
+  | "view-button";
```
iivvaannxx: Probably this should be derived from the `schema` in the `config` module?

packages/aio-commerce-lib-app/source/management/domains/admin-ui/types.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/** The `adminUi.order` shape — the superset entity (grid columns, mass actions, view buttons). */
+type AdminUiOrderEntity = NonNullable<AdminUi["order"]>;
+
+/** The config of a single Admin UI component, carried on an operation for review and auditing. */
+export type AdminUiComponentConfig =
+  | NonNullable<AdminUi["menu"]>
+  | NonNullable<AdminUiOrderEntity["gridColumns"]>
+  | NonNullable<AdminUiOrderEntity["massActions"]>[number]
+  | NonNullable<AdminUiOrderEntity["viewButtons"]>[number];
```
iivvaannxx: Probably is also a type belonging into the admin ui schema module, potentiatally derivable also.

iivvaannxx (commented):
I'm checking `plan` and `apply` logic but maybe you can act on the below comments now

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
vinayrao2000: Done fixed. refreshExtension now returns undefined when there's no id to report, instead of always returning { extensionId: null }.

Fixed in 6e1e3da3655b5cc855cc903aa93dbe51625b3413

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
vinayrao2000: Done, now importing HTTP_NOT_FOUND from @adobe/aio-commerce-lib-api/utils instead of defining it locally.

Fixed in 6e1e3da3655b5cc855cc903aa93dbe51625b3413

packages/aio-commerce-lib-app/source/management/domains/admin-ui/types.ts
vinayrao2000: Fixed. with one deviation: InferStepOutput<typeof registerExtensionStep> doesn't compile here, unlike the events steps, this one has an apply handler whose return type already references RegisterExtensionStepData, so deriving the type from the step itself is self-referential.

Used Awaited<ReturnType<typeof registerExtension>> instead, same intent (auto-derived, no hand-maintained shape), just sourced from the underlying function rather than the step wrapper, which doesn't have that cycle.

Fixed in 5cefdc5a742cf2dbf598f1e6cc219baa85056d61

packages/aio-commerce-lib-app/source/management/domains/admin-ui/types.ts
vinayrao2000: Done. AdminUiComponentEntity is now Exclude<keyof AdminUi, "menu"> derived from the schema. AdminUiComponentKind stays hand-written: its kebab-cased values (grid-columns, mass-action, etc.) don't match the schema's own field names (gridColumns, massActions), so there's nothing to derive it from.

Fixed in 8302f53ceb5215c60e87255587445f7b53bf0bb8

packages/aio-commerce-lib-app/source/management/domains/admin-ui/types.ts
vinayrao2000: Done moved to config/schema/admin-ui.ts, next to AdminUi.

Fixed in 8302f53ceb5215c60e87255587445f7b53bf0bb8

packages/aio-commerce-lib-app/source/config/schema/admin-ui.ts
```diff
@@ -233,6 +233,19 @@ export type AdminUiConfiguration = v.InferInput<typeof AdminUiSchema>;
...
+
+/**
+ * The validated config of a single Admin UI component: the menu, an entity's
+ * grid columns, a single mass action, or a single view button.
+ */
+export type AdminUiComponentConfig =
+  | NonNullable<AdminUi["menu"]>
+  | NonNullable<AdminUiOrderEntity["gridColumns"]>
+  | NonNullable<AdminUiOrderEntity["massActions"]>[number]
+  | NonNullable<AdminUiOrderEntity["viewButtons"]>[number];
```
iivvaannxx: I think this should use `v.InferOutput` on the individual schemas of `MenuSchema`, `MassActionSchema`... Instead of "selecting" from `AdminUi` type.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/types.ts
iivvaannxx: Yes, my point was to derive from config schema with `camelCase` not to rewrite with `kebab-case`

packages/aio-commerce-lib-app/source/management/domains/admin-ui/types.ts
```diff
@@ -0,0 +1,73 @@
...
+export type AdminUiOperationValue = {
+  component: AdminUiComponentRef;
+  config?: AdminUiComponentConfig;
+};
+
+/**
+ * The single whole-extension action `apply` performs to converge a plan's
+ * component operations. Commerce exposes no per-component API, so the granular
+ * adds and removes collapse to one call: `register` (first-time), `refresh`
+ * (re-sync an existing extension), or `unregister` (drop it). `null` is a no-op.
```
iivvaannxx:
```suggestion
 * (re-sync an existing extension), or `unregister` (drop it).
```

packages/aio-commerce-lib-app/source/management/domains/admin-ui/plan.ts
```diff
@@ -0,0 +1,298 @@
...
+    return baseKey;
+  }
+
+  let occurrence = 2;
+  while (components.has(`${baseKey}#${occurrence}`)) {
+    occurrence += 1;
+  }
+  return `${baseKey}#${occurrence}`;
+}
+
```
iivvaannxx: This is a patch over a "bug" in the config schema of admin ui. The valibot schemas should ensure that the IDs in the arrays are unique so that you don't need to disambiguate here. I suggest doing that change and tracking it with a separate `patch` changeset (it would be `patch` instead of a `major` because it's not supposed to behave like it's doing, and most likely Commerce would error un such a situation)

packages/aio-commerce-lib-app/source/management/domains/admin-ui/plan.ts
```diff
@@ -0,0 +1,298 @@
...
+  if (adminUi.menu) {
+    components.set("menu", {
+      config: adminUi.menu,
+      key: "menu",
+      label: "menu",
+      ref: { kind: "menu" },
+    });
+  }
+
+  for (const entity of ["order", "product", "customer"] as const) {
```
iivvaannxx: The inline const array could be extracted to a constant.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/branch.ts
```diff
@@ -16,23 +16,25 @@ import {
...
- * {@link registerExtensionStep} so Commerce accepts the extension registration.
- * Install-only: enabling the SDK is not reverted on uninstall since other
- * extensions may still rely on it.
+ * Leaf step that enables the Admin UI SDK (PUT) on install, before
+ * {@link registerExtensionStep}. Install-only: it has no uninstall handler.
  */
 const enableAdminUiSdkStep = defineLeafStep({
+  // Must run before registerExtensionStep so Commerce accepts the extension
+  // registration. Not reverted on uninstall since other extensions may still
+  // rely on the SDK being enabled.
```
iivvaannxx: Let's try to not add unnecessary comments that have nothing to do with the PR. It only adds noise.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/branch.ts
```diff
@@ -57,17 +63,35 @@ const registerExtensionStep = defineLeafStep({
...
+// endpoint returns none, and there is no read endpoint), but the extension is
+// still registered whenever this snapshot is persisted at all. Recording
+// `{ extensionId: null }` here (rather than persisting no snapshot at all)
+// keeps that distinct from "not registered".
+/** The register step's output (`{ extensionId }`), also persisted as the domain's snapshot data. */
+export type RegisterExtensionStepData = {
+  extensionId:
+    | Awaited<ReturnType<typeof registerExtension>>["extensionId"]
+    | null;
+};
```
iivvaannxx: If `InferStepOutput` can not be used, this should be the below
```suggestion
export type RegisterExtensionStepData = Awaited<ReturnType<typeof registerExtension>>;
```

I honestly don't understand why do we need to widen `extensionId` to return `null`? The type is meant to describe what `registerExtension` returns, it has nothing to do with `refresh`. And `registerExtension` returns that id always.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
```diff
@@ -83,22 +177,52 @@ export async function unregisterExtension(
...
+  } catch (error: unknown) {
+    if (
+      error instanceof HTTPError &&
+      error.response.status === HTTP_NOT_FOUND
+    ) {
+      logger.info(
+        `Admin UI extension "${extensionName}" was already absent; treating removal as complete.`,
+      );
+      return;
+    }
```
iivvaannxx, line 211: Are you sure that unregistering an extension returns a 404 if the extension is already removed? I would double check that in the backend uix module.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/plan.ts
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  return components;
+}
+
+/** Reads the baseline's persisted `extensionId`, or `null` when there is no baseline. */
+function extractBaselineExtensionId(
+  baseline: { data: RegisterExtensionStepData | null } | null,
+): string | null {
+  return baseline?.data?.extensionId ?? null;
```
iivvaannxx, line 110: Double check that you're accessing `data` well. The output data of steps are stored in the snapshots using their path. See `setAtPath` on the workflow runner to understand what I mean.

This might be already correct, it's just a heads-up.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/plan.ts
```diff
@@ -0,0 +1,298 @@
...
+    };
+  }
+
+  return {
+    before: value,
+    id: "remove:extension",
+    kind: "remove",
+    label: "Unregister Admin UI extension",
+  };
+}
```
iivvaannxx: I have mixed feelings about this. Why treat the extension as an entity? You're either adding a `menu`, a `massAction`, a `viewOrderButton` or a `gridColumn`. The extension comes implicit with those. Don't quite understand why do we need to do this?

packages/aio-commerce-lib-app/source/management/domains/admin-ui/plan.ts
```diff
@@ -0,0 +1,298 @@
...
+
+  const operations = diffComponents(baselineComponents, targetComponents);
+
+  // The extension is registered in Commerce whenever the `adminUi` block is
+  // present (`registerExtensionStep.install` runs on `hasAdminUi`, even with zero
+  // components), so its lifecycle tracks block presence, not component count:
+  // an appearing block registers, a disappearing block unregisters, and a block
+  // present on both sides refreshes only when a component actually changed.
+  const baselineHasBlock = Boolean(baselineAdminUi);
+  const targetHasBlock = Boolean(targetAdminUi);
```
iivvaannxx: I would not say the fact that `install` runs even if there are no components is acceptable. Sounds like a bug to me. I would not model this PR based on that.

packages/aio-commerce-lib-app/test/unit/management/domains/admin-ui/plan.test.ts
iivvaannxx, line 1: See if some of the code here is worth moving to `fixtures`. But do it rationally, not indiscriminately.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/apply.ts
```diff
@@ -0,0 +1,75 @@
...
+    await enableAdminUiSdk(context);
+    const { extensionId } = await registerExtension(context);
+    return { snapshotData: { extensionId } };
+  }
+
+  if (plan.extensionAction === "refresh") {
+    // Enable defensively: refresh (and the re-register fallback) require the SDK
+    // to be enabled, and Commerce rejects them otherwise. Enabling is an
+    // idempotent PUT, so it guards against the SDK being disabled out-of-band at
+    // no meaningful cost.
```
iivvaannxx: AI tends to add very noisy comments. Please consider trimming those down when a shorter explanation can do the same job.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/apply.ts
```diff
@@ -0,0 +1,75 @@
...
+    await enableAdminUiSdk(context);
+    const refreshed = await refreshExtension(context);
+    // The dedicated refresh endpoint returns no id (and there is no read endpoint
+    // to fetch one), so carry the baseline's id forward when there is nothing
+    // fresher to record. Always persist a snapshot here, even with a null id:
+    // `extensionAction` is only "refresh" when the baseline block was present,
+    // so the extension is registered regardless of whether its id is known —
+    // persisting no snapshot at all would misrepresent that as unregistered.
```
iivvaannxx: Same goes for this comment.

iivvaannxx (commented):
An empty adminUi block should be treated the same as no Admin UI config. Then we can base register/unregister on whether components exist and drop the synthetic extension component.

Probably you can fix `install` requiring at least a component before running, seems it's a related concern.

This PR works around a lot, trying to respect that when it's not very correct IMO.

.changeset/admin-ui-upgrade-additions-removals.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": minor
+---
+
+Add upgrade planning support for Admin UI component additions and removals: the Admin UI step now diffs the installed baseline against the target configuration component-by-component, registering newly added components and removing dropped ones, and unregistering the extension when Admin UI is removed entirely. Configuration changes to existing components are left untouched, and the removal path is validated so failures surface instead of being silently ignored.
```
iivvaannxx: No need to document a changeset with a negative (don't state what this PR doesn't do).

```suggestion
Add upgrade planning support for Admin UI component additions and removals: the Admin UI step now diffs the installed baseline against the target configuration component-by-component, registering newly added components and removing dropped ones, and unregistering the extension when Admin UI is removed entirely.
```

packages/aio-commerce-lib-app/source/config/schema/admin-ui.ts
vinayrao2000: Done, derived AdminUiComponentConfig from the individual schemas via v.InferOutput 

Fixed in 462c4feb9fb34f9c26d68720e0d106a197a7ed6b

mariam21arauj (approved):

packages/aio-commerce-lib-app/source/management/domains/admin-ui/plan.ts
vinayrao2000: Done, enforced unique ids at the schema level (v.check on the massActions/viewButtons arrays), so disambiguateKey is gone; components now key directly on their id. Tracked with a separate patch changeset as suggested, and added schema tests for the duplicate-id rejection.

Fixed in 8a98136d3a4099763a7650c7299d1be0956764a9

packages/aio-commerce-lib-app/source/management/domains/admin-ui/branch.ts
vinayrao2000: Done, RegisterExtensionStepData is now Awaited<ReturnType<typeof registerExtension>>, no null widening. Moved the "no id" case to apply, which persists a null snapshot instead of { extensionId: null }.

Fixed in a8e520f9bff3056ef9c50df42cff0b1a220598d6

packages/aio-commerce-lib-app/source/management/domains/admin-ui/plan.ts
vinayrao2000: Double-checked, it's correct. The runner already applies getAtPath(baseline.data, path) in planStep before calling plan, so input.baseline.data is this step's own slice ({ extensionId }), mirroring the setAtPath write in execute. Same access pattern as the webhooks domain (baseline.data?.subscribedWebhooks).

packages/aio-commerce-lib-app/source/management/domains/admin-ui/helpers.ts
vinayrao2000: Verified in the backend uix module: deleteByWorkspaceAndExtensionName → SelectedExtensionRepository throws NoSuchEntityException when affectedRows === 0, which Magento maps to HTTP 404.

packages/aio-commerce-lib-app/source/management/domains/admin-ui/plan.ts
vinayrao2000: Agreed, and removed. The extension op only existed because an empty adminUi: {} block registered on install, so a later drop needed something to carry the unregister past the engine's operations.length > 0 gate. I fixed it at the source instead: hasAdminUi now requires at least one component, so a component-less block.

Fixed in 1d2d23d24cd9a364acd381e34e4da934c0ddaf02

packages/aio-commerce-lib-app/source/management/domains/admin-ui/plan.ts
vinayrao2000: Done, fixed at the source instead of modeling around it. hasAdminUi now requires at least one component, so an empty adminUi: {} no longer registers on install. The block-presence logic is gone; register/unregister/refresh key purely on component count.

Fixed in 1d2d23d24cd9a364acd381e34e4da934c0ddaf02

packages/aio-commerce-lib-app/test/unit/management/domains/admin-ui/plan.test.ts
vinayrao2000: Done. Fixed in 6355393801e2a156a091ca943030c79654388b88

packages/aio-commerce-lib-app/source/management/domains/admin-ui/plan.ts
vinayrao2000: Done, extracted to a module-level ADMIN_UI_ENTITIES constant.

Fixed in c72017423ea45cf7e14ab8586aec218e8f8ae4f9

iivvaannxx (approved):

oshmyheliuk (approved):

PR #639, opened 2026-08-18
Fix generate/actions CWD-race flake; thread explicit project root

iivvaannxx:
## Description

Fixes the intermittent `generate/actions` test failure caused by a timed-out TypeScript bundle leaving the process working directory inside a deleted temporary project.

The change also establishes a consistent project-root boundary across the affected commands:

- Parameterless `exec()` entrypoints resolve the project root from the current working directory.
- Internal `run()` functions and leaf helpers require an explicit resolved `projectRoot` and do not resolve it again.
- `init` creates or discovers `package.json` before resolving the root, which supports greenfield projects.
- `generate all` intentionally composes the individual parameterless command entrypoints.
- `post-app-deploy` remains root-free because it does not access project files.
- The new `pre-app-dev` and `pre-app-run` hooks resolve the root before updating the project `.env`.
- Environment helpers and the auth IMS synchronization command follow the same boundary/core split.

The tests now use `withTempFiles` when exercising root-injected internal functions and reserve `withTempProject` for parameterless command boundaries that intentionally discover the root from CWD. `withTempProject` overrides `process.cwd()` instead of changing the OS working directory, while `withChdir` is hardened so an interrupted test cannot cascade `ENOENT` failures through the suite.

JSDoc has been updated to describe the explicit root contracts and current command behavior.

## Related Issue

No tracked issue. This fixes a CI flake observed on `main`.

## Motivation and Context

The affected tests ran `esbuild` for real while the process was chdir'd into a temporary project. Under CI load, bundling could exceed Vitest's five-second timeout. The temporary directory was then removed before the process working directory was safely restored, causing later tests to fail with `ENOENT`.

Increasing the bundle-test timeout removes the immediate trigger, and hardening `withChdir` contains interrupted callers. Passing a resolved project root through internal command code removes the underlying dependency on mutable global process state and prevents repeated root discovery.

## How Has This Been Tested?

- `@adobe/aio-commerce-lib-app`: 1,027 tests passing
- `@adobe/aio-commerce-lib-auth`: 126 tests passing
- `@aio-commerce-sdk/scripting-utils`: 156 tests passing
- `@adobe/aio-commerce-lib-config`: 212 tests passing
- Typechecks pass for app, auth, scripting-utils, and config
- Biome passes on all changed TypeScript files

Not applicable.

## Types of changes

## Checklist:

### DIFF
--- a/packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
+++ b/packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
@@ -17,7 +17,6 @@ import { dirname, extname, join, relative, sep } from "node:path";
 import {
   detectPackageManager,
   getPackageExecutionCommand,
-  getProjectRootDirectory,
   loadPackageJson,
   makeOutputDirFor,
 } from "@aio-commerce-sdk/scripting-utils/project";
@@ -145,10 +144,9 @@ async function writeJavaScriptAppConfigModule(
 async function bundleTypeScriptAppConfigModule(
   configFilePath: string,
   outputPath: string,
-  projectRoot?: string,
+  projectRoot: string,
 ) {
-  const root = projectRoot ?? (await getProjectRootDirectory());
-  const packageManager = await detectPackageManager(root);
+  const packageManager = await detectPackageManager(projectRoot);
 
   const { command, args } = getPackageExecutionCommand(
     packageManager,
@@ -167,7 +165,7 @@ async function bundleTypeScriptAppConfigModule(
   );
 
   const result = spawnSync(command, args, {
-    cwd: root,
+    cwd: projectRoot,
     stdio: "inherit",
   });
 
@@ -187,14 +185,13 @@ async function bundleTypeScriptAppConfigModule(
  * Write an ESM module that re-exports the static app manifest JSON so generated
  * actions can import it via the alias without needing a JSON import attribute.
  */
-async function prepareStaticAppConfigImportAlias(projectRoot?: string) {
-  const root = projectRoot ?? (await getProjectRootDirectory());
+async function prepareStaticAppConfigImportAlias(projectRoot: string) {
   const runtimeConfigPath = getRuntimeAppConfigPath();
-  const outputPath = join(root, runtimeConfigPath);
+  const outputPath = join(projectRoot, runtimeConfigPath);
 
-  await makeOutputDirFor(dirname(runtimeConfigPath), root);
+  await makeOutputDirFor(dirname(runtimeConfigPath), projectRoot);
   await writeJavaScriptAppConfigModule(
-    join(root, getManifestPath()),
+    join(projectRoot, getManifestPath()),
     outputPath,
   );
 }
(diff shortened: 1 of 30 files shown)


packages/aio-commerce-lib-app/source/commands/generate/actions/main.ts
```diff
@@ -35,55 +36,71 @@ import {
 
 import type { CommerceAppConfigOutputModel } from "#config/schema/app";
 
+/** Options for the generate actions command. */
+export type RunOptions = {
+  /** Working directory to resolve the project root from. Defaults to the CWD. */
+  cwd?: string;
+  /** The directory to load templates from, for testing purposes. Defaults to the generated actions template root. */
+  templatesDir?: string;
+};
```
oshmyheliuk: This is identical to `PostinstallOptions` and `PreAppBuildOptions`. Why can't it be one shared type?

packages/aio-commerce-lib-app/source/commands/generate/schema/main.ts
```diff
@@ -48,7 +51,7 @@ export async function run(appConfig: CommerceAppConfigOutputModel) {
     return;
   }
 
-  const projectDir = await getProjectRootDirectory();
+  const projectDir = projectRoot ?? (await getProjectRootDirectory());
```
oshmyheliuk: `const root = projectRoot ?? (await getProjectRootDirectory());` is used in 9 different places, only here it's 
`const projectDir = projectRoot ?? (await getProjectRootDirectory());`.
For consistency, I would rename `projectDir` -> `root`

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
```diff
@@ -145,9 +145,10 @@ async function writeJavaScriptAppConfigModule(
 async function bundleTypeScriptAppConfigModule(
   configFilePath: string,
   outputPath: string,
+  projectRoot?: string,
 ) {
-  const projectRoot = await getProjectRootDirectory();
-  const packageManager = await detectPackageManager(projectRoot);
+  const root = projectRoot ?? (await getProjectRootDirectory());
```
oshmyheliuk: `const root = projectRoot ?? (await getProjectRootDirectory());` repeats multiple times across the code. It makes sense to extract it to a separate function 
```
export async function resolveProjectRoot(projectRoot?: string) {
  return projectRoot ?? (await getProjectRootDirectory());
}
```

packages/aio-commerce-lib-app/source/commands/hooks/postinstall.ts
```diff
@@ -20,13 +21,22 @@ import { loadAppManifest } from "#commands/utils";
...
 export async function run(
   appManifest: CommerceAppConfigOutputModel,
-  templatesDir?: string,
+  { cwd = process.cwd(), templatesDir }: PostinstallOptions = {},
 ) {
-  await generateActionsCommand(appManifest, templatesDir);
-  await generateManifestCommand(appManifest);
-  await generateSchemaCommand(appManifest);
+  const projectRoot = await getProjectRootDirectory(cwd);
+  await generateActionsCommand(appManifest, { cwd: projectRoot, templatesDir });
```
oshmyheliuk: It's confusing that here `projectRoot` is passed as `cwd`, and after that it will be resolved in the action again as `const projectRoot = await getProjectRootDirectory(cwd);`. Looks like the action can be refactored.

packages/aio-commerce-lib-app/source/commands/generate/actions/main.ts
iivvaannxx: I ended up removing the type altogether, templates dir is only needed by generation commands, so reusing a type was kind of "forcing" it? Now parameters are non-objects. If we end up needing more params we can bring back objects as needed.

packages/aio-commerce-lib-app/source/commands/generate/schema/main.ts
iivvaannxx: Done in [90c674c](https://github.com/adobe/aio-commerce-sdk/pull/639/commits/90c674cf56b1f1e19f79e3acacc09762403a1820)

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
iivvaannxx: That function would have been a patch over a core issue (which you also correctly noticed in the comment below). We were making each command that needed to interact to the file system resolve repeatedly the root project dir all the time. 

The correct way to handle that is to resolve at the entry point of the process and make all intermediate helpers ask for a `projectRoot` instead of resolving the root from a `cwd`. The PR now includes this refactor.

packages/aio-commerce-lib-app/source/commands/hooks/postinstall.ts
iivvaannxx: See: https://github.com/adobe/aio-commerce-sdk/pull/639#discussion_r3813867806

packages/aio-commerce-lib-app/source/commands/generate/actions/main.ts
```diff
@@ -38,51 +39,57 @@ import type { CommerceAppConfigOutputModel } from "#config/schema/app";
 /**
  * Runs the generate actions command for the given extension point.
  * @param appManifest - The app manifest, used to determine which actions to generate and their content.
- * @param templatesDir - The directory to load templates from, for testing purposes. Defaults to the generated actions template root.
+ * @param options - Working directory and templates directory overrides.
```
oshmyheliuk: JSDoc doesn't reflect changes in this PR

oshmyheliuk (approved):

packages/aio-commerce-lib-app/source/commands/generate/actions/main.ts
iivvaannxx: Fixed in [d769e2a](https://github.com/adobe/aio-commerce-sdk/pull/639/commits/d769e2a8b1c0be89038b0e1375b6a172e1ac44ea)

PR #645, opened 2026-08-20
Create search criteria builder to make calling list endpoints easier

aligent-lturner:
## Description

Adds a search criteria builder to `@adobe/aio-commerce-lib-api` for the `searchCriteria[...]` query parameters that Adobe Commerce REST list endpoints expect.

The primitive is a plain `SearchCriteria` object plus a pure serializer, with a thin immutable fluent builder layered on top:

```typescript
// Fluent
const searchParams = searchCriteria()
  // status is 1 OR status is 2...
  .filterGroup({ field: "status", value: 1 }, { field: "status", value: 2 })
  // ...AND created on or after 2026-01-01
  .filter({ field: "created_at", value: "2026-01-01", conditionType: "gteq" })
  .sort("created_at", "DESC")
  .paginate({ pageSize: 50, currentPage: 1 })
  .toSearchParams();

// Declarative — suits conditionally-built queries
const searchParams = buildSearchCriteria({
  filterGroups: [[{ field: "sku", value: "24-MB01" }]],
  pageSize: 20,
});

await client.get("products", { searchParams });
```

New exports from the package root:

- `searchCriteria()` — fluent immutable builder; `filter()` AND-s, `filterGroup()` OR-s within the group
- `buildSearchCriteria()` — serializes to `URLSearchParams`
- `buildSearchCriteriaRecord()` — serializes to a plain record, for merging with unrelated query parameters
- Types: `SearchCriteria`, `SearchFilter`, `SearchFilterValue`, `ConditionType`, `SortOrder`, `SortDirection`, `Pagination`, `SearchCriteriaBuilder`

Nothing existing changes. The output goes into the `searchParams` option `ky` already accepts, so this is additive and orthogonal to `AdobeCommerceHttpClient` and `ApiClient`. The client deliberately does not grow a `searchCriteria` option; keeping the serializer decoupled means it works with a raw `ky` instance and is testable without constructing a client.

The builder is flavour-agnostic: SaaS exposes the same REST surface and the same `searchCriteria` convention, and the base path difference (`/V1/...` vs `rest/{storeViewCode}/V1/...`) is already handled by the client and never reaches the serializer.

### Condition types

`ConditionType` is a closed union of the 19 conditions that behave correctly over REST. Three that Commerce will accept are deliberately excluded, because each silently misbehaves rather than returning an error:

| Excluded   | Why                                                                                                                                                                                | Use instead                  |
| ---------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------- |
| `moreq`    | Listed in Adobe's condition type table, but absent from `Mysql::prepareSqlCondition`'s condition map. It falls through to the OR-branch, which discards the key, and becomes `eq`. | `gteq`                       |
| `fulltext` | An admin grid condition type applied over a FULLTEXT index by the UI component `FilterPool`. Repositories only reach a fallback that rewrites it to `like` with wildcards.         | `like` with explicit `%`     |
| `ntoa`     | Needs an IPv4 address stored as an integer. No core table does that — `sales_order.remote_ip` is a `varchar`, where `INET_NTOA` yields null and the filter matches nothing.        | `like`/`eq` on the raw value |

Conversely `regexp`, `seq` and `sneq` are supported by Commerce but absent from the documented table, so they are included. Each exclusion is documented inline in `types.ts` so the gaps are not "fixed" later by mistake.

The serializer also absorbs the per-condition rules that are otherwise undocumented and only discoverable by looking at the code: `conditionType` defaults to `eq`, `in`/`nin` array values are comma-joined, booleans map to `1`/`0` (how Commerce stores boolean attributes), empty filter groups are skipped so conditionally-built lists cannot emit a malformed query, and indices are derived from array position so they are always dense.

One subtlety worth flagging for review: values are forwarded even for `null`/`notnull`, whose SQL ignores them. On a non-static EAV attribute, `Eav\...\AbstractCollection::_getAttributeConditionSql` only switches the attribute join to `LEFT` when `isset($condition['null'])`, and an `INNER` join cannot match the missing rows a null check is looking for. 

## Related Issue

Closes #

## Motivation and Context

Every Commerce REST list endpoint (`products`, `orders`, `customers/search`, `invoices`, ...) is queried through the same positionally-indexed, bracket-nested `searchCriteria` convention. The SDK currently provides a well-configured HTTP client and then leaves consumers to assemble those parameters by hand:

```typescript
const searchParams = new URLSearchParams({
  "searchCriteria[filterGroups][0][filters][0][field]": "status",
  "searchCriteria[filterGroups][0][filters][0][value]": "1",
  "searchCriteria[filterGroups][0][filters][0][conditionType]": "eq",
  "searchCriteria[filterGroups][0][filters][1][field]": "status",
  // ...and so on
});
```

Verbosity is the least of the problems:

- **The group semantics are invisible.** Filters within one `filterGroups` entry are OR-ed; separate entries are AND-ed. Nothing in the hand-written form conveys that, and it is the most commonly misunderstood part of the convention. Someone wanting "status is 1 AND created after X" who writes both filters into group `0` gets "status is 1 OR created after X"; a query that returns results, just the wrong ones. `filter()` vs `filterGroup()` states the distinction at the call site.
- **Index bookkeeping is manual.** Indices must be dense and sequential; inserting a filter mid-set means renumbering every parameter after it, in string keys no type checker validates.
- **Per-condition rules are undiscoverable**, and as above, three condition types are actively unsafe over REST while looking fine.
- **It gets rewritten in every integration.** Because it is pure string assembly with no dependencies, every app builds its own partial version, and each gets a different subset of the rules right.

Adobe Commerce's PHP codebase solves this with `SearchCriteriaBuilder`/`FilterBuilder`/`FilterGroupBuilder`, so developers arriving from extension work expect an equivalent.

This lands in `aio-commerce-lib-api` because that package already owns Commerce REST wire-format concerns in `source/lib/commerce/` (path construction, flavors, store headers)..

Why the object is primary and the chain derived: PHP's builder is fluent because it mutates a DTO in a language without object literals. TypeScript has object literals, and a plain object is inspectable, serializable, storable in config, and composes with conditional logic easily. The fluent wrapper gives the "niceness" without duplicating serialization logic. `field` stays a bare `string` because Commerce attributes are per-installation, so any shipped union would be wrong for most stores while implying completeness.

## How Has This Been Tested?

35 new unit tests (Vitest) across three files; 110 pass in the package overall, with `pnpm typecheck` and Biome clean.

- `test/unit/lib/commerce/search-criteria/serialize.test.ts` — serialization: default `conditionType`, AND/OR group nesting, comma-joined `in`/`nin` arrays and scalars, boolean `1`/`0`, empty-group skipping and dense indices, sort orders, pagination (including `pageSize: 0`), percent-encoding, and the record variant.
- `test/unit/lib/commerce/search-criteria/builder.test.ts` — builder semantics: `filter()` vs `filterGroup()` grouping, sort defaulting to `ASC`, pagination replacement, accepting initial criteria, plus an equivalence test asserting the fluent and declarative forms serialize identically. Four immutability tests cover specialising a shared base, mutation attempts through a derived builder, through the initial criteria object, and through the result of `toJSON()`.
- `test/unit/lib/commerce/search-criteria/request.test.ts` — end-to-end through a mocked `AdobeCommerceHttpClient`, asserting the parameters actually reach the wire, and that both merge strategies work for endpoints taking search criteria alongside other query parameters (`fields` on any endpoint; `storeId`/`currencyCode` on `GET /V1/products-render-info`).

Two behaviours have dedicated regression tests because they are subtle enough to be "tidied" away:

- An empty-string value survives serialization — Commerce rewrites `seq`/`sneq` to a null check when the value is empty, so a falsy check here would break it.
- A value supplied with `null`/`notnull` is forwarded, for the EAV `LEFT` join reason above.

Verification of the Commerce-side behaviour was done against `magento/magento2@2.4-develop` source (`Mysql::prepareSqlCondition`, `FilterProcessor`, `Data/Collection/AbstractDb`, `Eav/.../AbstractCollection`, `Api/Filter`), including executing the `prepareSqlCondition` branch logic locally to confirm the `moreq` degradation and the `seq`/`sneq` key-reassignment path. Not tested against a live Commerce instance.

N/A — no user-facing UI.

## Types of changes

## Checklist:


packages/aio-commerce-lib-api/docs/usage.md
```diff
@@ -220,6 +221,172 @@ const updated = await commerceApiClient.updateProduct("test-sku", {
...
+  ]);
+}
+
+const searchParams = buildSearchCriteria({ filterGroups, pageSize: 100 });
+await commerceClient.get("products", { searchParams }).json();
+```
+
+#### Filter groups: AND vs OR
+
+This is the part of the Commerce convention that is easiest to get wrong. Filters **within** a group are OR-ed; **separate** groups are AND-ed. The builder makes the distinction explicit:
```
iivvaannxx: I wonder if the Boolean semantics could be made more explicit in the method names. What do you think about using something like `where()` for the common single-filter case, with `whereAll()` and `whereAny()`  for multiple filters?

```ts
  searchCriteria()
    .where(createdAfter)
    .whereAll(isEnabled, isVisible)
    .whereAny(statusIs1, statusIs2);
  // createdAfter AND (isEnabled AND isVisible) AND (statusIs1 OR statusIs2)
```
 
These are still only two semantic modes (`where()` is the singular form of `whereAll()`), whereas `whereAny()` creates a single ORed group. This seems more explicit than `filter()` / `filterGroup()`.

This is only a suggestion; you might argue that `filter` and `filterGroup` maps more directly to the semantics of Commerce, which I would consider fair. I just think that this allows you to make the "boolean" semantics more explicit (via the "all" and "any"), and the `AND` relation is kind of implicit among all the `where*` calls, as you're composing each on top of the other.

packages/aio-commerce-lib-api/docs/usage.md
```diff
@@ -220,6 +221,172 @@ const updated = await commerceApiClient.updateProduct("test-sku", {
...
+
+const page1 = recent
+  .paginate({ pageSize: 50, currentPage: 1 })
+  .toSearchParams();
+const page2 = recent
+  .paginate({ pageSize: 50, currentPage: 2 })
+  .toSearchParams();
+```
+
+`.toJSON()` returns the underlying `SearchCriteria`, so a fluent chain can be handed to code that expects the plain object, and `searchCriteria(initial)` accepts one to start from.
```
iivvaannxx: I get that `SearchCriteria` is Json serializable and therefore `toJson` can return it, but sounds a bit confusing to me that we use `toJson` for this? I read that method and I think more of a JSON string. But that might be only me.

iivvaannxx (approved):

packages/aio-commerce-lib-api/docs/usage.md
aligent-lturner: I'm not against it - I do like the idea of filter and filterGroup mapping more directly to the commerce parameters, but they're not all that clear themselves.
Happy to go with the consensus.

packages/aio-commerce-lib-api/docs/usage.md
aligent-lturner: Fair - maybe something like `toObject` would fit better?

aligent-lturner: I did forget one part of this, which was the `toJSON` has meaning for `JSON.stringify`, in that it will automatically be called. As such, if you called something like `JSON.stringify(builder)`, you'd get valid output instead of just `{}`. So I'd probably actually lean towards keeping it. Alternatively, I could keep that function for that purpose, but add another that's more obvious (e.g. `toObject`) for the purpose of getting a `SearchCriteria` object.

jcuerdo (approved):

packages/aio-commerce-lib-api/docs/usage.md
obarcelonap: let's skip Ivan's suggestion for the moment

packages/aio-commerce-lib-api/docs/usage.md
obarcelonap: I think we can keep `toJSON` and add `toObject` in the future if there's a clear use case.

obarcelonap:
@aligent-lturner Thanks for this contribution, I truly believe this is a great addition.

Thinking about this, I'd drop the fluent builder and only ship `buildSearchCriteria`/`buildSearchCriteriaRecord` over the plain `SearchCriteria` object?

A few reasons:
- One way to construct search criteria is easier to learn and discover than two.
- Less code to maintain, no `builder.ts`, no `copy()`, no immutability guarantees to test.
- It removes the `toJSON`/`toObject` naming question.
- Every other config type in this SDK (`CommerceHttpClientConfig`, `SaaSClientParams`, `ResolveCommerceHttpClientParamsOptions`, etc.) is a plain object, not a builder, so dropping it keeps this consistent with how the rest of the SDK is shaped.
- TypeScript autocomplete on the object literal already surfaces `filterGroups`, `sortOrders`, `pageSize`, `currentPage` in one place, so discoverability isn't actually better with a second, chained type to learn.
- The conditionally-built case (pushing filter groups based on optional params) already reads better with the object form, so I don't think we lose much for that use case either.

Happy to be argued out of this, but I think minimizing the public surface while keeping the same features and reducing the complexity is a clear win.

aligent-lturner:
@obarcelonap I think your points all make sense - the consistency one especially.  Also a couple of additional points in favour after looking at it again:
- Semver: Adding `searchCriteria()` later would be a minor change, while removing would be major.
- The chain doesn't buy anything at the type level. `.filter()` just hands back `SearchCriteriaBuilder`, unlike something like `HttpActionRouter.use()`, which returns `HttpActionRouter<TContext & TNew>` (and earns its place)

One small real loss - `.filter()` vs `.filterGroup()` named the AND/OR distinction in code, and `SearchFilter[][]` doesn't. But that's a docs problem and `usage.md` already has the table plus the annotated example, so I'll move those onto the object form.

I'll make the changes shortly

obarcelonap (approved):

PR #648, opened 2026-08-24
CEXT-6661: fix uninstallation failing on dynamicList business config fields

jcuerdo:
## Description

Uninstallation revalidates the recorded installation snapshot against the strict `businessConfig` schema, which requires `dynamicList` `options`/`default` to be functions. Those functions never survive the `JSON.stringify` used to persist the snapshot (State/Files stores), so uninstallation always failed for apps with a `dynamicList` field.

Adds a `Recorded*` schema variant (in `aio-commerce-lib-config` and `aio-commerce-lib-app`) that accepts a `dynamicList` field without `options`/`default`, since nothing in the uninstall workflow reads them, and uses it only for the recorded-config validation in `uninstall.ts`. Install, upgrade, and the domains that do need strict validation during uninstall (webhooks, events, adminUi, custom installation) are unaffected.

## Related Issue

Closes [CEXT-6661](https://jira.corp.adobe.com/browse/CEXT-6661)

## Motivation and Context

A partner developing an Algolia integration hit this on every uninstall of an app using a `dynamicList` Business Configuration field.

## How Has This Been Tested?

- New unit tests for the `Recorded*` schemas in `aio-commerce-lib-config` (accepts `dynamicList` without functions, still rejects real corruption).
- New unit tests for `validateRecordedCommerceAppConfig` in `aio-commerce-lib-app`.
- Fixed a test-double gap: the installation store mock passed values by reference instead of round-tripping through JSON like the real stores, which is why this bug was invisible to the existing suite. Added a serializing variant and an end-to-end regression test (install a `dynamicList` field, then uninstall) that fails without the fix.
- Reproduced manually end-to-end against a real deployed app before and after the fix.

## Types of changes

## Checklist:

### DIFF
--- a/packages/aio-commerce-lib-app/source/config/schema/app.ts
+++ b/packages/aio-commerce-lib-app/source/config/schema/app.ts
@@ -57,11 +57,6 @@ export const RecordedCommerceAppConfigSchema = v.looseObject({
   businessConfig: v.optional(RecordedSchemaBusinessConfig),
 });
 
-/** The output shape of the recorded commerce app config schema. */
-export type RecordedCommerceAppConfigOutputModel = v.InferOutput<
-  typeof RecordedCommerceAppConfigSchema
->;
-
 /** @internal Any commerce app config, input contract or validated output. */
 export type AnyCommerceAppConfig =
   | CommerceAppConfig
--- a/packages/aio-commerce-lib-app/source/config/schema/app.ts
+++ b/packages/aio-commerce-lib-app/source/config/schema/app.ts
@@ -15,7 +15,10 @@
 import * as v from "valibot";
 
 import { AdminUiSchema } from "./admin-ui";
-import { SchemaBusinessConfig } from "./business-configuration";
+import {
+  RecordedSchemaBusinessConfig,
+  SchemaBusinessConfig,
+} from "./business-configuration";
 import { getConfigDomains } from "./domains";
 import { EventingSchema } from "./eventing";
 import { InstallationSchema } from "./installation";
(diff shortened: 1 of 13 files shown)


packages/aio-commerce-lib-app/source/config/schema/app.ts
```diff
@@ -43,6 +46,22 @@ export type CommerceAppConfigOutputModel = v.InferOutput<
...
+ * `businessConfig.schema` `dynamicList` fields may have lost their
+ * `options`/`default` functions to a JSON round trip through storage.
+ */
+export const RecordedCommerceAppConfigSchema = v.looseObject({
+  ...CommerceAppConfigSchema.entries,
+  businessConfig: v.optional(RecordedSchemaBusinessConfig),
+});
+
+/** The output shape of the recorded commerce app config schema. */
+export type RecordedCommerceAppConfigOutputModel = v.InferOutput<
```
oshmyheliuk: This type is exported but is never used

packages/aio-commerce-lib-config/source/modules/schema/fields.ts
```diff
@@ -265,3 +265,74 @@ export const SchemaBusinessConfigSchema = v.pipe(
...
+const RecordedMultipleDynamicListSchema = v.object({
+  ...RecordedDynamicListEntriesCommon,
+  default: v.optional(
+    v.custom<MultipleDefaultFactory>(
+      (input) => typeof input === "function",
+      'Expected a function for "default"',
+    ),
+  ),
+  selectionMode: v.literal(
+    "multiple",
```
oshmyheliuk: Schemas added in this PR `Recorded*` are very similar to the `Multiple*` schemas. Probably can be refactored by using  `v.partial`

oshmyheliuk (approved):

packages/aio-commerce-lib-app/source/config/schema/app.ts
jcuerdo: Removed in a2a33fea.

packages/aio-commerce-lib-config/source/modules/schema/fields.ts
jcuerdo: Fixed in a2a33fea.

PR #653, opened 2026-08-28
CEXT-6696: Implement custom ACL metadata element

oshmyheliuk:
## Description

Adds support for declaring **custom, standalone ACL resources** under a new `adminUi.acl` config section, a helper to check them at runtime, and matching Admin UI skill documentation.

- New `adminUi.acl` schema in `@adobe/aio-commerce-lib-app`: a list of named resources, each a flat leaf or a one-level group of leaf children (`{ id, label, children? }`). Grouping is exactly one level deep — grandchildren are structurally rejected, ids must be unique per level, and a group must have at least one child. There is no `description` field: Commerce's User Roles tree renders only a node's title, so a description could never be shown.
- `hasAdminUi` now counts `acl`, and the Admin UI upgrade planner (`enumerateComponents`, `AdminUiComponentKind`/`AdminUiComponentEntity`) enumerates each top-level entry as an `acl` component so install/upgrade diffing picks the section up.
- New `getCustomAclResourceId(appId, resourceId, childId?)` in `@adobe/aio-commerce-lib-admin-ui`, reusing the internal id prefix/sanitization so it mirrors the Commerce `AclResourceIdGenerator` exactly; exported from the `api` entrypoint and checkable with the existing `getAdminUiPermissionClient`.
- The section reaches Commerce unchanged: the `app-config` action already serves the validated config verbatim, so adding `acl` to `AdminUiSchema` (which strips unknown keys) is sufficient — no router or registration changes.
- Updated the `commerce-app-admin-ui` skill: added custom ACL resources to `SKILL.md` (description, intent, extension-point table, config example) and a new `references/custom-acl.md`, keeping plugin content in sync with the SDK.
- Docs added to both packages' `usage.md`; cross-repo parity fixtures added; `minor` changesets for `@adobe/aio-commerce-lib-app`, `@adobe/aio-commerce-lib-admin-ui`, and `@adobe/aio-commerce-plugin-app-management`.

## Related Issue

[CEXT-6696](https://jira.corp.adobe.com/browse/CEXT-6696).

https://github.com/magento-commerce/adobe-commerce-backend-uix/pull/384

https://github.com/adobe/aio-commerce-sdk/pull/660

## Motivation and Context

Today an app can only obtain an ACL resource by marking an existing Admin UI element with `aclProtected: true`, which Commerce then enforces server-side (hides the element, blocks its URL) — there is no way to declare a permission that isn't bound to a UI element, one Commerce can't see and therefore can't enforce.

Apps need exactly that to gate their own logic: hide a button, tab, or panel rendered inside the app's iframe, guard a branch inside a runtime action, or model coarse per-role feature flags. This change lets an app declare such permissions; Commerce renders them in the User Roles tree so a merchant can grant or deny them, and the app checks them itself (app-checked, not server-enforced). It extends the CEXT-6150 ACL mechanism and reuses its deterministic id scheme and tree-injection machinery.

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/source/config/schema/admin-ui.ts
```diff
@@ -322,7 +388,8 @@ export function hasAdminUi<T extends AnyCommerceAppConfig>(
   }
 
   return Boolean(
-    adminUi.menu ||
+    adminUi.acl?.length ||
```
jcuerdo: With this change, buildAdminUiV2ExtConfig doesn't account for adminUi.acl. Since hasAdminUi now treats acl-only configs as present, this generates operations: `{}`, an empty/invalid extension registration for commerce/backend-ui/2. Is this expected?

packages/aio-commerce-lib-app/source/config/schema/admin-ui.ts
```diff
@@ -224,13 +224,72 @@ const MenuSchema = v.object({
...
+  ),
+  id: AclResourceIdSchema,
+  label: AclResourceLabelSchema,
+});
+
+const AdminUiAclSchema = v.pipe(
+  v.array(AclResourceEntrySchema),
+  v.minLength(1, "At least one ACL resource is required when acl is defined"),
+  v.check(
+    (items) => new Set(items.map((i) => i.id)).size === items.length,
```
jcuerdo, line 319: @oshmyheliuk maybe it worth to check this bug found by claude:

Pre-existing bug, not introduced by this PR: uniqueness is checked on the raw `id`, but Commerce resource ids are sanitized (lowercased, special chars  normalized via `sanitizeSegment`) when actually generated (see `getCustomAclResourceId`). `"Reports"` and `"reports"` both pass this check but collide into the same Commerce ACL resource, silently overwriting one.                                                                                                                       This same gap already exists for `massActions`/`viewButtons` ids above (also sanitized downstream via their own `acl-resource-id.ts` helpers), so this PR isn't introducing the issue, just copying an existing pattern into a new field. Flagging in case it's worth fixing across all three now, or tracking as a  
  separate ticket.

packages/aio-commerce-lib-app/source/config/schema/admin-ui.ts
oshmyheliuk: Yes, it makes sense not to generate operations and other stuff only if custom ACL are added. The logic was updated.

packages/aio-commerce-lib-app/source/config/schema/admin-ui.ts
oshmyheliuk: Fixed for ACL ids format validation in this commit https://github.com/adobe/aio-commerce-sdk/pull/653/changes/bd1cfc5447aea932d8811b9487f9373c447bc388
For the rest I will create a separate JIRA ticket

oshmyheliuk: https://jira.corp.adobe.com/browse/CEXT-6707

oshmyheliuk: Fix here https://github.com/adobe/aio-commerce-sdk/pull/660

jcuerdo (approved):

PR #661, opened 2026-09-02
[DRAFT] CEXT-6701: Bind Commerce App Management Service record ownership during app association

oshmyheliuk:
## Types of changes

## Checklist:


packages/aio-commerce-lib-app/source/actions/association/router.ts
```diff
@@ -40,16 +49,53 @@ export const router = new HttpActionRouter<AssociationActionContext>().use(
...
 
-  handler: async (req, { logger }) => {
-    const { commerceBaseUrl, commerceEnv } = req.body;
+  handler: async (req, { logger, rawParams }) => {
+    const { commerceBaseUrl, commerceEnv, commerceId, workspaceId, extId } =
+      req.body;
+
+    try {
+      const camsClient = createCamsClient({
+        authProvider: getImsAuthProvider(resolveImsAuthParams(rawParams)),
```
iivvaannxx, line 74: Are CAMS operations done in behalf of the user or the app? This is using app credentials. Might already be correct, just wondering which is the case here.

packages/aio-commerce-lib-app/source/management/cams/client.ts
```diff
@@ -0,0 +1,247 @@
...
+  /** Adobe I/O Developer Console workspace id (the record's natural-key part). */
+  workspaceId: string;
+
+  /** App Builder application id — asserted against the stored record. */
+  extId: string;
+};
+
+/** A status entry appended to the record's status history. */
+export type CamsStatusUpdate = {
+  status: string;
```
iivvaannxx, line 38: Nitpick: Are these the statuses: associated, partially installed, installed...? If so we could make a string union instead of a generic string

packages/aio-commerce-lib-app/source/management/cams/client.ts
```diff
@@ -0,0 +1,247 @@
...
+  /** Commerce App Management Service base URL. */
+  baseUrl: string;
+
+  /** The app's own S2S IMS auth provider — its `client_id` becomes the owner. */
+  authProvider: ImsAuthProvider;
+
+  /** Identifiers for the record this client operates on. */
+  identity: CamsExtensionIdentity;
+
+  logger: ReturnType<typeof AioLogger>;
```
iivvaannxx, line 54: `logger` is the only member without JSDoc. Probably obvious, but all the others have it, so it might be good to add it for consistency.

packages/aio-commerce-lib-app/source/management/cams/client.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * Backoff schedule (ms) for retrying a transient adopt failure; its length is
+   * the retry count. Defaults to {@link DEFAULT_ADOPT_RETRY_DELAYS_MS}.
+   */
+  retryDelaysMs?: readonly number[];
```
iivvaannxx, line 60: Probably I would do this in a more `native` way to `ky`'s? It already has an `Options` with a lot of configuration, including retry count and delay strategy. Probably you can do similar to what we do with `AdobeCommerceHttpClient` and `IoEventsHttpClient` and separate `fetchOptions` from the `config`, for consistency.

packages/aio-commerce-lib-app/source/management/cams/client.ts
```diff
@@ -0,0 +1,247 @@
...
+   * @throws {CamsUnavailableError} The service was unreachable or errored.
+   */
+  ensureAdopted: () => Promise<string>;
+
+  /** Appends a status entry (`POST /v1/extensions/{id}/status`); adopts first. */
+  postStatus: (update: CamsStatusUpdate) => Promise<void>;
+
+  /** Patches the stored app config (`PATCH /v1/extensions/{id}`); adopts first. */
+  patchConfig: (appConfig: unknown) => Promise<void>;
+};
```
iivvaannxx, line 87: I would suggest following the `HttpClient` vs `ApiClient` distinction we do for events and I/O and other APIs in the SDK. Also `appConfig` is typed as `unknown` for a specific reason? We do have a type for that don't we?

packages/aio-commerce-lib-app/source/management/cams/client.ts
```diff
@@ -0,0 +1,247 @@
...
+
+/** Default backoff schedule (ms) for retrying a transient adopt failure. */
+export const DEFAULT_ADOPT_RETRY_DELAYS_MS = [250, 500, 1000] as const;
+
+const HTTP_NOT_FOUND = 404;
+const HTTP_CONFLICT = 409;
+const HTTP_SERVER_ERROR_MIN = 500;
```
iivvaannxx, line 94: Retrieve those from `lib-core` those constants are also exported from there and the dependency is already in place.

packages/aio-commerce-lib-app/source/management/cams/client.ts
```diff
@@ -0,0 +1,247 @@
...
+/** Reads a JSON body defensively, returning `undefined` on any failure. */
+async function readJsonBody(
+  response: Response,
+): Promise<Record<string, unknown> | undefined> {
+  try {
+    return (await response.clone().json()) as Record<string, unknown>;
+  } catch {
+    return undefined;
+  }
+}
```
iivvaannxx, line 108: This function is only ever used once, and for how is it used I would say it's better to just wrap the callee into this same try/catch and handle inline.

packages/aio-commerce-lib-app/source/management/cams/client.ts
```diff
@@ -0,0 +1,247 @@
...
+    `The Commerce App Management Service returned HTTP ${status}: ${detail}`,
+    { cause: error, retryable: status >= HTTP_SERVER_ERROR_MIN },
+  );
+}
+
+/**
+ * Creates a {@link CamsClient} for a single record, backed by `ky` with the app's
+ * S2S auth attached on every request.
+ */
+export function createCamsClient(options: CamsClientOptions): CamsClient {
```
iivvaannxx, line 151: I think such a function is better modeled as a `class` similar to how we do it with other clients in the other packages.

packages/aio-commerce-lib-app/source/management/cams/client.ts
```diff
@@ -0,0 +1,247 @@
...
+          await sleep(retryDelaysMs[attempt]);
+          continue;
+        }
+
+        throw error;
+      }
+    }
+
+    throw lastError;
+  }
```
iivvaannxx, line 221: This retry pattern is already built into `ky` in a more native way, I don't think we need to build something like this again and instead leverage the library better.

packages/aio-commerce-lib-app/source/management/cams/client.ts
```diff
@@ -0,0 +1,247 @@
...
+    // (409) responses; ky's built-in retry cannot inspect the problem+json body.
+    retry: 0,
+  });
+
+  async function adoptOnce(): Promise<string> {
+    try {
+      const response = await http.post("v1/extensions:adopt", {
+        json: identity,
+      });
+      const body = (await response.json()) as { id: string };
```
iivvaannxx, line 184: Nitpick: `.json<T>()` is generic over `T`, you don't need a type assertion. Additionally, this assumes the output is `id: string` without any validation. This might lead to a successful association handling, even if somehow the response payload is malformed.

packages/aio-commerce-lib-app/source/management/cams/config.ts
```diff
@@ -0,0 +1,60 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+/**
+ * Action input holding an explicit Commerce App Management Service base URL.
+ * When set it overrides {@link DEFAULT_CAMS_BASE_URL} — internal/stage
+ * deployments point this at the stage host; customers leave it unset and get the
+ * default (production) endpoint.
+ */
+export const CAMS_BASE_URL_INPUT = "AIO_COMMERCE_APP_MANAGEMENT_SERVICE_URL";
```
iivvaannxx, line 19: Nitpick: To make this easier to type maybe we could name it `AIO_COMMERCE_APP_MGMT_SERVICE_URL`. 

Just a suggestion, feel free to ignore if you don't see any value in it.

packages/aio-commerce-lib-app/source/management/cams/config.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @param params - The runtime action inputs.
+ */
+export function resolveCamsBaseUrl(params: Record<string, unknown>): string {
+  const override = params[CAMS_BASE_URL_INPUT];
+  const baseUrl =
+    typeof override === "string" && override.trim() !== ""
+      ? override.trim()
+      : DEFAULT_CAMS_BASE_URL;
+  return ensureScheme(baseUrl);
+}
```
iivvaannxx, line 60: Probably you can drop most of this code in favor of the native `URL` object in Node, should do most of the validation for you.

packages/aio-commerce-lib-app/source/commands/generate/actions/constants.ts
```diff
@@ -29,10 +29,29 @@ export const COMMERCE_VARIABLES = [
...
-export const COMMERCE_ACTION_INPUTS = Object.fromEntries(
-  COMMERCE_VARIABLES.map((variable) => [variable, `$${variable}`] as const),
-);
+export const COMMERCE_ACTION_INPUTS = {
+  ...Object.fromEntries(
+    COMMERCE_VARIABLES.map((variable) => [variable, `$${variable}`] as const),
+  ),
+  [CAMS_BASE_URL_VARIABLE]: `$${CAMS_BASE_URL_VARIABLE}`,
+  [IMS_ENVIRONMENT_VARIABLE]: `$${IMS_ENVIRONMENT_VARIABLE}`,
+};
```
iivvaannxx, line 54: Why not add it to the `COMMERCE_VARIABLES` array?

packages/aio-commerce-lib-app/source/management/cams/errors.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Thrown when the Commerce App Management Service is unreachable or responded
+ * with an unexpected status (network failure, 5xx, or an unhandled 4xx).
+ *
+ * `retryable` reflects whether the underlying condition was transient (network
+ * or 5xx) — defaults to `true`.
+ */
+export class CamsUnavailableError extends CamsError {
+  public override readonly retryable: boolean;
```
iivvaannxx, line 74: I like this `retryable` thing, maybe we could bake it into the base `CommerceSdkError`. Not necessarily in this PR if it's a big change.

packages/aio-commerce-lib-app/source/commands/generate/actions/config.ts
```diff
@@ -117,7 +117,16 @@ export function buildAppManagementExtConfig(
...
             "app-config": createActionDefinition("app-config"),
-            association: createActionDefinition("association"),
+            // The association action adopts the app's Commerce App Management
+            // Service record with the app's own S2S credentials, so it needs the
+            // Commerce auth inputs (and the optional service URL override).
+            association: createActionDefinition(
+              "association",
+              {},
+              {
+                inputs: { ...COMMERCE_ACTION_INPUTS, LOG_LEVEL: "$LOG_LEVEL" },
```
iivvaannxx, line 127: You can pre-create this `inputs` in a variable and reuse with the `installation` action.

packages/aio-commerce-lib-app/source/actions/association/schema.ts
```diff
@@ -22,4 +25,10 @@ export const AssociationRequestBodySchema = v.object({
     ),
   ),
   commerceEnv: CommerceEnvSchema,
+
+  // Identifiers for the `:adopt` handshake with the Commerce App Management
+  // Service. Supplied by the Commerce App Management frontend on association.
+  commerceId: nonEmptyStringValueSchema("commerceId"),
```
iivvaannxx, line 31: Some questions:

- The `commerceId` is only the url and the env combined and digested via `base64` is it really needed?
- What is the `extId`? Is the id assigned by Extension Manager? Why does the frontend send this instead of the service?

Adding these parameters non-optionally makes this a breaking change. Is there a way to make this work with optional parameters?

packages/aio-commerce-lib-app/source/actions/association/router.ts
```diff
@@ -40,16 +49,53 @@ export const router = new HttpActionRouter<AssociationActionContext>().use(
...
+    } catch (error) {
+      if (error instanceof CamsAdoptConflictError) {
+        logger.error(`Adoption rejected: ${error.message}`);
+        return conflict(error.message);
+      }
+
+      logger.warn(
+        `Adoption deferred; ownership will bind on the next write: ${
+          error instanceof Error ? error.message : String(error)
+        }`,
```
iivvaannxx, line 95: What does "next write" mean? You mean this kind of ownership is attempted on every operation?

packages/aio-commerce-lib-app/source/management/cams/config.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export const CAMS_BASE_URL_INPUT = "AIO_COMMERCE_APP_MANAGEMENT_SERVICE_URL";
+
+// TODO: replace with the public production URL once the Commerce App Management
+// Service is provisioned in production. Until then this points at the stage host
+// (mirrors the same placeholder in the Commerce App Management frontend's
+// `useCommerceAppManagementApi`).
+/** Default Commerce App Management Service base URL used when no override is set. */
+export const DEFAULT_CAMS_BASE_URL =
+  "https://commerce-app-management-dev.adobe.io";
```
iivvaannxx, line 27: I think you missed the TODO of replacing this with the production endpoint

iivvaannxx (commented):
Heads up, the API is changed but the `openapi.json` has not been updated.

packages/aio-commerce-lib-app/source/management/cams/client.ts
```diff
@@ -0,0 +1,247 @@
...
+    },
+    prefixUrl: baseUrl,
+    // Own the retry loop so we can distinguish retryable (404/5xx) from terminal
+    // (409) responses; ky's built-in retry cannot inspect the problem+json body.
+    retry: 0,
+  });
+
+  async function adoptOnce(): Promise<string> {
+    try {
+      const response = await http.post("v1/extensions:adopt", {
```
iivvaannxx, line 181: When I read this `extensions:adopt` thing I was confused since I never saw this kind of notation in API Design. Then I learned it's [a pattern by Google](https://google.aip.dev/136), so it was nice getting to learn that. 

However, for me, it still reads odd. So I wonder if some alternatives have been considered? I was wondering if something like this would fit better in the typical REST-like resource notation?

- Extension `PATCH` under a different identity: `/v1/extensions/commerceId/workspaceId`
- Extension `PATCH` under the existing `/v1/extensions/extId` but we support different identities via a tagged locator with an ID that could be generated from commerceId + workspaceId via base64 (or the normal and current extId as another tagged locator)

obarcelonap: If I'm not mistaken, the adopt endpoint exists to guarantee that only the app itself can update the backend's metadata. That's why we pass the `clientId`, so we can validate the origin. Otherwise any app could update another app's data.

I was wondering if we could move this to the association step and save ourselves this endpoint. In other words, when an `/associate` is done on the app, the app returns the `clientId`, that way we'd always know which `clientId` is allowed to update the app.

This would also reduce the number of requests, which simplifies the architecture and cuts down on moving parts. Keep in mind that with the new design, requests from the app to the backend should be minimized, since the backend owns the orchestration.

We'd need to think about the migration scenario: apps that are already installed without a `clientId`. Maybe forcing an unassociation/association is good enough, or we might need to find a better solution.

packages/aio-commerce-lib-app/source/actions/association/router.ts
oshmyheliuk: CAMS operations can be done based on a user token or server-to-server credentials from the app.
In the association from the app, we need App credentials to bind ownership by client ID.

packages/aio-commerce-lib-app/source/management/cams/client.ts
oshmyheliuk: Added union in this commit https://github.com/adobe/aio-commerce-sdk/pull/661/changes/05d9c046ded666b68d933c7c5b69b0026e73f24e

packages/aio-commerce-lib-app/source/management/cams/client.ts
oshmyheliuk: Added https://github.com/adobe/aio-commerce-sdk/pull/661/changes/05d9c046ded666b68d933c7c5b69b0026e73f24e

packages/aio-commerce-lib-app/source/management/cams/client.ts
oshmyheliuk: Refactored here https://github.com/adobe/aio-commerce-sdk/pull/661/changes/119f014548df8be0b90b4808182e3986dd58b180

packages/aio-commerce-lib-app/source/management/cams/client.ts
oshmyheliuk: Done here https://github.com/adobe/aio-commerce-sdk/pull/661/changes/0ce1c9d4c35d7b7c3372d993b2cc2fee40d0efb3

packages/aio-commerce-lib-app/source/management/cams/client.ts
oshmyheliuk: Refactored https://github.com/adobe/aio-commerce-sdk/pull/661/changes/0ce1c9d4c35d7b7c3372d993b2cc2fee40d0efb3

packages/aio-commerce-lib-app/source/management/cams/client.ts
oshmyheliuk: Was refactored here https://github.com/adobe/aio-commerce-sdk/commit/119f014548df8be0b90b4808182e3986dd58b180

packages/aio-commerce-lib-app/source/management/cams/config.ts
oshmyheliuk: Looks like `URL` logic is little bit different:

```
Why new URL() doesn't do this:
- new URL("cams.example.com") throws — it does not prepend https://. So URL alone would regress the scheme-less-override case from "default to https" to "crash." The whole reason ensureScheme exists is exactly the bit URL won't do.
- new URL(x).toString() also normalizes — it appends a trailing slash (https://cams.stage.example.com → …example.com/), which would break the exact-string assertions in 3+ tests and change what's handed to ky as prefixUrl.
```

packages/aio-commerce-lib-app/source/actions/association/schema.ts
oshmyheliuk: `commerceId` isn't just the URL and env combined and Base64-encoded, it also includes sanitization logic in PHP. 
We can replicate the logic in CAMS, but it should be the same logic, and it can break if the PHP part changes.

`extId` can be a little bit confusing. This field was migrated from the Extension Manager schema, and it's actually [the App Builder project ID ](https://github.com/OneAdobe/commerce-app-management/blob/main/src/hooks/application/association/useAssociateApp.ts#L139)(`extId: project.appId`). Maybe we need to rename it, I don't know why it was named in such a way.

Yes, I will change them to optional. By design, the adoption is done as `best-effort`

oshmyheliuk: Update here https://github.com/adobe/aio-commerce-sdk/pull/661/changes/2c4f799ec3582363a6ed08473b7028aef2893d6f

packages/aio-commerce-lib-app/source/actions/association/router.ts
oshmyheliuk: Not on every operation, only once per client instance creation, as it's memoized. With the move to the "orchestrator" logic, the App calls the CAMS only when an upgrade is available. So it runs once per upgrade, which isn't a frequent operation.

packages/aio-commerce-lib-app/source/management/cams/config.ts
oshmyheliuk: It was replaced in the stacked PR. I will update here as well.

packages/aio-commerce-lib-app/source/management/cams/client.ts
oshmyheliuk: And removed later, this was for pre-orchestration implementation when the App sends status updates, now the orchestrator is polling, so it's not needed anymore

packages/aio-commerce-lib-app/source/management/cams/errors.ts
oshmyheliuk: Let's do it in a separate PR.

packages/aio-commerce-lib-app/source/management/cams/config.ts
oshmyheliuk: Updated

packages/aio-commerce-lib-app/source/management/cams/client.ts
oshmyheliuk: >I was wondering if we could move this to the association step and save ourselves this endpoint. In other words, when an /associate is done on the app, the app returns the clientId, that way we'd always know which clientId is allowed to update the app.

Currently, Commerce App Management still calls the association directly with the app, not the service, so returning the client ID won't solve the issue. Commerce App Management would then need to make another call to the backend to update the client ID.

If we move the association logic to the backend, it will work, but it hasn't been done yet. We can create a separate task for it.

> This would also reduce the number of requests, which simplifies the architecture and cuts down on moving parts. Keep in mind that with the new design, requests from the app to the backend should be minimized, since the backend owns the orchestration.

Yes, after moving to orchestrator logic, the app calls the backend only for application upgrade and adoption, we can remove the adoption call as well when we move to service-driven adoption.

@iivvaannxx, regarding endpoints, I don't see a problem with endpoints like `/v1/extensions:adopt`, it's much more obvious than `PATCH /v1/extensions/commerceId/workspaceId`. But anyway, we probably will get rid of that endpoint if the association logic moves fully to the service.

packages/aio-commerce-lib-app/source/commands/generate/actions/constants.ts
oshmyheliuk: Added

packages/aio-commerce-lib-app/source/commands/generate/actions/config.ts
oshmyheliuk: Done here https://github.com/adobe/aio-commerce-sdk/pull/661/changes/6e91f5c48683f3b263004dc29fce3d17b10e78a9

packages/aio-commerce-lib-app/source/management/cams/client.ts
oshmyheliuk: I created a separate ticket for it https://jira.corp.adobe.com/browse/CEXT-6816

oshmyheliuk: But the SDK part should probably be part of this PR. I will update it

iivvaannxx: If it has this direct benefit of saving us the endpoint, I would do it as soon as we can, not necessarily in this PR, but before going into production and releasing, since after that, we will need to maintain it. 

>  don't see a problem with endpoints like /v1/extensions:adopt, it's much more obvious than PATCH /v1/extensions/commerceId/workspaceId.

I don't have much of a problem either, but I usually kind of dislike verbs in the endpoints, that's why I proposed it. In the end, this is patching an extension in the server; the extension is the resource. For me, the URL doesn't need to describe an action since makes the API "closed" to that specific action, while a `PATCH` can be reused for any kind of edit which is what we're doing in CAMS if I am not mistaken.

oshmyheliuk:
@obarcelonap @iivvaannxx This PR was opened before we discussed moving to the orchestrator logic. So I agree that now it makes more sense to do the adoption through the backend service, which simplifies the logic.
I'm closing this PR and opening a new one https://github.com/adobe/aio-commerce-sdk/pull/702

PR #665, opened 2026-09-04
CEXT-6659 Handle target-system rejections safely during upgrade replacements

vinayrao2000:
## Description

During an upgrade, an identity change on a webhook or event subscription is applied as remove-then-add. If the target rejects the replacement, the previous entity was already gone — leaving the app broken and the stored baseline out of sync with reality.

This adds a `RecoveryScope` abstraction (`management/common/workflow/recovery.ts`) that both the webhooks and events domains use: each domain records a compensation immediately after committing a change to the target, and on failure replays those compensations in reverse to roll the target back to the stored baseline before surfacing the original error. If recovery itself also fails, the failure carries both the original and recovery errors and flags that the target may no longer match the baseline (`UPGRADE_RECOVERY_FAILED`).

Extended past the ticket's literal replacement scope to also cover **multi-add atomicity** for events: adding several new event subscriptions in one upgrade is no longer partial — if the target rejects one, the ones already created in that attempt are rolled back too (webhooks already had this by virtue of its single-loop structure; events didn't).

## Related Issue

[CEXT-6659](https://jira.corp.adobe.com/browse/CEXT-6659)

## Motivation and Context

Some values pass SDK schema validation but are rejected by the target system (e.g. an invented `webhook_method`, or an event name Commerce doesn't support) — the schema can't enumerate every valid Commerce operation name. Without recovery, a rejected replacement left the app in a broken, inconsistent state relative to its own stored baseline.

## How Has This Been Tested?

- Unit tests: rejection → restore, recovery-also-fails (both errors recorded + baseline-divergence flag), 404/400 tolerance during rollback, multi-add rollback (including a wholly-new-provider case), in-place-update-failure no longer destroys a healthy subscription.
- Integration tests: real plan/apply against `msw`-mocked Commerce + I/O Events, for both webhooks (identity-change rejection → restore) and events (multi-add rejection → rollback).
- Live E2E against a real Commerce/Adobe I/O Events tenant: webhook identity-change rejection (restore + original error), multi-webhook all-or-nothing, config-only update path, and happy-path replacement — all confirmed. Also caught and fixed a real bug during live testing (rollback attempting to unsubscribe the rejected event's own never-created subscription; Commerce returns 400, not 404, for that case).
- Full package suite: 1288/1288 passing. Typecheck, biome, build, and `publint` all clean.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/.changeset/yummy-places-teach.md
+++ b/.changeset/yummy-places-teach.md
@@ -1,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": patch
+---
+
+Recover safely when Commerce rejects a webhook or event-subscription replacement or addition during an upgrade. An identity change is applied as a remove-then-add; if the target accepts the removal but rejects the replacement, the previous entity is now restored and the upgrade fails reporting the original rejection, instead of leaving the target out of sync with the stored baseline. Adding several new event subscriptions in one upgrade is no longer partial either: if the target rejects one, the ones already created in that attempt are rolled back too. When recovery itself cannot restore the baseline, the failure records both errors and flags that the target may no longer match the baseline. Fresh installs keep their existing best-effort behavior.
--- a/packages/aio-commerce-lib-app/source/management/common/workflow/recovery.ts
+++ b/packages/aio-commerce-lib-app/source/management/common/workflow/recovery.ts
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { unwrapHttpError } from "@adobe/aio-commerce-lib-api/utils";
+
+/** Failure key set when an upgrade replacement fails and recovery cannot fully restore the baseline. */
+export const UPGRADE_RECOVERY_FAILED = "UPGRADE_RECOVERY_FAILED";
+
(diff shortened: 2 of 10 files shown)


.changeset/yummy-places-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": patch
+---
+
+Recover safely when Commerce rejects a webhook or event-subscription replacement or addition during an upgrade. An identity change is applied as a remove-then-add; if the target accepts the removal but rejects the replacement, the previous entity is now restored and the upgrade fails reporting the original rejection, instead of leaving the target out of sync with the stored baseline. Adding several new event subscriptions in one upgrade is no longer partial either: if the target rejects one, the ones already created in that attempt are rolled back too. When recovery itself cannot restore the baseline, the failure records both errors and flags that the target may no longer match the baseline. Fresh installs keep their existing best-effort behavior.
```
oshmyheliuk, line 5: You can simplify the changeset description
```
Changeset messages must be user-facing and concise — describe what the developer can now do or what changed for them, not what the implementation does internally
```

packages/aio-commerce-lib-app/source/management/common/workflow/recovery.ts
```diff
@@ -0,0 +1,125 @@
...
+ * reverse (last-applied first), best-effort, to roll the target back to baseline when a later
+ * change fails:
+ *
+ * - if every compensation succeeds, the original error is rethrown as-is;
+ * - if any compensation fails, throws a {@link WorkflowStepError} keyed {@link UPGRADE_RECOVERY_FAILED},
+ *   carrying both the original and recovery errors and a flag that the target may no longer match
+ *   the baseline.
+ */
+export class RecoveryScope {
+  private readonly compensations: Compensation[] = [];
```
oshmyheliuk, line 66: @vinayrao2000 I don't see how changes in this PR would work in the case of the removed event provider.
For example, the provider was removed before recovery started. It doesn't look like the provider will be restored. The same with event registrations, etc..
Or is such a case deliberately not covered?
It's not a trivial task to restore an event provider, as it would create a new event provider, it would require a new event subscription and registering this provider in Commerce.
So if we provide a recovery and this is only a partial recovery, doesn't it make the system even more broken?

PR #672, opened 2026-09-09
CEXT-6727 Event upgrade fails with 404 (no event metadata) when a duplicate event provider exists in the org

vinayrao2000:
## Description

Fixes event provider resolution during an upgrade when more than one provider record for the same app exists in the org. `resolveDeployedProvider` now resolves candidates in a defined order — the current workspace-scoped instance id (`appId-providerKey-workspaceId`) first, and the legacy workspace-less id only as a fallback — and, among records sharing an id, prefers the one that actually carries event metadata over an empty duplicate.

## Related Issue

[CEXT-6727](https://jira.corp.adobe.com/browse/CEXT-6727)

## Motivation and Context

The deprecated instance-id scheme (`appId-providerKey`, no workspace) is not unique within an org. When a duplicate provider existed from the same app installed in another workspace, provider churn, or a legacy provider the old unordered lookup could match the deprecated candidate and pick a stale provider from a different workspace. The registration update then targeted the wrong provider, and the upgrade failed with a 404 ("no such event metadata"). Ordering the candidates and preferring the metadata-bearing record makes resolution deterministic and always favors the workspace's own current provider.

This only affects resolution of duplicates of the *same* provider; distinct providers within an app are resolved independently by key and are unchanged.

## How Has This Been Tested?

- Added a unit test that reproduces the bug: a stale deprecated-scheme provider listed ahead of the correct current-scheme provider. The test fails on the old code (wrong provider selected) and passes with the fix, asserting the registration update targets the correct provider id.
- `pnpm typecheck` and the events `apply` suite pass; Biome clean.
- The original 404 was first observed and root-caused live on a workspace containing a duplicate provider.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/.changeset/three-singers-listen.md
+++ b/.changeset/three-singers-listen.md
@@ -1,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": patch
+---
+
+Resolve the correct event provider during an upgrade when a duplicate provider exists in the org. Provider resolution now prefers the current workspace-scoped provider over a stale legacy (workspace-less) one, and prefers a provider that carries metadata over an empty duplicate — so an upgrade no longer fails with a 404 "no such event metadata" by targeting the wrong provider.
--- a/.changeset/three-singers-listen.md
+++ b/.changeset/three-singers-listen.md
@@ -2,4 +2,4 @@
 "@adobe/aio-commerce-lib-app": patch
 ---
 
-Resolve the correct event provider during an upgrade when a duplicate provider exists in the org. Provider resolution now prefers the current workspace-scoped provider over a stale legacy (workspace-less) one, and prefers a provider that carries metadata over an empty duplicate — so an upgrade no longer fails with a 404 "no such event metadata" by targeting the wrong provider.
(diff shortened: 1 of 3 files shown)


.changeset/three-singers-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-app": patch
+---
+
+Resolve the correct event provider during an upgrade when a duplicate provider exists in the org. Provider resolution now prefers the current workspace-scoped provider over a stale legacy (workspace-less) one, and prefers a provider that carries metadata over an empty duplicate — so an upgrade no longer fails with a 404 "no such event metadata" by targeting the wrong provider.
```
oshmyheliuk: You can simplify the changeset description without going into deep detail.

```
Changeset messages must be user-facing and concise — describe what the developer can now do or what changed for them, not what the implementation does internally
```

oshmyheliuk (approved):

.changeset/three-singers-listen.md
vinayrao2000: Done, simplified it to a user-facing line.

PR #676, opened 2026-09-14
CEXT-6527: Default upgradeMode to manual while auto-execution stabilizes

obarcelonap:
## Description

`metadata.upgradeMode` now defaults to `"manual"` instead of `"auto"`. Planning an upgrade stays local-only either way, but only `auto` mode executes it — this closes that off by default until it's safe to re-enable. The `commerce-app-init` and `commerce-app-migrate` skills, which previously wrote `upgradeMode: "auto"` explicitly into every generated config, are updated to match, along with the schema default test and `usage.md`.

This is a temporary, experimental default and will revert to `auto` once automatic upgrade execution is validated end-to-end.

## Related Issue

CEXT-6527

## Motivation and Context

The generated `post-app-deploy` hook auto-executes planned upgrades against Extension Manager and the new backend service. Neither is ready to receive those calls yet, so the default needs to stop doing that automatically until they are, without disabling the upgrade feature outright.

## How Has This Been Tested?

- `pnpm --filter @adobe/aio-commerce-lib-app test` — full suite (1276 tests) passes, including the updated default-value assertion in `metadata.test.ts`.
- `pnpm --filter @adobe/aio-commerce-lib-app typecheck` passes.
- `pnpm lint` passes across all 17 tasks.
- `pnpm format:markdown` made no changes.

N/A

## Types of changes

## Checklist:


obarcelonap:
@iivvaannxx heads up on two `checkJs` type bugs I found and fixed on this branch while looking into an unrelated build failure report — both only surface once a project has at least one user-authored `.ts` action, since that's what makes `ts-loader` build a full TS program that also type-checks the generated `.generated/actions/**/*.js` files (plain `.js`-only projects never hit `ts-loader`, so these were invisible until now).

1. **`app.commerce.config.js` passthrough**: it re-exports `app.commerce.manifest.json` via a `type: "json"` import, which TS widens string literals (e.g. `metadata.upgradeMode: "manual"`) to plain `string` — conflicting with `CommerceAppConfig`'s literal-typed picklist. Fixed with an inline JSDoc type assertion on the re-export.
2. **`config.js.template`**: dereferenced `config.businessConfig.schema` directly. That's safe at runtime (the generator only emits this action when `hasBusinessConfigSchema()` is true), but `businessConfig` and its `schema` field are both optional on `CommerceAppConfig`, so `checkJs` doesn't know that. Fixed the same way, with narrowing casts at both levels.

Also updated a `usage.md` example that had the same unguarded `businessConfig.schema` access, using the exported `hasBusinessConfigSchema` guard instead.

Verified both with a standalone `tsc --checkJs` repro mirroring the scaffolded root `tsconfig.json`, confirming the exact reported errors before the fix and a clean run after. Audited the rest of the generated action templates — no other instances of either pattern.

packages/aio-commerce-lib-app/source/commands/generate/actions/lib.ts
```diff
@@ -121,6 +121,14 @@ async function writeJavaScriptAppConfigModule(
   const isJson = extname(configFilePath) === ".json";
   const importAttributes = isJson ? ' with { type: "json" }' : "";
 
+  // `checkJs` widens `type: "json"` imports to plain `string` fields, which then
+  // fail to satisfy the literal-typed CommerceAppConfig the generated actions pass
+  // it to. Cast it back with a JSDoc type assertion rather than a `@type` tag, since
+  // a `@type` tag would still check (and reject) the widened initializer.
+  const defaultExport = isJson
```
iivvaannxx, line 128: This will work, but will lose the actual typing of the exported file, which we intended to keep via `defineConfig<const T>`. I believe the below should also work and keep the actual type?

```ts
export default /** @type {typeof appConfig & import("@adobe/aio-commerce-lib-app/config").CommerceAppConfigOutputModel} */ (appConfig);
```

iivvaannxx (approved):
Approving because this should work. However I would suggest trying to keep the re-exported type. As a better long-term solution I would evaluate actually getting rid of the passthroughs all together, and making `#app.commerce.config` alias just point to the root config file. 

Since the queued changes are breaking, and this is one, I believe it would be the right moment for this.

PR #682, opened 2026-09-15
fix(skills): document numeric field coercion in commerce-app-eventing

jcuerdo:
## Description

Commerce serializes some event fields as strings inconsistently, even
within the same event — e.g. an id field delivered as `"3"` while a total
on the same payload stays a plain number. `commerce-app-eventing`'s
handler skeleton didn't warn about this: the naive `value.field` access it
showed will silently break any comparison or downstream call (e.g. a REST
body) that expects a real number.

This PR adds a small coercion helper to the handler skeleton, a Common
Issues entry describing the exact symptom, and a Quality Bar check.

## Related Issue

None open — found live while building an event handler for
`observer.sales_order_save_commit_after` that holds large orders: the
handler skipped every event as "invalid" because `entity_id` arrived as
`"3"` (string) instead of `3` (number), even though `grand_total` on the
same payload was a plain number.

## Motivation and Context

Without this, following `commerce-app-eventing` exactly as documented
produces a handler that silently no-ops on real Commerce traffic whenever
a numeric-looking field happens to be serialized as a string — no error,
no build failure, just events being skipped that shouldn't be.

## How Has This Been Tested?

Reproduced against a real event payload:
`{ "entity_id": "3", "grand_total": 59 }`. A handler using
`typeof value.entity_id === "number"` rejects it; the documented
`toFiniteNumber` coercion accepts it and returns `3`. Verified working
end-to-end (build + deploy + real Commerce event) in a downstream app.

Documentation-only change (SKILL.md content); no code paths modified.

N/A

## Types of changes

## Checklist:


plugins/commerce/app-management/skills/commerce-app-eventing/SKILL.md
```diff
@@ -150,6 +150,19 @@ export async function main(params: Record<string, unknown>) {
...
+
+```typescript
+function toFiniteNumber(value: unknown): number | null {
+  if (typeof value === "number") return Number.isFinite(value) ? value : null;
+  if (typeof value === "string" && value.trim() !== "") {
+    const parsed = Number(value);
+    return Number.isFinite(parsed) ? parsed : null;
+  }
+  return null;
+}
```
iivvaannxx: Probably we should throw if we cannot convert the number safely? Let me know your thoughts

Otherwise, we will force the consumers of this code (in case the AI generates it) to handle that null when in most situations you'll just need to refuse continuing.


```suggestion
function toFiniteNumber(value: unknown): number {
  if (typeof value === "number" && Number.isFinite(value)) {
    return value;
  }

  if (typeof value === "string" && value.trim() !== "") {
    const parsed = Number(value);
    if (Number.isFinite(parsed)) {
      return parsed;
    }
  }

  throw new Error(`Could not convert "${value}" to a finite number`);
}
```

iivvaannxx (approved):

PR #689, opened 2026-09-17
CEXT-6282: Support OAuth Server-to-Server credentials in IMS auth

obarcelonap:
## Description

Adds support for the `include-ims-credentials` App Builder action annotation to `@adobe/aio-commerce-lib-auth`'s IMS auth resolution, alongside the existing manually-wired `AIO_COMMERCE_AUTH_IMS_*` params.

- `resolveImsAuthParams` now resolves OAuth Server-to-Server credentials from the annotation (via `@adobe/aio-lib-core-auth`'s `resolveCredentials`) before falling back to the manually-wired params.
- `technicalAccountId`/`technicalAccountEmail` are now optional on `ImsAuthParams`, since the annotation-injected shape has no technical account.
- `getImsAuthProvider` mints tokens through `@adobe/aio-lib-core-auth`'s `generateAccessToken` when no technical account is present, and through the existing `@adobe/aio-lib-ims` flow otherwise — unchanged for current callers.
- `resolveAuthParams` tries IMS resolution first (covering both shapes) before falling back to Integration auth.
- Fixes `makeWorkspaceConfig` (Commerce Eventing) to omit empty technical account fields when none are resolved.
- Corrects `commerce-app-storage` skill docs/comments that described the annotation's injection mechanism incorrectly.

## Related Issue

CEXT-6282

## Motivation and Context

Adobe's `@adobe/aio-lib-core-auth` (PR [adobe/aio-lib-core-auth#15](https://github.com/adobe/aio-lib-core-auth/pull/15)) adds a leaner way for App Builder actions to get OAuth Server-to-Server credentials: the `include-ims-credentials: true` annotation, with no manual `sync-ims-credentials`/`inputs:` wiring needed. This wires that mechanism into this SDK's existing IMS auth resolution and provider, without changing behavior for existing callers.

**Blocked on an external release:** `@adobe/aio-lib-core-auth` PR #15 is still open/unmerged, and the published version is `1.1.0` (no `resolveCredentials` export yet). This PR adds a dependency on `^1.2.0`, which doesn't exist yet — `pnpm install`/CI will fail until that PR merges and releases. Opening as a draft for early review; not mergeable until then.

## How Has This Been Tested?

Added/updated unit tests for `resolveImsAuthParams`, `getImsAuthProvider`, `resolveAuthParams`, and `makeWorkspaceConfig`. Ran `biome check` directly against the edited files (bypassing `pnpm install`, since it can't resolve the pending dependency yet). Full `pnpm test`/`typecheck` are blocked on the same external release.

N/A

## Types of changes

## Checklist:


packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
iivvaannxx, line 1: What I tested was using dummy values for these; I did not test with them absent. If there is some kind of schema validation it will probably fail.

packages/aio-commerce-lib-auth/docs/usage.md
```diff
@@ -70,6 +70,44 @@ export const main = async function (params: Record<string, unknown>) {
 };
 ```
 
+### OAuth Server-to-Server via the `include-ims-credentials` annotation
+
+Annotating an action with `include-ims-credentials: true` in `app.config.yaml` is a leaner
+alternative to manually wiring `AIO_COMMERCE_AUTH_IMS_*` inputs (see [Automatic Auth
+Resolution](#automatic-auth-resolution)): App Builder's runtime injects the workspace's OAuth
+Server-to-Server credentials directly into the action's `params`, with no `sync-ims-credentials`
+step and no explicit `inputs:` entries.
```
iivvaannxx: Very specific IMO, since `sync-ims-credentials` is usually automated by lib-app and not directly run I would trim the details.
```suggestion
Server-to-Server credentials directly into the action's `params`.
```

packages/aio-commerce-lib-auth/docs/usage.md
iivvaannxx, line 1: Nitpick: can we trim em-dashes? It screams AI

packages/aio-commerce-lib-auth/source/lib/ims-auth/provider.ts
```diff
@@ -81,6 +82,13 @@ export function isImsAuthProvider(
 
 /**
  * Creates an {@link ImsAuthProvider} based on the provided configuration.
+ *
+ * Both flavors of {@link ImsAuthParams} are supported transparently: when a technical account
```
iivvaannxx: Nitpick: "Both flavors" makes it sound as if you had already introduced what those are, but you didn't. Quite a long description too, probably can be shortened.

packages/aio-commerce-lib-auth/source/lib/ims-auth/utils.ts
```diff
@@ -111,6 +112,12 @@ export function assertImsAuthParams(
...
  * Resolves an {@link ImsAuthParams} from the given App Builder action inputs.
+ *
+ * Supports both ways of getting OAuth Server-to-Server credentials into an action: the
+ * `include-ims-credentials: true` action annotation is tried first, and the manually-wired
+ * `AIO_COMMERCE_AUTH_IMS_*` params are used as a fallback if the annotation isn't present.
+ *
+
  * @param params The App Builder action inputs to resolve the IMS authentication parameters from.
  * @throws {CommerceSdkValidationError} If the parameters are invalid and cannot be resolved.
  *
```
iivvaannxx, line 122: Typo (empty line):

```suggestion
 * `AIO_COMMERCE_AUTH_IMS_*` params are used as a fallback if the annotation isn't present.
 *
 * @param params The App Builder action inputs to resolve the IMS authentication parameters from.
 * @throws {CommerceSdkValidationError} If the parameters are invalid and cannot be resolved.
 *
```

.changeset/lucky-goats-resolve.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/aio-commerce-lib-auth": minor
```
iivvaannxx, line 2:
```suggestion
"@adobe/aio-commerce-lib-auth": major
```

Let's do a major. This is an interface contract break. Might be a small change, but we should flag it as such, not hide it under a minor.

packages/aio-commerce-lib-app/source/management/domains/events/utils.ts
obarcelonap: Did a research on eventing php module: while it seems these values are not ever used, they're deserialized and it will fail since the receiving type does not allow empty values.

> Summary of findings
> 
> Where the fields live: module-adobe-io-events-client — technicalAccountEmail/technicalAccountId are defined on both JWT and OAuth DTOs (Model/Data/AdobeConsoleConfiguration/), populated from the Adobe Console workspace-config JSON in ConfigurationFactory.php.
> 
> Are they used? Given S2S-only support (OAuth), no:
> - OAuth::technicalAccountEmail — never read anywhere.
> - OAuth::technicalAccountId — never read anywhere.
> - (Only JWT::technicalAccountId is actually consumed — as the JWT sub claim in the legacy/deprecated JWT auth flow, which you're not using.)
> 
> Would omitting them break config today? Yes, but not where you'd expect:
> - Saving the config in admin would still succeed (no validation on save).
> - But consuming the config (e.g., testing the connection, fetching a token) would throw an uncaught fatal TypeError — ConfigurationFactory::createOauth() does unguarded $data["technical_account_email"] array access into a non-nullable string setter under strict_types=1. The surrounding catch (Exception $e) doesn't catch TypeError (which extends Error), so it's not even a friendly error — it's a fatal crash.
> 
> To safely drop them you'd need to: make the OAuth DTO fields nullable or remove them entirely (plus getters/setters), stop reading those keys in createOauth(), and separately consider fixing catch (Exception $e) → catch (\Throwable $e) since that's a latent bug regardless.

Considering this finding, I changed the approach to default to a composed valued if they're not present.

iivvaannxx (approved):

PR #691, opened 2026-09-17
CEXT-6746: invalidate cached config across all state regions

obarcelonap:
## Description

Configuration cache invalidation was only being applied to a single `aio-lib-state` region per action instance. When a config value changed, action instances running in other regions could keep serving the stale, previously-cached value until it expired on its own, instead of picking up the change immediately. This fix invalidates the cached entry in every supported region whenever a config value is changed or removed.

## Related Issue

CEXT-6746

## Motivation and Context

Configuration changes need to be reflected consistently regardless of which region serves a given request.

## How Has This Been Tested?

Added/updated unit tests covering multi-region cache invalidation (including partial-region failures) and confirmed existing coverage for `aio-commerce-lib-config` still passes, along with `pnpm lint` and `pnpm typecheck` across the monorepo.

N/A

## Types of changes

## Checklist:


packages/aio-commerce-lib-config/test/unit/modules/configuration/configuration-repository.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      expect(mockFiles.write).toHaveBeenCalledWith(
+        FILE_PATH,
+        JSON.stringify(payload),
+      );
+
+      // Every region (including the current one) gets its stale entry cleared...
+      for (const state of mockStatesByRegion) {
+        expect(state.delete).toHaveBeenCalledWith(
+          `configuration.${SCOPE_CODE}`,
+        );
```
iivvaannxx: Not a blocker, but I would prefer this being tested in a more integration-like manner faking the store instead of asserting the internal method was called.

obarcelonap: Good point — rewrote these to seed the fake state/files stores directly and assert on what they actually hold after the call (via `get`/`read`), instead of asserting which internal methods were invoked. Kept the TTL check call-based since the fake doesn't model expiration. See 6e65f0d9.

obarcelonap: done

iivvaannxx (approved):

